Issue #15, opened 2015-01-06
CWMetricsFactory.java has a Hard-Coded 200 Max Metrics before Flush


christopheclc:
Hi,

After seeing our Cloudwatch bill surge to $100+/day, we found the issue seems to be in the KCL. More specifically, in CWMetricsFactory.java at line 33:

/*
     * If the CWPublisherRunnable accumulates more than FLUSH_SIZE distinct metrics, it will call CloudWatch
     * immediately instead of waiting for the next scheduled call.
     */
    private static final int FLUSH_SIZE = 200;

Some KCL metrics are being put in cloudwatch 50,000 times per minute. Can you please make this configurable so that we can set it much higher?

Thank you

gauravgh:
Thank you for the suggestion. We'll look into making this configurable.

Would you be able to start a thread on our forum (https://forums.aws.amazon.com/forum.jspa?forumID=169) with some more details about your use case? For example, the number of shards in the stream being processed by the KCL, number of Workers in your application, whether you're using CWMetricsFactory to publish your own application metrics (in addition to the metrics published by the KCL). We may be able to provide a better response based on those details.

Sincerely,
Gaurav

andrepintorj:
There is also another setting that must be configurable related to this issue. Suppose I don't want custom metrics, I need to configure KCL worker to use the NullMetricsFactory instead of CWMetricsFactory. This is a must have for this library.

PR #24, opened 2015-04-24
MultiLangDaemon: Customizing CloudWatch Metrics Factory

andrepintorj:
This pull request refers to an issue that we faced using the MultiLangDaemon that is publishes custom metrics continuously to CloudWatch, thus increasing our bill on this service. Since we wanted the ability to use the NullMetricsFactory on some of our apps, we have decided implementing this feature. It consists in reading the parameter customCloudWatchMetricsFactoryName on the properties file. The possible values are "CWMetricsFactory" (the default), NullMetricsFactory and LogMetricsFactory.

This feature can be somewhat related to the issue #15 .


manango:
@andrepintorj thanks for your feedback. With KCL 1.5.0 release, we have added two configuration options related to metrics to allow greater control over how KCL CloudWatch metrics are reported. More information is available on [the KCL monitoring](http://docs.aws.amazon.com/kinesis/latest/dev/monitoring-with-kcl.html) page.

rdsubhas:
@manango The page is not really clear. There are lots of tables, and its looks as if I have to disable each metric by name. Your comment says *two configuration options* - can you please help me find those two? Thanks for the help.

rdsubhas:
If I'm right - are those two properties `metricsLevel` and `metricsEnabledDimensions` ?

PR #31, opened 2015-07-09
Add ProcessRecordsInput.getGetRecordsResult. Fixes #27.

### LINKED ISSUE #27: Expose MillisBehindLatest from GetRecords() call
The Kinesis API recently started returning MillisBehindLatest in the response to GetRecords(). 

http://docs.aws.amazon.com/kinesis/latest/APIReference/API_GetRecords.html

>MillisBehindLatest
>The number of milliseconds the GetRecords response is from the tip of the stream, indicating how far behind current time the consumer is. A value of zero indicates record processing is caught up, and there are no new records to process at this moment.
>
>Type: Long
>
>Valid range: Minimum value of 0.

The current KCL only exposes the records to the IRecordProcessor. It would be useful to expose the MillisBehindLatest to the IRecordProcessors as well, so they would know how far behind they are.

stephenh:
Admittedly just kind of threw this together without hearing from the ticket about which approach you preferred/if any, but hey it was easy. Probably easier for you to just re-code than port the patch into your internal repo. :-)


manango:
Thanks @stephenh for the suggestion. We have already exposed millisBehindLatest value to record processor with the 1.5.0 KCL release today.

Issue #23, opened 2015-04-23
Better proxy support out of the box


rileymcdowell:
The main issue is that the library doesn't use system proxy environment variables by default. On most linux systems, the environment variables http_proxy, https_proxy, and no_proxy are assumed to describe the correct proxies to use for http and https connections. This library does not use these values even if they are present. 

Passing -Dhttp.proxyHost and -Dhttp.proxyPort to java do not seem to make a difference.

kevincdeng:
Thank you for the suggestion lambdaman. We will look into this.

rrahul963:
is there a away to specify the proxy settings?

rileymcdowell:
I am currently using a workaround to the lack of http proxy suport. I set up a SOCKS5 proxy on port 1080 in my local environment using an ssh tunnel to a machine that lives inside of an AWS VPC. The machine in AWS has unrestricted outgoing internet access. 

This means that I can append `-DsocksProxyHost=localhost -DsocksProxyPort=1080` to the java command locally and it will tunnel all http traffic out through the remote connection. I'm using the autossh utility to keep this SOCKS5 proxy connection open at all times.

You might assume that the http proxy versions of these java runtime flags would work without the need for configuring a SOCKS5 proxy. Unfortunately, passing `-DhttpProxyHost=<host_name> -DhttpProxyPort=<port_number>`  did not work for me, although these options are better documented in the java community than the socks variants.

mscharp:
Has any support been added in the 3+ years this request has been opened?

PR #33, opened 2015-07-16
Add some basic support for http proxy settings to MultiLangDaemon

samcday:
Addresses #23

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
+++ b/src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
@@ -16,6 +16,12 @@ package com.amazonaws.services.kinesis.multilang;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.InetSocketAddress;
+import java.net.Proxy;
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
```diff
@@ -100,6 +117,29 @@ public MultiLangDaemonConfig(String propertiesFile,
         prepare(processingLanguage);
     }
 
+    private ClientConfiguration buildClientConfig(final String serviceName, KinesisClientLibConfiguration config)
+    {
+        try
+        {
+            Region region = RegionUtils.getRegion(config.getRegionName());
+            String serviceEndpoint = serviceName + "." + region.getName() + "." + region.getDomain();
```
samcday: FYI - I know this should be using `Region#getServiceEndpoint()`, but the [regions.xml](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/resources/com/amazonaws/regions/regions.xml) that ships with even the latest AWS SDK doesn't include us-west-1 as a valid region for Kinesis endpoint ...

PR #59, opened 2016-03-02
Handle spurious lease renewal failures gracefully.

samcday:
If the request to conditionally update a lease counter in DynamoDB fails, it's
considered a failure to renew the lease. This is a good thing, except if the
request failure was just because of connectivity problems. In this case the
counter _did_ update in DynamoDB, but the Dynamo client retries the request
which then fails the update condition (since the lease counter no longer
matches expected value).

To handle this gracefully we opt to get the lease record from Dynamo and
examine the lease owner and counter. If it matches what we were expecting,
then we consider renewal a success.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseManager.java
+++ b/src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseManager.java
@@ -19,6 +19,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
+import com.amazonaws.services.kinesis.leases.util.DynamoUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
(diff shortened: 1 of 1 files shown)


samcday:
FYI we've been running this change in production for a couple of weeks now in a KCL app that consumes from a 50 shard Kinesis stream. It completely eliminated the spurious shard dropping issues we were experiencing previously.

pfifer:
I've started the process of merging the pull request internally. Unfortunately I had to delay merging this for the 1.6.4 release. I do plan to have this merged, and released in the near future.

pfifer:
Please confirm that we can use, modify, copy, and redistribute this contribution. After I get the confirmations I can begin the release process.

Thanks.

samcday:
Confirmed.

pfifer:
I'm still working on merging this.  Since adding the unit tests this causes them to fail, since they don't expect a second call to be made.  I just need to fix the unit tests to handle the new behavior.

pfifer:
Merged as part of #247 

Thanks for providing this, and sorry about the long delay in merging it.

samcday:
Better late than never! :)

PR #65, opened 2016-04-12
Add support for configuring DynamoDB endpoint

mjchoi:
Adding a new field named `dynamoDBEndpoint` to the .properties file
that gets passed into the KCL multi-lang daemon. We need this ability
to point the KCL worker at a local instance of DynamoDB rather than in
AWS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 target/
 AwsCredentials.properties
+*.swp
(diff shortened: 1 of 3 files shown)


pfifer:
This was merged as part of #99. Thanks for submitting the PR.

PR #67, opened 2016-05-09
Fix #36 Avoid failure on start due to CreateTable

### LINKED ISSUE #36: LeaseManager requires dynamoDb:CreateTable IAM permissions even if the table already exists.
The lease manager determines if the DynamoDb table already exists by attempting to create it, and ignoring a  `ResourceInUseException` .  This is non-optimal in an environment where the creation of resources is restricted, and the table has already been initialized.  

It would be more intuitive to use a DescribeTableRequest to determine if the table exists, since it's considered a read-only operation.

delphyne:
Current implementation uses a ResourceInUse side-effect to detect that the table already exists.
This requires that the user/role running the KCL have the CreateTable permission.  Update the
LeaseManager to check if the table exists to allow limited privilege users to run KCL code.


pfifer:
The change is committed, and will be released to Maven soon.

PR #82, opened 2016-06-27
Join metrics publishing thread when shutting factory down.

rankinc:
We need `CWMetricsFactory.publicationThread` to complete before we can shut its `AwsCloudWatch` client down cleanly.


pfifer:
I think I would prefer to see the use of join with a timeout, and log message.  This would ensure that if the metrics thread got stuck, there would be an indication of why shutdown is being blocked.  

Otherwise this looks good, and should ensure consistency during shutdown.

Finally I just need you to confirm that we can use, modify, copy, and redistribute this contribution. 

Thanks.

rankinc:

PR #84, opened 2016-06-29
Always shutdown leaseRenewalThreadpool on exit.

rankinc:
Stopping `LeaseCoordinator` should also shutdown its `leaseRenewalThreadpool` field.


pfifer:
This looks good to me.  

I will begin to merge the changes after getting the confirmation.

Thanks.

rankinc:

Issue #85, opened 2016-07-01
please expose "ApproximateArrivalTimestamp" via JsonFriendlyRecord


zdjohn:
Even though ApproximateArrivalTimestamp is becoming available in Record. However it is not exposed in JsonFriendlyRecord. 
This means "ApproximateArrivalTimestamp" is not exposed to anyone who uses KCL and use ProcessRecordsMessage to consume Kinesis Stream. (e.g. .Net developers like me)

The fix is relative simple, and I am happy to provide PR for it.

PS: https://github.com/awslabs/amazon-kinesis-client-net is over a year old, is there any plan on that regards?

pfifer:
Thanks for the feedback and the PR.  We want to get all of the MultiLanguageDaemon clients up to date with the latest release of the Kinesis Client.  We should be able to add the approximateArrivalTimestamp as part of the updates.  

The PR looks pretty straight forward, and I will look into merging it.  This will also require updating the other clients to expose the property as well.

zdjohn:
@pfifer awesome thanks!

pfifer:
I've merged the PR.  I'm looking into adding support for the other languages, but I don't have a timeline on that right now.

pfifer:
Support for the field is now available, but there are steps required to update the clients.  Currently the Python client has been updated to use the new functionality, and we're investigating updating the remainder of the clients.

pfifer:
I created a issue to track adding the new features to the .NET client.  I'm going to close this for now, please feel free to reopen if you have more questions.

PR #86, opened 2016-07-01
expose approximateArrivalTimestamp from JsonFriendlyRecord

zdjohn:
regarding on issue https://github.com/awslabs/amazon-kinesis-client/issues/85

> Even though ApproximateArrivalTimestamp is becoming available in Record. However it is not exposed in JsonFriendlyRecord. 
>This means "ApproximateArrivalTimestamp" is not exposed to anyone who uses KCL and use ProcessRecordsMessage to consume Kinesis Stream. (e.g. .Net developers like me)

tested locally and working as expected


pfifer:
Please confirm that we can use, modify, copy, and redistribute this contribution. After that I can begin merging the changes.

Thanks.

zdjohn:

sid88in:
@zdjohn I am trying to insert approximateArrivalTimestamp in the S3 from the record in firehose. 

POST API -> FIREHOSE -> S3

Is it possible to get approximateArrivalTimestamp in S3 without using lambda.

PR #99, opened 2016-08-22
Allow for credential providers with basic arg types

joshuamorris3:
I've added the ability to use credential providers with string type arguments in their constructors. Specifically this was so I could use a credential provider like com.amazonaws.auth.profile.ProfileCredentialsProvider, which you can provide a profile name to, or profile config file path and profile name to when instantiating it. The basic implementation is append the args to the provider name with the | delimiter. For example in a the properties file used by KCL to setup a client you can provide the following:
AWSCredentialsProvider = com.amazonaws.auth.profile.ProfileCredentialsProvider|profileName
where the ProfileCredentialsProvider will be instantiated using the profile name single arg constructor.

I've also merged in the https://github.com/awslabs/amazon-kinesis-client/pull/65 PR for my own local use.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
@@ -14,6 +14,7 @@
  */
 package com.amazonaws.services.kinesis.clientlibrary.lib.worker;
 
+import java.util.Date;
 import java.util.Set;
 
 import com.amazonaws.ClientConfiguration;
@@ -40,7 +41,7 @@ public class KinesisClientLibConfiguration {
     /**
      * Fail over time in milliseconds. A worker which does not renew it's lease within this time interval
      * will be regarded as having problems and it's shards will be assigned to other workers.
-     * For applications that have a large number of shards, this msy be set to a higher number to reduce
+     * For applications that have a large number of shards, this may be set to a higher number to reduce
      * the number of DynamoDB IOPS required for tracking leases.
      */
     public static final long DEFAULT_FAILOVER_TIME_MILLIS = 10000L;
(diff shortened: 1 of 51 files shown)


shakdoesgithub:
Hi @joshuamorris3 do you know what the merge conflicts are? 
I've forked from master and applied the changes you have created, I can submit a PR for this w/ the conflicts resolved if you want me to.  This feature is something we really need.

joshuamorris3:
Hi @shakdoesgithub apologies for the delay in getting back to you. It's been a busy week. I see that you've created another PR (https://github.com/awslabs/amazon-kinesis-client/pull/106) with just the credential provider enhancements. I should have split the changes originally as a good practice. I'll look into this PR's conflicts, though feel free to leave the other PR open.

shakdoesgithub:
sounds good @joshuamorris3!  This is a very helpful pull request!

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
```diff
@@ -276,6 +278,7 @@ public KinesisClientLibConfiguration(String applicationName,
     public KinesisClientLibConfiguration(String applicationName,
             String streamName,
             String kinesisEndpoint,
+            String dynamoDBEndpoint,
```
pfifer: Since this changes a public constructor, could you please add a constructor override?  This could break people's code due to the addition of a new constructor parameter.

joshuamorris3: @pfifer I think I created the backwards compatibility you were asking for. I created another constructor that matches the old signature. This delegates to the new constructor with the dynamoDBEndpoint addition.

pfifer:
Looks good, thanks again for doing this.  I just need one more confirmation before I can start merging the change.

Thanks.

joshuamorris3:
Yes, you can use, modify, copy, and redistribute this contribution.

Thanks.

pfifer:
Might have been my merger order, but there is now a conflict.  Could you please resolve it, and update the PR.

Thanks

joshuamorris3:
Done.

PR #106, opened 2016-09-28
Allow for credential providers with basic arg types(updated, no conflicts)

shakdoesgithub:
This is the same pull request as https://github.com/awslabs/amazon-kinesis-client/pull/99 except it doesn't have the dynamo db endpoint changes.  This is a straight copy and paste from that pull request.  This pull request is off master so it shouldn't have any conflicts like the other one.  

All credits should go to @joshuamorris3.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,6 +34,11 @@
   <dependencies>
     <dependency>
       <groupId>com.amazonaws</groupId>
+      <artifactId>aws-java-sdk-sts</artifactId>
+      <version>${aws-java-sdk.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.amazonaws</groupId>
       <artifactId>aws-java-sdk-dynamodb</artifactId>
       <version>${aws-java-sdk.version}</version>
     </dependency>
--- a/src/test/java/com/amazonaws/services/kinesis/clientlibrary/config/AWSCredentialsProviderPropertyValueDecoderTest.java
+++ b/src/test/java/com/amazonaws/services/kinesis/clientlibrary/config/AWSCredentialsProviderPropertyValueDecoderTest.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Amazon Software License (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ * http://aws.amazon.com/asl/
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.services.kinesis.clientlibrary.config;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.io.ByteArrayInputStream;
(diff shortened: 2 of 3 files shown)


src/test/java/com/amazonaws/services/kinesis/clientlibrary/config/AWSCredentialsProviderPropertyValueDecoderTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.commons.lang.StringUtils;
+import org.junit.Test;
+
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.auth.AWSCredentialsProviderChain;
+import com.amazonaws.auth.profile.ProfileCredentialsProvider;
+import com.amazonaws.services.kinesis.clientlibrary.config.AWSCredentialsProviderPropertyValueDecoder;
+
+public class AWSCredentialsProviderPropertyValueDecoderTest {
```
pfifer: The tests don't appear to ever test if the correct class was instantiated.  At least for the custom classes it should be possible to determine if they're instantiated correctly by calling getCredentials on the provider.

pom.xml
```diff
@@ -34,6 +34,11 @@
   <dependencies>
     <dependency>
       <groupId>com.amazonaws</groupId>
+      <artifactId>aws-java-sdk-sts</artifactId>
```
pfifer: What classes need the sts dependency?

src/test/java/com/amazonaws/services/kinesis/clientlibrary/config/AWSCredentialsProviderPropertyValueDecoderTest.java
joshuamorris3: Let me add those. Again this will be in PR [#99](https://github.com/awslabs/amazon-kinesis-client/pull/99)

joshuamorris3: I've updated the tests on PR 99.

pom.xml
shakdoesgithub: My use case was using the STSAssumeRoleSessionCredentialsProvider

pfifer: This is something I'm working on to make it easier to launch MultiLang Daemon clients that might work for you.  It does require the addition of Maven, but only to create a launch script, and combined set of jars.

https://gist.github.com/pfifer/738c4f8b2eeb5757b1d8962d60e6f1e0

To build this you just run
```
mvn compile appassembler:assemble
```

That will create a launcher script, and jar directory in target/appassembler.  You just need to copy all the files in that to where ever you want the launcher to live, though the easiest is to co-locate it with where your properties file is located.

To run the application you need to do two steps:

1. Set the variable `CLASSPATH_PREFIX` to the directory where your properties file is located.
2. Start the daemon by running bin/multilang-daemon <properties files>
     * I've not tested this on Windows, but you should be able to run bin/multilang-daemon.bat

When using the generated script it will not use the jars download by the language specific module, but instead the jars that will be in the repo directory.

The example gist I've included adds 3 extra dependencies specifically logback, and slf4j bridges for Apache Commons Log.  
You should be able to add the STS credentials dependency, and then use the code in this PR to load, and use the new provider.

If you try this out please let me know how it works for you.  I haven't had much time to test this yet, but I do plan to do more testing.  This changes will make logging configuration for multilang daemon applications easier.

pfifer:
I want to make sure this clear, and I have it right.  

I won't be merging this PR, but instead will be merging #99.  Do I have that correct?

pfifer:
Closing This for Upcoming Release.

pom.xml
shakdoesgithub: Coming back to this, what are your suggestions if one doesn't have maven?

PR #109, opened 2016-09-30
Allow for a Graceful Shutdown of the Worker

pfifer:
Add a new method to the worker requestShutdown that allows the worker to
gracefully shutdown all record processors.  The graceful shutdown gives
the record processors a last chance to checkpoint before they're
terminated.

To use these new features the record processor must implement
IShutdownnotificationaware.

Some cleanup to try and make the state transitions of the record
processor more clear.


DavidDeCoding:
What is the difference between the shutdown and shutdownRequested? What is the use case for both of them? Can you provide an example for both? Thanks. @pfifer

pfifer:
@DavidDeCoding The `shutdown` method will terminate the worker, without giving the record processor any change to checkpoint.  The `requestShutdown` method notifies the record processor, if it implements `IShutdownNotificationAware`, of the upcoming shutdown, and gives them a chance to checkpoint.  `requestShutdown` will also shutdown the worker instance same as `shutdown`.

In general I would recommend using `requestShutdown` unless you have a need to shutdown without giving the record processor an explicit chance to checkpoint.

DavidDeCoding:
Thanks @pfifer. I thought so too.

PR #111, opened 2016-10-02
Feature/multiple cred providers

rgfindl:
Add ability to specify different credential providers for Kinesis, DynamoDB, and CloudWatch. This is needed when accessing a cross-account Kineses stream using an assumed role.  #106 will help in defining the STSAssumeRoleSessionCredentialsProvider.

### DIFF
--- a/src/test/java/com/amazonaws/services/kinesis/clientlibrary/config/KinesisClientLibConfiguratorTest.java
+++ b/src/test/java/com/amazonaws/services/kinesis/clientlibrary/config/KinesisClientLibConfiguratorTest.java
@@ -329,6 +329,29 @@ public class KinesisClientLibConfiguratorTest {
         }
     }
 
+    @Test
+    public void testWithDifferentAWSCredentialsForDynamoDBAndCloudWatch() {
+        String test = StringUtils.join(new String[] {
+                "streamName = a",
+                "applicationName = b",
+                "AWSCredentialsProvider = " + credentialName1,
+                "AWSCredentialsProviderDynamoDB = " + credentialName1,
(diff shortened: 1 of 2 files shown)


src/test/java/com/amazonaws/services/kinesis/clientlibrary/config/KinesisClientLibConfiguratorTest.java
```diff
@@ -329,6 +329,29 @@ public void testWithAWSCredentialsFailed() {
         }
     }
 
+    @Test
+    public void testWithDifferentAWSCredentialsForDynamoDBAndCloudWatch() {
```
pfifer: This test appears to succeed with the current code.  It might make sense to test with different class names for each credential provider.  The credential decoder has the ability to load [custom class names](https://github.com/awslabs/amazon-kinesis-client/blob/master/src/main/java/com/amazonaws/services/kinesis/clientlibrary/config/AWSCredentialsProviderPropertyValueDecoder.java#L97).

pfifer:
Looks like the tests failed due to a race condition that I haven't been able to track down. 

Thanks for sending this PR. I just need one final confirmation, before I can merge this.

Thanks.

shakdoesgithub:
@rgfindl this pull request is really helpful!

rgfindl:
Yes, you can use, modify, copy, and redistribute this contribution.  Thanks!

holmesjr:
Do any of the KCL clients (e.g. python) support this feature via config files?

holmesjr:
Sorry, I understand now. The config file entries just haven't been updated in the sample file.

PR #119, opened 2016-11-17
Allow systems to track consumer initialization error

daltrogama:
When some problem occours on initialization of DynamoDB or simple credentials validation, the exception is logged but the calling system won't never get notified. It's interesting that the exception is throwed so that some action may be taked at application level.


pfifer:
Thanks for submitting this, but the unit tests failed for this change.  Can you check the failures, and update the unit tests

PR #125, opened 2017-01-08
update AWS SDK dependency to 1.11.16

acgray:
The current AWS SDK dependency (1.11.14) does not support IAM roles for ECS task roles.

Bumping dependency version to 1.11.16.


pfifer:
Thanks for submitting this, but I went for a newer version of the SDK.

PR #130, opened 2017-01-20
Update README: fix 'MultiLangDaemon' typo

jim-slattery-rs:


pfifer:
Thanks for doing this, just need a confirmation for distribution.

jim-slattery-rs:
Confirmed: you can use, modify, copy, and redistribute this contribution.

Thanks!

PR #135, opened 2017-02-15
Use ExecutorService.newFixedThreadPool for LeaseRenewer

slam:
The existing ThreadPoolExecutor was misconfigured and caused
the thread pool to recycle idle threads continuously. VisualVM
showed that a thread got recycled about every 3 seconds, creating
and destroying several hundred threads in about 15 minutes.

Use ExecutorService.newFixedThreadPool instead, as recommended
by the javadoc for ExecutorService.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseCoordinator.java
+++ b/src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseCoordinator.java
@@ -19,11 +19,9 @@ import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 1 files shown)


slam:
Here is a VisualVM screenshot of one of our production consumer that had been running for about 10 hours. Note the LeaseRenewer thread number went to about 17000...

image

slam:
With this fix, the LeaseRenewer thread count stays at 20.

pfifer:
Thanks for sending this.  I'm taking a look at this now.

pfifer:
Before I forget, please confirm that we can use, modify, copy, and redistribute this contribution. 

Thanks.

slam:
Yes, confirmed.

joelittlejohn:
@pfifer Any chance of a new release to include this fix sometime soon? It's a reasonably serious issue that would be worth a release IMO.

pfifer:
This should be released a little later today if all my testing goes right.  I've been testing these changes for a while so I don't expect any real issues.

Issue #51, opened 2015-12-11
Can't set DynamoDB endpoint for the MultiLangDaemon


rusvdw:
This was mentioned in issue #34, but there doesn't appear to be any way to address this from the MultiLangDaemon. Could we have it added as a "dynamodbEndpoint" value in the properties file by any chance?

robotrobot:
+1

robotrobot:
A workaround might be to extend `com.amazonaws.services.kinesis.multilang.MultiLangDaemon` and override the main method and add in the local dynamo config.

whummer:
+1 would be great to have this property to simplify local test setups. Thanks

pfifer:
It should be possible to set the DynamoDB endpoint by including the property dynamoDBEndpoint in the configuration for the MultiLangDaemon.

This was added in [Release 1.7.1](https://github.com/awslabs/amazon-kinesis-client/releases/tag/v1.7.1).

Feel free to reopen if you have more questions.

chris-chambers:
This functionality is partially broken as of 1.7.3.  In the Worker class, there are two paths that needed to be modified, but only one of them was updated in the initial implementation.

Compare [here (present)](https://github.com/awslabs/amazon-kinesis-client/blob/31615a5d1896ecf2fe4929aebe074ae80c146c04/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java#L259) with [here (missing)](https://github.com/awslabs/amazon-kinesis-client/blob/31615a5d1896ecf2fe4929aebe074ae80c146c04/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java#L1058).

Specifically, the "missing" path is hit when using the `aws-kcl` package from npm.  I haven't checked the behavior of other ways of running the KCL.

There are a few inconsistencies between these two blocks (CloudWatch setup seems to differ).  Could the two paths be unified?

PR #142, opened 2017-02-19
Add missing init for DynamoDB client in Worker (#51)

chris-chambers:
When the `dynamoDBEndpoint` configuration property was added, not all paths that
should have made use of it were updated.  Both paths now use the config value to
set up the DynamoDB client.


pfifer:
Thanks for doing this.  If you could please confirm that we can use, modify, copy, and redistribute this contribution. 

Thanks.

chris-chambers:
Yes, of course!  You may use, modify, copy, and redistribute this contribution.

PR #174, opened 2017-06-01
Clean shutdown when multilang deamen is terminated by giving one last chance to checkpoint.

mikramulhaq:


src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangRecordProcessor.java
```diff
@@ -136,6 +141,24 @@ public void shutdown(ShutdownInput shutdownInput) {
...
+    @Override
+    public void shutdownRequested(IRecordProcessorCheckpointer checkpointer) {
+        LOG.info("Shutdown is requested.");
+        if (!initialized) {
+            LOG.info("Record processor was not initialized so no need to initiate a final checkpoint.");
+            return;
+        }
+        try {
+            LOG.info("Requesting a checkpoint on shutdown notification.");
+            checkpointer.checkpoint();
```
pfifer: Whether to checkpoint is determined by the application, so checkpointing here isn't safe.  This should use a similar dispatch mechanism that processRecords uses.  It might be possible to shortcut the dispatch at this time, and just make whether to checkpoint a configuration flag.

The [dispatch code for processRecords](https://github.com/awslabs/amazon-kinesis-client/blob/master/src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangProtocol.java#L84) for reference.

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemon.java
```diff
@@ -140,11 +140,27 @@ public static void main(String[] args) {
...
 
         // Daemon
-        MultiLangDaemon daemon = new MultiLangDaemon(
+        final MultiLangDaemon daemon = new MultiLangDaemon(
                 config.getKinesisClientLibConfiguration(),
                 config.getRecordProcessorFactory(),
                 executorService);
 
+        Runtime.getRuntime().addShutdownHook(new Thread()
+        {
```
pfifer: Can you please match the existing style with the curly brace on the same line.

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemon.java
```diff
@@ -140,11 +140,27 @@ public static void main(String[] args) {
...
+            @Override
+            public void run()
+            {
+                LOG.info("Process terminanted, will initiate shutdown.");
+                try {
+                    Future<Void> fut = daemon.worker.requestShutdown();
+                    fut.get();
```
pfifer: In a worst case scenario this could cause the shutdown to become stuck.  It would be best to start with some timeout value of how long to wait for the shutdown request to complete.  I usually start with something in the range of 5 seconds.  In the future it may make sense to make it configurable.

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangRecordProcessor.java
```diff
@@ -136,6 +141,24 @@ public void shutdown(ShutdownInput shutdownInput) {
...
+        } catch (InvalidStateException e) {
+            LOG.error("Checkpoint triggered during shutdown encountered InvalidStateException: " + e.toString());
```
pfifer: Java will automatically call toString during string concatenation.  Additionally for exceptions it's common to pass the exception as the second argument to the logger call.  This will allow the logger to emit a stack trace as well e.g.
``` java
LOG.error("Checkpoint triggered during shutdown encountered InvalidStateException: " + e, e);
```

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemon.java
```diff
@@ -140,11 +143,26 @@ public static void main(String[] args) {
...
                 config.getRecordProcessorFactory(),
                 executorService);
 
+        Runtime.getRuntime().addShutdownHook(new Thread() {
+            @Override
+            public void run()
+            {
+                LOG.info("Process terminanted, will initiate shutdown.");
+                try {
+                    Future<Void> fut = daemon.worker.requestShutdown();
```
anandswaminathan: You need to ensure that when `requestShutdown ` is invoked. There is no process records invoked. Need a lock may be ?!

pfifer: The way dispatch is handled in the worker already guarantees this.  The record processors are represented as a state machine, and all request shutdown does is request that on the change event to go to the Shutdown Notification Requested state.  [This is the diagram for the state changes](https://github.com/awslabs/amazon-kinesis-client/blob/master/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ConsumerStates.java#L3).

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangProtocol.java
```diff
@@ -100,6 +101,18 @@ boolean shutdown(IRecordProcessorCheckpointer checkpointer, ShutdownReason reaso
     }
 
     /**
+     * Writes a {@link ShutdownRequestedMessage} to the child process's STDIN and waits for the child process to respond with a
+     * {@link StatusMessage} on its STDOUT.
+     *
+     * @param checkpointer A checkpointer.
+     * @return Whether or not this operation succeeded.
+     */
+    boolean shutdownRequested(IRecordProcessorCheckpointer checkpointer) {
```
pfifer: Could you please add some tests for this method in [MultiLangProtocolTest.java](https://github.com/awslabs/amazon-kinesis-client/blob/master/src/test/java/com/amazonaws/services/kinesis/multilang/MultiLangProtocolTest.java)

pfifer:
Thanks everything looks good.  Starting the process to merge the changes now.

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangProtocol.java
mikramulhaq: done!

pfifer:
One last thing:

Please confirm that we can use, modify, copy, and redistribute this contribution. Thanks.

mikramulhaq:
@pfifer - yes confirmed. You can modify, copy and redistribute this contribution. Thanks,

PR #180, opened 2017-06-20
Fix for Issue 55

cyb3rd0g1:
Changed getAllShardIds to use the 'shards' variable rather than calling getShardList() twice, added private set class variable 'shardIDSet' to hold shard IDs and logic to save them in getShardList(), and logic to check whether the shard ID exists when calling getShard(). If the shard ID is not contained in the set, getShardList() is called. This is a fix for issue 55 where workers would throw a warning if they were assigned to a shard whose parent they had never processed. This generally occurred after splitting a shard.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/proxies/KinesisProxy.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/proxies/KinesisProxy.java
@@ -66,6 +66,8 @@ public class KinesisProxy implements IKinesisProxyExtended {
 
     private final String streamName;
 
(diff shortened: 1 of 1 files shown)


pfifer:
Thanks for submitting this, just a couple of issues:
* This uses a set in a possible multi-threaded code path.  Can you modify this to be thread safe?
* There are no tests for the change.  Tests help us verify the feature, and prevent possible regressions.  Can you add some tests?

stair-aws:
Closing due to merge conflicts, and ask (from 2017) for thread-safe code that is unit tested. If desired, please reopen and address the issues.

PR #188, opened 2017-07-13
Add two phase commit

wdbaruni:
Add support to checkpoint record processing in idempotent way. It works by first creating an intent to checkpoint for a shard, which will store pendingCheckpointValue in the lease table. Upon failover, the pendingCheckpointValue will be passed to the new RecordProcessor's InitializationInput method to be able to checkpoint at the same record.

To reduce the impact of this change on current interfaces,  two phase commits is only available for v2 of IRecordProcessor


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PreparedCheckpointer.java
```diff
@@ -0,0 +1,49 @@
...
+    public PreparedCheckpointer(ExtendedSequenceNumber snToCheckpoint, IRecordProcessorCheckpointer checkpointer) {
+        this.snToCheckpoint = snToCheckpoint;
+        this.checkpointer = checkpointer;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public ExtendedSequenceNumber getSNOfPendingCheckpoint() {
```
wdbaruni: renamed to getPendingCheckpoint() to be consistent with ICheckpoint.getCheckpoint()

wdbaruni:
Merged with 1.8.1

PR #204, opened 2017-08-16
MultiLangDaemon: Make shutdown grace configurable

AlexCharlton:
The shutdown grace added to the MultiLangDaemon in 1.7.6 proved to be essential for my application, but the hard-coded 5 second grace is too harsh for my needs. This adds a configuration parameter for the grace period.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
@@ -173,6 +173,11 @@ public class KinesisClientLibConfiguration {
     public static final ShardPrioritization DEFAULT_SHARD_PRIORITIZATION = new NoOpShardPrioritization();
 
     /**
+     * The amount of milliseconds to wait before graceful shutdown forcefully terminates.
+     */
+    public static final long DEFAULT_SHUTDOWN_GRACE_MILLIS = 5000L;
+
+    /**
      * The size of the thread pool to create for the lease renewer to use.
      */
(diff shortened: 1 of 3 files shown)


pfifer:

Thanks.

AlexCharlton:
You may use, modify, copy, and redistribute this contribution.

pfifer:
One more question: Is the change coming from yourself, or from Instacart?

jackdempsey:
@pfifer Can you share some info on the context behind the question? (why it matters, what it means, etc). Thanks

pfifer:
If it's coming from Instacart we just need to confirm that Alex has authority from Instacart to contribute the change to us under the same language:
> Please confirm that we can use, modify, copy, and redistribute this contribution.

I'm sorry for all the extra steps.   

Also can you let me know which of the multi-lang daemon clients use.

jackdempsey:
Hey pfifer, thanks and no worries. What do you need to confirm he has the authority? (He's on my team here and I can confirm he does)

AlexCharlton:
Yes @pfifer, these changes are coming from Instacart. I can also confirm that @jackdempsey is my manager, and that I have it from the highest authority that you can use, modify, copy and redistribute this contribution :)

Sorry, I'm not sure what you mean by:

> Also can you let me know which of the multi-lang daemon clients use.

pfifer:
I need to update the multi-lang clients to use the newer version of the KCL, and was curious which language you're were using.  

I do plan to include this change in the upcoming release.

AlexCharlton:
Ah, I see. We're using Ruby, but not https://github.com/awslabs/amazon-kinesis-client-ruby . We've written our own client library, which we'll probably open-source before long. LKM if yo'd like to be notified if we do.

Thanks!

PR #219, opened 2017-09-20
integrated prefetch with shardconsumer

BtXin:
Integrated prefetch config to shardconsumer.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
@@ -173,6 +173,11 @@ public class KinesisClientLibConfiguration {
     public static final ShardPrioritization DEFAULT_SHARD_PRIORITIZATION = new NoOpShardPrioritization();
 
     /**
+     * The amount of milliseconds to wait before graceful shutdown forcefully terminates.
+     */
+    public static final long DEFAULT_SHUTDOWN_GRACE_MILLIS = 5000L;
+
+    /**
      * The size of the thread pool to create for the lease renewer to use.
      */
     public static final int DEFAULT_MAX_LEASE_RENEWAL_THREADS = 20;
@@ -213,6 +218,7 @@ public class KinesisClientLibConfiguration {
     // This is useful for optimizing deployments to large fleets working on a stable stream.
     private boolean skipShardSyncAtWorkerInitializationIfLeasesExist;
     private ShardPrioritization shardPrioritization;
+    private long shutdownGraceMillis;
 
     @Getter
     private Optional<Integer> timeoutInSeconds = Optional.empty();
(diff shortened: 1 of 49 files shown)


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
```diff
@@ -1139,6 +1252,28 @@ public KinesisClientLibConfiguration withMaxGetRecordsThreadPool(final int maxGe
     }
 
     /**
+     *
```
sahilpalvia: You are missing one more config, i.e. max number of records in the cache.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ProcessTask.java
```diff
@@ -119,8 +117,38 @@ public ProcessTask(ShardInfo shardInfo, StreamConfig streamConfig, IRecordProces
                        RecordProcessorCheckpointer recordProcessorCheckpointer, KinesisDataFetcher dataFetcher,
                        long backoffTimeMillis, boolean skipShardSyncAtWorkerInitializationIfLeasesExist,
                        Optional<Integer> retryGetRecordsInSeconds, Optional<Integer> maxGetRecordsThreadPool) {
-        this(shardInfo, streamConfig, recordProcessor, recordProcessorCheckpointer, dataFetcher, backoffTimeMillis,
-                skipShardSyncAtWorkerInitializationIfLeasesExist,
+        this(shardInfo, streamConfig, recordProcessor, new SimpleRecordsFetcherFactory(streamConfig.getMaxRecords()),
```
sahilpalvia: Why are you recreating the factory, try and create the factory only once. Trying passing the client lib config around instead of passing the factory.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardConsumer.java
```diff
@@ -133,6 +135,59 @@
                   Optional<Integer> maxGetRecordsThreadPool) {
         this.streamConfig = streamConfig;
         this.recordProcessor = recordProcessor;
+        this.recordsFetcherFactory = new SimpleRecordsFetcherFactory(streamConfig.getMaxRecords());
```
sahilpalvia: Again creating multiple instances of the factory beats the purpose of creating the factory in the first place. Try and pass the client lib config around.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ProcessTask.java
BtXin: I have added new constructors to pass the config, I created new factory instance here because I thought some user might have direct dependency on this constructor, change the signature of this constructor might break their code.  If you think it is fine to get rid of this I can change it.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardConsumer.java
BtXin: Same reason here.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
BtXin: I think this is the maxRecords config, which is been configured at line 922.

sahilpalvia: That is different than max number of records in the cache. This is the number of records the cache can hold versus the maxRecords is the max records each get call can return. This is the maxRecordsCount variable in the PrefetchGetRecordsCache.

PR #221, opened 2017-09-21
Adding test classes for the Blocking and Prefetch getRecords cache.

sahilpalvia:
Changes included in the PR

* Added test classes for Blocking and Prefetch GetRecordsCache.
* KinesisDataFetcher no longer can return a null object.
* The cache now returns a ProcessRecordsInput object instead of the GetRecordsResult.
* The ProcessRecordsInput class has a new method to provide time spent in the cache.
* Calling shutdownNow instead of shutdown to shutdown the ExecutorService.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
@@ -1,3 +1,18 @@
+/*
+ *  Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Licensed under the Amazon Software License (the "License").
+ *  You may not use this file except in compliance with the License.
+ *  A copy of the License is located at
+ *
+ *  http://aws.amazon.com/asl/
+ *
+ *  or in the "license" file accompanying this file. This file is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License. 
+ */
+
 package com.amazonaws.services.kinesis.clientlibrary.lib.worker;
 
 import java.time.Instant;
@@ -46,7 +61,8 @@ public class PrefetchGetRecordsCache implements GetRecordsCache {
         this.executorService = executorService;
     }
 
-    void start() {
+    @Override
+    public void start() {
         if (!started) {
             log.info("Starting prefetching thread.");
             executorService.execute(new DefaultGetRecordsCacheDaemon());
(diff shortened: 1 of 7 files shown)


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -40,11 +42,11 @@ public PrefetchGetRecordsCache(final int maxSize, final int maxByteSize, final i
...
         this.maxRecordsCount = maxRecordsCount;
         this.getRecordsResultQueue = new LinkedBlockingQueue<>(this.maxSize);
-        prefetchCounters = new PrefetchCounters();
+        this.prefetchCounters = new PrefetchCounters();
         this.executorService = executorService;
     }
-    
-    private void start() {
+
+    void start() {
```
pfifer: This probably should be public, and in the interface.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -53,13 +55,14 @@ private void start() {
...
-    public GetRecordsResult getNextResult() {
+    public ProcessRecordsInput getNextResult() {
         if (!started) {
             start();
         }
-        GetRecordsResult result = null;
+        ProcessRecordsInput result = null;
         try {
             result = getRecordsResultQueue.take();
+            result.withCacheExitTime(Instant.now());
```
pfifer: This could be combined with the previous line.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -69,47 +72,55 @@ public GetRecordsResult getNextResult() {
...
+
     private class DefaultGetRecordsCacheDaemon implements Runnable {
         @Override
         public void run() {
             while (true) {
+                if (Thread.interrupted()) {
+                    log.warn("Prefetch thread was interrupted.");
+                    break;
+                }
                 if (prefetchCounters.byteSize < maxByteSize && prefetchCounters.size < maxRecordsCount) {
```
pfifer: It may make sense to provide the counters with the limits, and just have a method like `shouldRetrieveData()`

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -53,13 +55,14 @@ private void start() {
     }
 
     @Override
-    public GetRecordsResult getNextResult() {
+    public ProcessRecordsInput getNextResult() {
         if (!started) {
```
pfifer: I'm not a huge fan of automatically starting like this.  I prefer to require that start is called or an `IllegalStateException` is thrown.  Though it may not make much of a difference in this case.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -69,47 +72,55 @@ public GetRecordsResult getNextResult() {
...
-                        getRecordsResultQueue.put(getRecordsResult);
-                        prefetchCounters.added(getRecordsResult);
+                        ProcessRecordsInput processRecordsInput = new ProcessRecordsInput()
+                                .withRecords(getRecordsResult.getRecords())
+                                .withMillisBehindLatest(getRecordsResult.getMillisBehindLatest())
+                                .withCacheEntryTime(Instant.now());
+                        getRecordsResultQueue.put(processRecordsInput);
+                        prefetchCounters.added(processRecordsInput);
                     } catch (InterruptedException e) {
                         log.error("Interrupted while adding records to the cache", e);
```
pfifer: This isn't an error.  This probably should be info, and indicate that a shutdown was likely requested.

PR #234, opened 2017-09-27
Moving the idleMillisBetweenCalls to the cache. Fixed unit and integ …

sahilpalvia:
…tests. Calling StrategyShutdown from within the cache thread.


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/RecordsFetcherFactory.java
```diff
@@ -35,5 +35,7 @@
     void setMaxRecordsCount(int maxRecordsCount);
 
     void setDataFetchingStrategy(DataFetchingStrategy dataFetchingStrategy);
+    
+    void setIdleMillisBetweenCalls(long idleMillisBetweenCalls);
```
pfifer: Please add documentation.

PR #235, opened 2017-09-29
Handle custom metric scope

BtXin:
Added startScope and EndScope to handle custom metric scope and tested with test harness

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/GetRecordsCache.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/GetRecordsCache.java
@@ -16,7 +16,6 @@
 package com.amazonaws.services.kinesis.clientlibrary.lib.worker;
 
 import com.amazonaws.services.kinesis.clientlibrary.types.ProcessRecordsInput;
-import com.amazonaws.services.kinesis.metrics.interfaces.IMetricsScope;
 
 /**
  * This class is used as a cache for Prefetching data from Kinesis.
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/GetRecordsCache.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/GetRecordsCache.java
@@ -1,20 +1,41 @@
+/*
+ *  Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Licensed under the Amazon Software License (the "License").
+ *  You may not use this file except in compliance with the License.
+ *  A copy of the License is located at
+ *
+ *  http://aws.amazon.com/asl/
+ *
+ *  or in the "license" file accompanying this file. This file is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License. 
+ */
+
 package com.amazonaws.services.kinesis.clientlibrary.lib.worker;
 
-import com.amazonaws.services.kinesis.model.GetRecordsResult;
-
-import java.util.Collections;
+import com.amazonaws.services.kinesis.clientlibrary.types.ProcessRecordsInput;
 
 /**
  * This class is used as a cache for Prefetching data from Kinesis.
  */
 public interface GetRecordsCache {
     /**
+     * This method calls the start behavior on the cache, if available.
+     */
+    void start();
+    
+    /**
      * This method returns the next set of records from the Cache if present, or blocks the request till it gets the
      * next set of records back from Kinesis.
      * 
      * @return The next set of records.
      */
-    GetRecordsResult getNextResult();
-    
+    ProcessRecordsInput getNextResult();
+
+    /**
+     * This method calls the shutdown behavior on the cache, if available.
+     */
     void shutdown();
 }
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/GetRecordsCache.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/GetRecordsCache.java
@@ -41,6 +41,4 @@ public interface GetRecordsCache {
      * This method calls the shutdown behavior on the cache, if available.
      */
     void shutdown();
-
-    void setMetricOperation(String operation);
 }
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/GetRecordsCache.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/GetRecordsCache.java
@@ -16,6 +16,7 @@
 package com.amazonaws.services.kinesis.clientlibrary.lib.worker;
 
 import com.amazonaws.services.kinesis.clientlibrary.types.ProcessRecordsInput;
+import com.amazonaws.services.kinesis.metrics.interfaces.IMetricsScope;
 
 /**
  * This class is used as a cache for Prefetching data from Kinesis.
@@ -40,4 +41,6 @@ public interface GetRecordsCache {
      * This method calls the shutdown behavior on the cache, if available.
      */
     void shutdown();
+
+    void setMetricOperation(String operation);
 }
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/GetRecordsCache.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/GetRecordsCache.java
@@ -33,6 +33,8 @@ public interface GetRecordsCache {
      * @return The next set of records.
      */
     ProcessRecordsInput getNextResult();
+    
+    GetRecordsRetrievalStrategy getGetRecordsRetrievalStrategy();
 
     /**
      * This method calls the shutdown behavior on the cache, if available.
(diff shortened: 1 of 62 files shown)


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/RecordsFetcherFactory.java
```diff
@@ -36,4 +38,6 @@
 
     void setDataFetchingStrategy(DataFetchingStrategy dataFetchingStrategy);
 
+    void setMetricsFactory(IMetricsFactory metricsFactory);
```
pfifer: This isn't needed, instead pass the metrics factory in as part of the create call.  The ShardConumer has access to the metrics factory.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/GetRecordsCache.java
```diff
@@ -40,4 +41,6 @@
      * This method calls the shutdown behavior on the cache, if available.
      */
     void shutdown();
+
+    void setMetricOperation(String operation);
```
sahilpalvia: Doesn't need to be a method.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -138,6 +151,7 @@ public void run() {
                     log.warn("Prefetch thread was interrupted.");
                     break;
                 }
+                MetricsHelper.startScope(metricsFactory, operation);
```
sahilpalvia: Will fail if operation is not set, take the operation as a parameter to the constructor

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ProcessTask.java
```diff
@@ -148,7 +148,7 @@ public TaskResult call() {
         scope.addDimension(MetricsHelper.SHARD_ID_DIMENSION_NAME, shardInfo.getShardId());
         scope.addData(RECORDS_PROCESSED_METRIC, 0, StandardUnit.Count, MetricsLevel.SUMMARY);
         scope.addData(DATA_BYTES_PROCESSED_METRIC, 0, StandardUnit.Bytes, MetricsLevel.SUMMARY);
-
+        getRecordsCache.setMetricOperation(this.getClass().getSimpleName());
```
sahilpalvia: I wouldn't use this. I would pass the value as a String "ProcessTask"

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
BtXin: The GetRecordsCache was created inside the constructor of ShardConsumer, I don't think the shardConsumer can get the operation's name inside its constructor.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ProcessTask.java
BtXin: what's the difference between hardcode "PorcessTask" here and hardcode it inside startScope?

sahilpalvia: The difference is that of the class name changes or gets refactored, we want the metric operation to be the same. That is why hardcoding it would be better.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
sahilpalvia: The getRecordsCache gets started in the InitializeTask, which comes before the ProcessTask is called. This would cause the given line to fail.

BtXin: Initialize task and ProcessTask needs different operation name, using constructor to pass the operation would still require set the operation when ProcessTask calls it. Add the setMetricOpertion call to initialize task should be able to fix it.

sahilpalvia: The InitializeTask starts the cache. This will try to use the metrics, if the value is not set in the initialize task, it will cause the cache to keep on failing. This will lead to no records gotten from Kinesis, which in turn will lead to no calls being made to the process task.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/GetRecordsCache.java
```diff
@@ -16,6 +16,7 @@
 package com.amazonaws.services.kinesis.clientlibrary.lib.worker;
 
 import com.amazonaws.services.kinesis.clientlibrary.types.ProcessRecordsInput;
+import com.amazonaws.services.kinesis.metrics.interfaces.IMetricsScope;
```
sahilpalvia: Can you optimize the imports.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -80,8 +88,10 @@ public PrefetchGetRecordsCache(final int maxPendingProcessRecordsInput, final in
         this.getRecordsResultQueue = new LinkedBlockingQueue<>(this.maxPendingProcessRecordsInput);
         this.prefetchCounters = new PrefetchCounters();
         this.executorService = executorService;
+        this.metricsFactory = new ThreadSafeMetricsDelegatingFactory(metricsFactory);
         this.idleMillisBetweenCalls = idleMillisBetweenCalls;
         this.defaultGetRecordsCacheDaemon = new DefaultGetRecordsCacheDaemon();
+        this.operation = operation;
```
sahilpalvia: I would even check for empty strings.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/RecordsFetcherFactory.java
```diff
@@ -23,10 +25,12 @@
      *
      * @param getRecordsRetrievalStrategy GetRecordsRetrievalStrategy to be used with the GetRecordsCache
      * @param shardId ShardId of the shard that the fetcher will retrieve records for
+     * @param metricsFactory MetricsFactory used to create metricScope
+     * @param operation operation used to set metricScope
```
sahilpalvia: Operation name used for the metrics to be emitted.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
BtXin: If the string is empty do I manually assign "ProcessTask" to operation?

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -52,6 +57,8 @@
 
     private boolean started = false;
 
+    private String operation;
```
sahilpalvia: Can you make this final and remove the empty blank lines

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/SimpleRecordsFetcherFactory.java
```diff
@@ -44,7 +45,7 @@ public GetRecordsCache createRecordsFetcher(GetRecordsRetrievalStrategy getRecor
                             .setDaemon(true)
                             .setNameFormat("prefetch-cache-" + shardId + "-%04d")
                             .build()),
-                    idleMillisBetweenCalls);
+                    idleMillisBetweenCalls, metricsFactory, operation);
```
sahilpalvia: The operation doesn't need to be passed from the shardConsumer. It can be hardcoded out here.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardConsumer.java
```diff
@@ -235,7 +235,7 @@ private static final GetRecordsRetrievalStrategy makeStrategy(KinesisDataFetcher
         this.dataFetcher = kinesisDataFetcher;
         this.getRecordsCache = config.getRecordsFetcherFactory().createRecordsFetcher(
                 makeStrategy(this.dataFetcher, retryGetRecordsInSeconds, maxGetRecordsThreadPool, this.shardInfo),
-                this.getShardInfo().getShardId());
+                this.getShardInfo().getShardId(), metricsFactory, "ProcessTask");
```
sahilpalvia: Can you use this.metricsFactory instead of metricsFactory.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/SimpleRecordsFetcherFactory.java
```diff
@@ -44,7 +45,7 @@ public GetRecordsCache createRecordsFetcher(GetRecordsRetrievalStrategy getRecor
                             .setDaemon(true)
                             .setNameFormat("prefetch-cache-" + shardId + "-%04d")
                             .build()),
-                    idleMillisBetweenCalls);
+                    idleMillisBetweenCalls, metricsFactory, "processTask");
```
sahilpalvia: It should be ProcessTask and not processTask.
Add the 2 new parameters in new lines each.

PR #238, opened 2017-10-02
Shutdown that throws an exception will be retried.

jmooreoliva:
Without this change a transient error on shutdown with reason terminate prevents
child shards from starting.


BtXin:

Thanks.

jmooreoliva:
confirmed.

PR #239, opened 2017-10-03
Fixed the issue with Exception not being handled in the Prefetch cache.

sahilpalvia:
* Gave the prefetch cache a thread name
* Handling condition when shutdown is called versus exceptions being thrown
* Changed from catching errors to catching throwables
* Making prefetch a daemon thread

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
@@ -1289,12 +1289,13 @@ public class KinesisClientLibConfiguration {
 
     /**
      *
-     * @param maxCacheSize the max number of records stored in the getRecordsCache
+     * @param maxPendingProcessRecordsInput The max number of ProcessRecordsInput that can be stored in the cache before
+     *                                      blocking
      * @return this configuration object
      */
-    public KinesisClientLibConfiguration withMaxCacheSize(final int maxCacheSize) {
-        checkIsValuePositive("maxCacheSize", maxCacheSize);
-        this.recordsFetcherFactory.setMaxSize(maxCacheSize);
+    public KinesisClientLibConfiguration withMaxPendingProcessRecordsInput(final int maxPendingProcessRecordsInput) {
+        checkIsValuePositive("maxPendingProcessRecordsInput", maxPendingProcessRecordsInput);
+        this.recordsFetcherFactory.setMaxPendingProcessRecordsInput(maxPendingProcessRecordsInput);
         return this;
     }
 
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
@@ -1298,17 +1298,30 @@ public class KinesisClientLibConfiguration {
         return this;
     }
 
+    /**
+     * @param maxCacheByteSize Max byte size for the cache at any given point of time. After this threshold is crossed
+     *                         the KinesisDataFetcher will be blocked until the cache has more space available.
+     * @return KinesisClientLibConfiguration
+     */
     public KinesisClientLibConfiguration withMaxCacheByteSize(final int maxCacheByteSize) {
         checkIsValuePositive("maxCacheByteSize", maxCacheByteSize);
         this.recordsFetcherFactory.setMaxByteSize(maxCacheByteSize);
         return this;
     }
 
+    /**
+     * @param dataFetchingStrategy The strategy for fetching data from kinesis.
+     * @return KinesisClientLibConfiguration
+     */
     public KinesisClientLibConfiguration withDataFetchingStrategy(String dataFetchingStrategy) {
         this.recordsFetcherFactory.setDataFetchingStrategy(DataFetchingStrategy.valueOf(dataFetchingStrategy.toUpperCase()));
         return this;
     }
 
+    /**
+     * @param maxRecordsCount The maximum number of records in the cache, accross all ProcessRecordInput objects
+     * @return KinesisClientLibConfiguration
+     */
     public KinesisClientLibConfiguration withMaxRecordsCount(final int maxRecordsCount) {
         checkIsValuePositive("maxRecordsCount", maxRecordsCount);
         this.recordsFetcherFactory.setMaxRecordsCount(maxRecordsCount);
@@ -1334,7 +1347,7 @@ public class KinesisClientLibConfiguration {
 
     /** 
      * @param idleMillisBetweenCalls Idle time between 2 getcalls from the data fetcher.
-     * @return
+     * @return KinesisClientLibConfiguration
      */
     public KinesisClientLibConfiguration withIdleMillisBetweenCalls(long idleMillisBetweenCalls) {
         checkIsValuePositive("IdleMillisBetweenCalls", idleMillisBetweenCalls);
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
@@ -29,16 +29,16 @@ import lombok.extern.apachecommons.CommonsLog;
 
 /**
  * This is the prefetch caching class, this class spins up a thread if prefetching is enabled. That thread fetches the
- * next set of records and stores it in the cache. The size of the cache is limited by setting maxSize i.e. the maximum
- * number of GetRecordsResult that the cache can store, maxByteSize i.e. the byte size of the records stored in the
- * cache and maxRecordsCount i.e. the max number of records that should be present in the cache across multiple
- * GetRecordsResult object. If no data is available in the cache, the call from the record processor is blocked till
- * records are retrieved from Kinesis.
+ * next set of records and stores it in the cache. The size of the cache is limited by setting
+ * maxPendingProcessRecordsInput i.e. the maximum number of GetRecordsResult that the cache can store, maxByteSize
+ * i.e. the byte size of the records stored in the cache and maxRecordsCount i.e. the max number of records that should
+ * be present in the cache across multiple GetRecordsResult object. If no data is available in the cache, the call from
+ * the record processor is blocked till records are retrieved from Kinesis.
  */
 @CommonsLog
 public class PrefetchGetRecordsCache implements GetRecordsCache {
     LinkedBlockingQueue<ProcessRecordsInput> getRecordsResultQueue;
-    private int maxSize;
+    private int maxPendingProcessRecordsInput;
     private int maxByteSize;
     private int maxRecordsCount;
     private final int maxRecordsPerCall;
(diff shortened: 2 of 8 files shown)


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -50,6 +50,17 @@
 
     private boolean started = false;
 
+    /**
+     * Constructor
```
pfifer: Please provide more detail about the class here, and what it's doing.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
```diff
@@ -1298,17 +1298,30 @@ public KinesisClientLibConfiguration withMaxCacheSize(final int maxCacheSize) {
...
+     * @return KinesisClientLibConfiguration
+     */
     public KinesisClientLibConfiguration withMaxCacheByteSize(final int maxCacheByteSize) {
         checkIsValuePositive("maxCacheByteSize", maxCacheByteSize);
         this.recordsFetcherFactory.setMaxByteSize(maxCacheByteSize);
         return this;
     }
 
+    /**
+     * @param dataFetchingStrategy The strategy for fetching data from kinesis.
```
pfifer: Add the valid value, and the outcome of those values.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
```diff
@@ -1298,17 +1298,30 @@ public KinesisClientLibConfiguration withMaxCacheSize(final int maxCacheSize) {
         return this;
     }
 
+    /**
+     * @param maxCacheByteSize Max byte size for the cache at any given point of time. After this threshold is crossed
```
pfifer: Please add some documentation covering what the method does.  It should include the expectations, and outcome of the method.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
```diff
@@ -1298,17 +1298,30 @@ public KinesisClientLibConfiguration withMaxCacheSize(final int maxCacheSize) {
...
+     * @param dataFetchingStrategy The strategy for fetching data from kinesis.
+     * @return KinesisClientLibConfiguration
+     */
     public KinesisClientLibConfiguration withDataFetchingStrategy(String dataFetchingStrategy) {
         this.recordsFetcherFactory.setDataFetchingStrategy(DataFetchingStrategy.valueOf(dataFetchingStrategy.toUpperCase()));
         return this;
     }
 
+    /**
+     * @param maxRecordsCount The maximum number of records in the cache, accross all ProcessRecordInput objects
```
pfifer: This conflicts with the documentation on withMaxCacheSize

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -143,6 +156,7 @@ private void callShutdownOnStrategy() {
             if (!getRecordsRetrievalStrategy.isShutdown()) {
                 getRecordsRetrievalStrategy.shutdown();
             }
+            isShutdown = true;
```
pfifer: Where is this changed to true to trigger shutdown.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -132,7 +145,7 @@ public void run() {
                     } catch (InterruptedException e) {
                         log.info("Thread was interrupted, indicating shutdown was called on the cache.");
                         callShutdownOnStrategy();
```
pfifer: This can be moved outside the while loop.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -110,9 +121,11 @@ public void shutdown() {
...
     private class DefaultGetRecordsCacheDaemon implements Runnable {
+        private volatile boolean isShutdown = false;
+        
         @Override
         public void run() {
-            while (true) {
+            while (!isShutdown) {
                 if (Thread.currentThread().isInterrupted()) {
                     log.warn("Prefetch thread was interrupted.");
                     callShutdownOnStrategy();
```
pfifer: This can be moved outside the while loop.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -132,7 +145,7 @@ public void run() {
                     } catch (InterruptedException e) {
                         log.info("Thread was interrupted, indicating shutdown was called on the cache.");
                         callShutdownOnStrategy();
-                    } catch (Error e) {
+                    } catch (Throwable e) {
```
pfifer: When running in synchronous this can catch exceptions thrown from the SDK.  These exceptions aren't serious errors, and shouldn't be caught here.  This is specifically the last ditch catch to prevent thread death.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -132,7 +145,7 @@ public void run() {
...
                         log.error("Error was thrown while getting records, please check for the error", e);
```
pfifer: This error message should contain a lot more information, and guidance.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/RecordsFetcherFactory.java
```diff
@@ -24,18 +24,48 @@
     /**
      * Returns a records fetcher processor to be used for processing data records for a (assigned) shard.
```
pfifer: The fetcher isn't used to process data, but to retrieve it.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/RecordsFetcherFactory.java
```diff
@@ -24,18 +24,48 @@
...
      *
-     * @return Returns a record fetcher object
+     * @param getRecordsRetrievalStrategy GetRecordsRetrievalStrategy to be used with the GetRecordsCache
+     * @param shardId ShardId of the shard for which the GetRecordsCache is to be returned
```
pfifer: Probably should be the shardId that the fetcher will retrieve records for.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/RecordsFetcherFactory.java
```diff
@@ -24,18 +24,48 @@
...
+     *                
+     * @return Returns a GetRecordsCache object
      */
-    GetRecordsCache createRecordsFetcher(GetRecordsRetrievalStrategy getRecordsRetrievalStrategy);
+    GetRecordsCache createRecordsFetcher(GetRecordsRetrievalStrategy getRecordsRetrievalStrategy, String shardId);
 
+    /**
+     * This method sets the maximum number of ProcessRecordsInput objects the GetRecordsCache can hold at any give time.
```
pfifer: You don't need to include the set.  For many of the methods I would use something like
> Sets the maximum number of ProcessRecordsInput that the GetRecordsCache can hold, before further requests are blocked.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/RecordsFetcherFactory.java
```diff
@@ -24,18 +24,48 @@
...
+     * 
+     * @param maxSize Max size for the cache.
```
pfifer: Max size is misleading, be specific: the maximum number of the ProcessRecordsInput objects that the cache will accept before blocking.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/SimpleRecordsFetcherFactory.java
```diff
@@ -32,12 +34,17 @@ public SimpleRecordsFetcherFactory(int maxRecords) {
...
+    public GetRecordsCache createRecordsFetcher(GetRecordsRetrievalStrategy getRecordsRetrievalStrategy, String shardId) {
         if(dataFetchingStrategy.equals(DataFetchingStrategy.DEFAULT)) {
             return new BlockingGetRecordsCache(maxRecords, getRecordsRetrievalStrategy, idleMillisBetweenCalls);
         } else {
             return new PrefetchGetRecordsCache(maxSize, maxByteSize, maxRecordsCount, maxRecords,
-                    getRecordsRetrievalStrategy, Executors.newFixedThreadPool(1), idleMillisBetweenCalls);
+                    getRecordsRetrievalStrategy,
+                    Executors.newFixedThreadPool(1, new ThreadFactoryBuilder()
+                            .setDaemon(true)
+                            .setNameFormat("prefetch-get-records-cache-" + shardId + "-%d")
```
pfifer: You might want to shorten the name a bit, as it's going to be quite long.  Also I would recommend doing %04d

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
sahilpalvia: That description is given at the class level. Didn't want to duplicate the same comments. Will add a few lines to describe.

https://github.com/awslabs/amazon-kinesis-client/blob/prefetch/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java#L29-L36

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
sahilpalvia: Moving it to outside the loop will just check it at the start, I had kept it inside the loop just for the case if it the thread gets the interrupt at that point.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
sahilpalvia: So should we not be logging the exceptions or should we try to catch the other kinds of exceptions and then the throwable?

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -20,61 +20,68 @@
...
 /**
- * This is the default caching class, this class spins up a thread if prefetching is enabled. That thread fetches the
- * next set of records and stores it in the cache. The size of the cache is limited by setting maxSize i.e. the maximum
- * number of GetRecordsResult that the cache can store, maxByteSize i.e. the byte size of the records stored in the
- * cache and maxRecordsCount i.e. the max number of records that should be present in the cache across multiple
- * GetRecordsResult object. If no data is available in the cache, the call from the record processor is blocked till
- * records are retrieved from Kinesis.
+ * This is the prefetch caching class, this class spins up a thread if prefetching is enabled. That thread fetches the
+ * next set of records and stores it in the cache. The size of the cache is limited by setting
+ * maxPendingProcessRecordsInput i.e. the maximum number of GetRecordsResult that the cache can store, maxByteSize
```
pfifer: Remember that JavaDoc is HTML, so you can use HTML elements.  I would recommend describing the configuration like
``` html
<dl>
  <dt>maxPendingProcessRecordsInput</dt>
  <dd>The maximum number of ProcessRecordsInputs that can be in the cache at one time</dd>
  <dt>maxByteSize</dt>
  <dd>...</dd>
</dl>
```

Issue #210, opened 2017-09-14
assertAllParentShardsAreClosed check interacts poorly with high shard count dynamodb streams


zerth:
This assertion during shard sync can prevent worker initialization for workers processing dynamodb streams associated with tables having large numbers of partitions:

https://github.com/awslabs/amazon-kinesis-client/blob/1ec0b656c9c77a355640b9c32c1cbca0d592e546/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncer.java#L117

Example error message:
```
Sep 13, 2017 4:40:31 PM com.amazonaws.services.kinesis.clientlibrary.lib.worker.ShardSyncTask call
SEVERE: Caught exception while sync'ing Kinesis shards and leases
com.amazonaws.services.kinesis.clientlibrary.exceptions.internal.KinesisClientLibIOException: Parent shardId shardId-00000001505307450567-xxxxxxxx is not closed. This can happen due to a race condition between describeStream and a reshard operation.
    at com.amazonaws.services.kinesis.clientlibrary.lib.worker.ShardSyncer.assertAllParentShardsAreClosed(ShardSyncer.java:161)
    at com.amazonaws.services.kinesis.clientlibrary.lib.worker.ShardSyncer.syncShardLeases(ShardSyncer.java:117)
    at com.amazonaws.services.kinesis.clientlibrary.lib.worker.ShardSyncer.checkAndCreateLeasesForNewShards(ShardSyncer.java:88)
    at com.amazonaws.services.kinesis.clientlibrary.lib.worker.ShardSyncTask.call(ShardSyncTask.java:68)
    at com.amazonaws.services.kinesis.clientlibrary.lib.worker.MetricsCollectingTaskDecorator.call(MetricsCollectingTaskDecorator.java:49)
    at com.amazonaws.services.kinesis.clientlibrary.lib.worker.Worker.initialize(Worker.java:427)
    at com.amazonaws.services.kinesis.clientlibrary.lib.worker.Worker.run(Worker.java:356)
    at com.amazonaws.services.kinesis.multilang.MultiLangDaemon.call(MultiLangDaemon.java:111)
    at com.amazonaws.services.kinesis.multilang.MultiLangDaemon.call(MultiLangDaemon.java:58)
    at java.util.concurrent.FutureTask.run(FutureTask.java:262)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
[...]

PR #240, opened 2017-10-11
[Issue 210] - allow unexpected child shards to be ignored

zerth:
This PR, which currently lacks tests, addresses issue https://github.com/awslabs/amazon-kinesis-client/issues/210.

Now instead of always throwing an assertion if a child shard has an
open parent, consider worker configuration before doing so.  If
configured to ignore such shards, do not create leases for them during
shard sync.  This is intended to mitigate failing worker init when
processing dynamodb streams with many thousands of shards (which can
happen for tables with thousands of partitions).

This new behavior can be enabled by adding the following to a
configuration/properties file:

```
ignoreUnexpectedChildShards = true
```


pfifer:
Thanks for submitting this, just a couple of requests:

Can you update this with the new changes, and remove the white space changes?  I would also like to see some tests to verify that it doesn't break the existing behavior, and tests for the new behavior.

zerth:
Will do.

zerth:
@pfifer: Branch has been updated and unrelated whitespace changes removed, and two basic testcases added which exercise the new behavior.  Looking at the existing tests, it seems to me they adequately cover the default behavior.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
```diff
@@ -351,6 +359,7 @@ public KinesisClientLibConfiguration(String applicationName,
                                          long parentShardPollIntervalMillis,
                                          long shardSyncIntervalMillis,
                                          boolean cleanupTerminatedShardsBeforeExpiry,
+                                         boolean ignoreUnexpectedChildShards,
```
pfifer: You're changing a public constructor and adding a parameter in the middle of it.  This will break anyone who uses the constructor.  I would prefer to add a new constructor with the parameter.  The other option is not to have the parameter, and use the with/set operations to configure the feature.

At some point we will look into switching to a builder pattern for the configuration.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncer.java
```diff
@@ -149,19 +161,39 @@ private static synchronized void syncShardLeases(IKinesisProxy kinesisProxy,
...
+     * @throws KinesisClientLibIOException
+     */
+    private static void assertAllParentShardsAreClosed(Set<String> inconsistentShardIds)
+        throws KinesisClientLibIOException {
+        if (!inconsistentShardIds.isEmpty()) {
+            String ids = "";
+            for (String id : inconsistentShardIds) {
+                ids += " " + id;
+            }
+            throw new KinesisClientLibIOException(String.valueOf(inconsistentShardIds.size()) + " open child shards (" + ids + ") are inconsistent."
```
pfifer: This might be cleaner using String.format.

zerth:
@pfifer: most recent comments addressed.  I removed the config constructor changes to avoid polluting that module (and since my own usage involves only `.properties` files).

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncer.java
```diff
@@ -149,19 +161,40 @@ private static synchronized void syncShardLeases(IKinesisProxy kinesisProxy,
 
     /** Helper method to detect a race condition between fetching the shards via paginated DescribeStream calls
      * and a reshard operation.
+     * @param inconsistentShardIds
+     * @throws KinesisClientLibIOException
+     */
+    private static void assertAllParentShardsAreClosed(Set<String> inconsistentShardIds)
+        throws KinesisClientLibIOException {
+        if (!inconsistentShardIds.isEmpty()) {
+            String ids = "";
```
pfifer: You can use [StringUtils#join](https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringUtils.html) here instead of manually building the string.

zerth: Updated.  I don't normally use Java as has no doubt been apparent. :-)

pfifer:
Please confirm that we can use, modify, copy, and redistribute this contribution. Thanks.

zerth:
> Please confirm that we can use, modify, copy, and redistribute this contribution.

I confirm this to be true.

ivoanjo:
Hey there, and sorry for necro'ing 😆

I got here because we were getting bitten by this exact same issue, and AWS support pointed us at https://github.com/awslabs/amazon-kinesis-client/issues/210 which did solve our issue.

But I'm really curious -- @pfifer can you comment on why this setting is not the default? E.g. when would it be relevant for me to care about this exception?

Thanks a lot!

Issue #147, opened 2017-03-13
is there a way to check whether a Worker is ready for business?


elisehuard:
For our tests, we'd like to check whether a worker is ready to start consuming before we start our tests producing events - programatically.  We can't see a way in the code (no attributes are being set) - did we miss something, is there a way to do this?
If not, is there a plan to do this, or if not, would you be open to a pull request?

mantree:
Created a PR that adds a flag for this:

https://github.com/awslabs/amazon-kinesis-client/pull/148

PR #245, opened 2017-10-17
Adding initialized boolean to record worker state

NorTheHypnos:


NorTheHypnos:
Added a unit test for https://github.com/awslabs/amazon-kinesis-client/issues/147

pfifer:
Out of curiosity what is the use case of this?

NorTheHypnos:
Same as mention in [148](https://github.com/awslabs/amazon-kinesis-client/pull/148).

We have a bunch of integration tests that fail sporadically mainly because it takes a long time for the  KCL worker to start and the tests time out. With this flag we can wait for the Worker to be initialized and ready before we start the tests.

Dak0ta:
This would be beneficial to my client application as well.

Dak0ta:
I think the changes from issue #291 includes in 1.9.1 that allow you to subscribe to the states of the worker initialization likely make this change unnecessary.

https://github.com/awslabs/amazon-kinesis-client/pull/291/files

Issue #248, opened 2017-10-19
Check lease owner when deciding if a lease is a candidate for cleanup


johnsonwj:
Currently, `com.amazonaws.services.kinesis.clientlibrary.lib.worker.ShardSyncer` does not examine the `leaseOwner` attribute in the lease record to determine if it should be cleaned up (relevant method: `isCandidateForCleanup`).

Presumably the `ShardSyncer` should ignore any leases that do not belong to it -- i.e. leases for which the `leaseOwner` does not match the id of the worker that called `syncShardLeases`.

From examining the code, `ShardSyncer` and its friend `ShardSyncTask` do not care about the worker or app ids at all. If instead they belonged to a particular worker / application, then we could maintain only one Dynamo table to track leases across multiple input streams.

johnsonwj:
Alternatively, allow clients to set a custom `ILeaseManager` for the worker, allowing us to implement `listLeases` in such a way that only leases owned by that worker/app id are listed as candidates for cleanup.

Currently none of the public constructors for `Worker` take an `ILeaseManager`, and the builder always constructs its own.

johnsonwj:
Second alternative: update `com.amazonaws.services.kinesis.leases.impl.LeaseManager` so that **its** implementation of `listLeases` behaves as mentioned in the previous comment.

PR #250, opened 2017-10-19
Allow specifying a custom LeaseManager in Worker.Builder

johnsonwj:
This is the "do-it-yourself" workaround for #248


johnsonwj:
I believe the build failure is due to a preexisting build problem on master...

Issue #211, opened 2017-09-15
Application didn't checkpoint at end of shard


posac:
Hi, 

I have issue with checkpoints  and ending shards. 
In my processor, after each processRecords call I am calling checkpoint with sequence number. Sequence number is also stored in processor field. 

On shutdown I am also calling : 
      if (shutdownInput.getShutdownReason() == ShutdownReason.TERMINATE) {
            checkpoint(shutdownInput.getCheckpointer());
        }

But after "Reached end of shard " I got exception:
`java.lang.IllegalArgumentException: Application didn't checkpoint at end of shard shardId-00000001505383995695-aa7bbe9c
	at com.amazonaws.services.kinesis.clientlibrary.lib.worker.ShutdownTask.call(ShutdownTask.java:107) ~[amazon-kinesis-client-1.8.1.jar:?]`

I have debug your code and for me there is bug in : RecordProcessorCheckpointer.java in method advancePosition. 

I believe that in condition : 
` if (extendedSequenceNumber != null && !extendedSequenceNumber.equals(lastCheckpointValue)) {`

should be : 
`if (checkpointToRecord != null && !checkpointToRecord.equals(lastCheckpointValue)) {`

Other wise I am not able to store SHARD_END - cause when processRecord calls checkpoint in checkpoint there is no sequenceNumberAtShardEnd yet, It is when it is called from shutdown but then mentioned condition is not satisfy -> lastCheckpointValue is equal to extendedSequenceNumber

lskrajny:
+1

sahilpalvia:
This issue seems to happen with prepareCheckpoint method. This issue was fixed with the [PR#301](https://github.com/awslabs/amazon-kinesis-client/pull/301). The release for the new version of the KCL is TBD.

klesniewski:
I have also encountered this problem. It is annoying, as when it happens the worker cannot progress causing service downtime (it doesn't crash, but it stops working as expected). It may happen when instance of `IRecordProcessor` checkpoints specific records (sequence numbers) instead of using `IRecordProcessorCheckpointer.checkpoint()`.

Here is how it is exhibited in logs:

```
2018-07-20 22:00:20,469 DEBUG      StreamsRecordProcessor> Processing 1 records
2018-07-20 22:00:20,480  INFO KafkaStreamsRecordProcessor> KafkaSendCompleted to 'devices-ingress' at 14:a4
[...]

PR #261, opened 2017-11-02
Issue #211 Application didn't checkpoint at end of shard

posac:


sahilpalvia:
This PR can't be accepted. Blanket checkpointing around shard end could be dangerous. Children shard get picked up for processing only when the parent has been marked as SHARD_END. This could lead into a situation where the parent has been shutdown because of shard end and the children aren't picked up for processing due to the checkpoint of the parent not being SHARD_END leading to data loss.

PR #263, opened 2017-11-08
Fix for the lease losses in the PrefetchCache and AsyncGetRecordsStrategy caused due to ExpiredIteratorException

sahilpalvia:
This commit includes
* Fix for ExpiredIteratorException in AsyncGetRecordsStrategy and PrefetchGetRecordsCache.
* PrefetchGetRecordsCache can restart the iterator, which fetches a new iterator.
* Unit tests for the changes.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
@@ -23,10 +23,13 @@ import java.util.concurrent.LinkedBlockingQueue;
 import org.apache.commons.lang.Validate;
 
 import com.amazonaws.SdkClientException;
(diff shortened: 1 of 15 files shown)


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PrefetchGetRecordsCache.java
```diff
@@ -162,6 +171,14 @@ public void run() {
                         prefetchCounters.added(processRecordsInput);
                     } catch (InterruptedException e) {
                         log.info("Thread was interrupted, indicating shutdown was called on the cache.");
+                    } catch (ExpiredIteratorException e) {
+                        log.info(String.format("ShardId %s: getRecords threw ExpiredIteratorException - restarting"
+                                + " after greatest seqNum passed to customer", shardId), e);
```
pfifer: This isn't actually true, it restarting from the last accepted sequence number.  That may be in the queue waiting to be processed.

PR #274, opened 2018-01-05
Added IKinesisProxy injector in Worker.Builder

parijatsinha:
Added IKinesisProxy injector in Worker.Builder to allow injecting custom proxy implementations

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
@@ -35,7 +35,6 @@ import java.util.concurrent.TimeoutException;
 import com.amazonaws.services.kinesis.clientlibrary.proxies.IKinesisProxy;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import lombok.Getter;
 
 import com.amazonaws.regions.Region;
 import com.amazonaws.regions.RegionUtils;
@@ -1064,7 +1063,6 @@ public class Worker implements Runnable {
     /**
      * Builder to construct a Worker instance.
      */
-    @Getter
     public static class Builder {
 
         private IRecordProcessorFactory recordProcessorFactory;
@@ -1269,7 +1267,7 @@ public class Worker implements Runnable {
                 }
             }
             if (metricsFactory == null) {
-                metricsFactory = Worker.getMetricsFactory(cloudWatchClient, config);
+                metricsFactory = getMetricsFactory(cloudWatchClient, config);
             }
             if (shardPrioritization == null) {
                 shardPrioritization = new ParentsFirstShardPrioritization(1);
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
@@ -35,6 +35,7 @@ import java.util.concurrent.TimeoutException;
 import com.amazonaws.services.kinesis.clientlibrary.proxies.IKinesisProxy;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import lombok.Getter;
 
 import com.amazonaws.regions.Region;
 import com.amazonaws.regions.RegionUtils;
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -1192,6 +1194,19 @@ public Builder shardPrioritization(ShardPrioritization shardPrioritization) {
             return this;
         }
 
+        /**
+         * Set KinesisProxy for the worker.
+         *
+         * @param kinesisProxy
+         *            KinesisProxy uses the AmazonKinesis client to get data from Kinesis or DynamoDBStreams
```
sahilpalvia: Can you fix this comment, it would be confusing for the customers since KCL already has DDB endpoints. Can you remove DynaoDBStreams from the comment.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -1252,18 +1269,20 @@ public Worker build() {
                 }
             }
             if (metricsFactory == null) {
-                metricsFactory = getMetricsFactory(cloudWatchClient, config);
+                metricsFactory = Worker.getMetricsFactory(cloudWatchClient, config);
```
sahilpalvia: Why is this being changed? The getMetricsFactory method is a static method which is accessible within the same class.

src/test/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/WorkerTest.java
```diff
@@ -1474,6 +1484,71 @@ public void testEndOfShardAfterShutdownRequest() throws Exception {
...
+    @Test
+    public void testBuilderWithDefaultKinesisProxy() {
+        IRecordProcessorFactory recordProcessorFactory = mock(IRecordProcessorFactory.class);
+
+        Worker.Builder builder = new Worker.Builder()
+            .recordProcessorFactory(recordProcessorFactory)
+            .config(config);
+        builder.build();
+        assertNotNull(builder.getKinesisProxy());
+        assertTrue(builder.getKinesisProxy() instanceof KinesisProxy);
```
sahilpalvia: This is not the correct way to test the proxy that is set. We need to make sure that the worker is getting the proxy that we expect it to use.

src/test/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/WorkerTest.java
```diff
@@ -1474,6 +1484,71 @@ public void testEndOfShardAfterShutdownRequest() throws Exception {
...
+            .recordProcessorFactory(recordProcessorFactory)
+            .kinesisProxy(kinesisProxy)
+            .config(config);
+        builder.build();
+        assertNotNull(builder.getKinesisProxy());
+        assertTrue(builder.getKinesisProxy() instanceof DummyKinesisProxy);
+    }
+
+    // KinesisProxyImplementation to test KinesisProxy injection.
+    private class DummyKinesisProxy implements IKinesisProxy {
```
sahilpalvia: We should be able to use [KinesisLocalFileProxy](https://github.com/awslabs/amazon-kinesis-client/blob/master/src/test/java/com/amazonaws/services/kinesis/clientlibrary/proxies/KinesisLocalFileProxy.java) instead of creating a new one. Let me get back to you on that one.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -1062,6 +1064,7 @@ private static ExecutorService getExecutorService() {
     /**
      * Builder to construct a Worker instance.
      */
+    @Getter
```
sahilpalvia: There is no use for the builder to have a getter, we should be testing what the builder has created, if that has the correct value.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
parijatsinha: It was changed because the Lombok Getter had the same name as the static method.

PR #278, opened 2018-01-10
Upgraded commons-lang dependency to commons-lang3

jmgrimes:
Upgraded commons-lang to commons-lang3 since the old 2.6 branch appears all but abandoned (targeting Java 1.2) and the 3.x line targets 1.6+.

Simple drop in replacement with package change, reran tests with no failures.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,9 +58,9 @@
       <version>2.6.1</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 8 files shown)


pfifer:
Thanks for submitting this, and sorry for not getting to this. 

We will complete this in #406

PR #292, opened 2018-02-02
Removed not needed assertNotNull before assertTrue for instanceof

lc-nyovchev:
https://stackoverflow.com/questions/2950319/is-null-check-needed-before-calling-instanceof


lc-nyovchev:
As it was a small cosmetic change, no need to keep this one open

Issue #280, opened 2018-01-12
Worker.Builder tries to mutate possibly-immutable Dynamo client


mrhota:
We build a `Worker` using code like this:
```
worker = new Worker.Builder().recordProcessorFactory(processorFactory)
                             .config(kinesisClientLibConfiguration)
                             .dynamoDBClient(dynamoDBClient)
                             .kinesisClient(kinesisClient)
                             .build();
```

The provided `dynamoDBClient` is constructed using the _recommended_ builder pattern. The dynamo client's `build()` method sets the resulting client as immutable, which results in this exception from the `Worker.Builder`.

```
Exception in thread "main" java.lang.UnsupportedOperationException: Client is immutable when created with the builder.
    at com.amazonaws.AmazonWebServiceClient.checkMutability(AmazonWebServiceClient.java:843)
    at com.amazonaws.AmazonWebServiceClient.setRegion(AmazonWebServiceClient.java:349)
    at com.amazonaws.services.kinesis.clientlibrary.lib.worker.Worker$Builder.build(Worker.java:921)
...
```

The `Worker.Builder` should check for dynamo mutability before trying to `setRegion` on it.

pfifer:
Thanks for reporting this, we're looking into it.

mrhota:
@pfifer any update here?

mrhota:
i ask because it seems like we can't migrate away from the deprecated amazon dynamo db client ctors until this is resolved. Is there a better way to build a `Worker.Builder` that I'm missing?

pfifer:
A fix is coming for this soon.

mrhota:
ok, cool.

In the meantime, for future visitors, I think providing a `KinesisClientLibConfiguration` lacking endpoint and region name properties for kinesis, dynamo, et al will avoid the `UnsupportedOperationException`. Clients have to be configured carefully to ensure their endpoints/regions match what the config's properties would otherwise have.

sahilpalvia:
Try to setup the immutable client with the endpoint/region required and don't pass it via the KinesisClientLibraryConfiguration. The worker tries to set it only when it gets it from the config. The fix for this issue is in progress.

smartYi:
Hey, is this fixed already? I am using:         
<dependency>
            <groupId>com.amazonaws</groupId>
            <artifactId>amazon-kinesis-client</artifactId>
            <version>1.9.0</version>
        </dependency>

sdhanavath:
Hi!

[...]

PR #305, opened 2018-03-05
Fix for Worker to handle immutable clients

sahilpalvia:
*Issue #280*

*Description of changes:*
* This PR fixes the issue with the Worker trying to set the region/endpoint for an immutable client.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
@@ -228,6 +228,80 @@ public class Worker implements Runnable {
                 execService);
     }
 
+    // Backwards compatible constructors
+    /**
+     * This constructor is for binary compatibility with code compiled against version of the KCL that only have
+     * constructors taking "Client" objects.
+     *
+     * @param recordProcessorFactory
+     *            Used to get record processor instances for processing data from shards
+     * @param config
+     *            Kinesis Client Library configuration
+     * @param kinesisClient
+     *            Kinesis Client used for fetching data
+     * @param dynamoDBClient
+     *            DynamoDB client used for checkpoints and tracking leases
+     * @param cloudWatchClient
+     *            CloudWatch Client for publishing metrics
+     */
+    public Worker(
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -1101,6 +1096,46 @@ private static ExecutorService getExecutorService() {
         return new WorkerThreadPoolExecutor(threadFactory);
     }
 
+    private static void setRegionForClient(@NonNull final Object client, @NonNull final String region) {
```
pfifer: You might be able to simplify all of this to something like:
``` java
private <S, T> void trySetField(S source, String field, Consumer<T> f, T value) {
        try {
            f.accept(value);
        } catch (UnsupportedOperationException uoe) {
            LOG.debug("Write not supported on " + source.getClass().getSimpleName() + " for " + field);
        }
    }
```
called from the constructor like
``` java
trySetField(kinesisClient, "region", kinesisClient::setRegion, region);
trySetField(dynamoDBClient, "region", dynamoDBClient::setRegion, region);
```

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -1292,5 +1337,30 @@ public Worker build() {
...
+                        + region + ".");
+                builder.withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration(endpointUrl, region));
+            } else if (StringUtils.isNotEmpty(region)) {
+                LOG.warn("Received configuration for region as " + region + ".");
+                builder.withRegion(region);
+            } else {
+                LOG.warn("No configuration received for endpoint and region, will default region to us-east-1");
+                builder.withRegion(Regions.US_EAST_1);
+            }
+            return (AmazonWebServiceClient) builder.build();
```
pfifer: You might be able to get away without the cast if change the signature to something like:
``` java
<R, T extends AwsClientBuilder<T, R>> R setupClient(T builder, AWSCredentialsProvider creds ...) {
    return builder.withCredentials(creds).build();
}
```

Issue #312, opened 2018-03-15
Feature? Expose exceptions in the Worker to the caller to report up or take action on


Dak0ta:
Currently any exceptions that occur in `initialize()` or `runProcessLoop()` for example are unavailable to the caller of the thread.

So if there are credential issues, or permission issues, connection problems, etc... they are all logged trough slf4j which is great, but I can't capture them and do something with them.

lastException is something that's used during `initialize()`, even having access to get that from the Worker and check if it's != null if you so choose would be a great start, but ideally it would be great to be able to pass some sort of custom exception handling implementation into the the `KinesisClientLibConfiguration` itself.

If there's already a way to get this I apologize, but the try { } catch { } blocks within Worker look to be pretty standard with no ability to get them out.

PR #315, opened 2018-03-15
Make lastException from initialize() available

Dak0ta:
Issue #312

During each run of initialize() an exception could be encountered, this makes that exception available via a getter to whatever created the thread to be able to take further action with it.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
@@ -104,6 +104,7 @@ public class Worker implements Runnable {
     private volatile boolean shutdown;
     private volatile long shutdownStartTimeMillis;
     private volatile boolean shutdownComplete = false;
+    private volatile Exception lastException = null;
 
     // Holds consumers for shards the worker is currently tracking. Key is shard
     // info, value is ShardConsumer.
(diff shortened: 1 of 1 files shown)


Dak0ta:
@pfifer - Just curious what the acceptance criteria is for contributions and what kind of timelines are involved?

Dak0ta:
Bump!

@pfifer - Just curious what the acceptance criteria is for contributions and what kind of timelines are involved?

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -881,6 +881,14 @@ boolean isShutdownComplete() {
     WorkerStateChangeListener getWorkerStateChangeListener() {
         return workerStateChangeListener;
     }
+    
+    /**
+     * The last of any exception encountered during initialize() will be available here
+     */
+    public Exception getLastException()
```
pfifer: Can you match the existing code style?

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -104,6 +104,7 @@
     private volatile boolean shutdown;
     private volatile long shutdownStartTimeMillis;
     private volatile boolean shutdownComplete = false;
+    private volatile Exception lastException = null;
```
pfifer: Can you name this something like `lastInitializationException`?

pfifer:
Thanks for providing this, just a couple of small changes should work.  I would prefer to integrate this with the worker state change listener that was added, but that will be a larger change.

Issue #341, opened 2018-05-25
Enable property option to use AT_TIMESTAMP shard iterator in MultiLangDaemon clients


marcinc:
Since support for Time based shard iterators was added in https://github.com/awslabs/amazon-kinesis-client/pull/94, it would be good to enable option to pass initial timestamp for `AT_TIMESTAMP` iterator from properties file. This would enable Multilang Daemon clients to take advantage of this iterator type. This was previously requested in https://github.com/awslabs/amazon-kinesis-client/issues/145.

It looks like it could be achieved by simply adding appropriate Date parameter type decoder in `KinesisClientLibConfigurator` which would then allow use of the following property: 
```
# app.properties

timestampAtInitialPositionInStream = 1527254694
```
Internally this would result in correct setting of `initialPositionInStream` to `AT_TIMESTAMP` and its initial timestamp to  `1527254694`.

pfifer:
This should be fixed in 1.9.2 and should be available in Maven over the next 24 hours.

Feel free to reopen if you run into the issue again.

PR #342, opened 2018-05-25
Enables property for setting AT_TIMESTAMP shard iterator initial time…

marcinc:
…stamp (#341)

*Issue #341*

*Description of changes:*

Added `DatePropertyValueDecoder` for use in `KinesisClientLibConfigurator` which enables `timestampAtInitialPositionInStream` property for setting initial timestamp for `AT_TIMESTAMP` shard iterator.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/config/DatePropertyValueDecorder.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/config/DatePropertyValueDecorder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Amazon Software License (the "License").
  * You may not use this file except in compliance with the License.
@@ -35,7 +35,11 @@ class DatePropertyValueDecoder implements IPropertyValueDecoder<Date> {
      */
     @Override
     public Date decodeValue(String value) {
-        return new Date(Long.parseLong(value) * 1000L);
+        try {
+            return new Date(Long.parseLong(value) * 1000L);
+        } catch (NumberFormatException e) {
+            throw new IllegalArgumentException("Date property value must be numeric.");
+        }
     }
 
     /**
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/config/DatePropertyValueDecorder.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/config/DatePropertyValueDecorder.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2014 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Amazon Software License (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ * http://aws.amazon.com/asl/
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazonaws/services/kinesis/clientlibrary/config/DatePropertyValueDecorder.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2014 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pfifer: Can you update this to 2018?

pfifer:
Could you also add a basic unit test?  

Thanks

marcinc:
Hi @pfifer. Fixup commit added with update header and tests.

marcinc:
Hi @pfifer. Please let me know if there is anything else I should address in this PR. Otherwise, if you're happy with the change, it would be great to get it merged. Thanks.

src/test/java/com/amazonaws/services/kinesis/clientlibrary/config/DatePropertyValueDecoderTest.java
```diff
@@ -0,0 +1,68 @@
...
+    private static final String TEST_VALUE = "1527267472";
+
+    @Test
+    public void testNumericValue() {
+        Date timestamp = decoder.decodeValue(TEST_VALUE);
+        assertEquals(timestamp.getClass(), Date.class);
+        assertEquals(timestamp, new Date(Long.parseLong(TEST_VALUE) * 1000L));
+    }
+
+    @Test
```
pfifer: Minor: You can do 
``` java
@Test(expected = IllegalArgumentException.class)`
```
which would remove the requirement for the try/catch.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/config/DatePropertyValueDecorder.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * @param value property value as String
+     * @return corresponding variable in correct type
+     */
+    @Override
+    public Date decodeValue(String value) {
+        try {
+            return new Date(Long.parseLong(value) * 1000L);
```
pfifer: Just calling this out for myself, and any future readers: 

If someone accidentally uses milliseconds this won't work the way they expect.  In the future it might make sense to use Instant from java.time.  Instant#parse can deal with textual date representations, and can be converted to a java.util.Date pretty easily.

src/test/java/com/amazonaws/services/kinesis/clientlibrary/config/KinesisClientLibConfiguratorTest.java
```diff
@@ -189,6 +204,49 @@ public void testWithInitialPositionInStreamVariables() {
...
+    }
+
+    @Test
+    public void testWithNonNumericTimestampAtInitialPositionInStreamVariables() {
+        KinesisClientLibConfiguration config =
+                getConfiguration(StringUtils.join(new String[] {
+                        "streamName = a",
+                        "applicationName = b",
+                        "AWSCredentialsProvider = ABCD," + credentialName1,
+                        "timestampAtInitialPositionInStream = 123abc"
```
pfifer: Calling this out for myself, and future readers: 

Looking at the code this will at least log a warning.  That isn't part of your code, but I don't know if that is a good outcome.  If the InitialPositionInStream is set to AT_TIMESTAMP then the KCL will be unable to startup if the lease table doesn't exist or isn't populated.  Otherwise the KCL should be able to startup, but that could be really confusing.

pfifer:
Thanks for providing the change.  I'm going to start the process for accepting this on my side.  

For my own information which of the MultiLang clients do you use?

src/test/java/com/amazonaws/services/kinesis/clientlibrary/config/DatePropertyValueDecoderTest.java
marcinc: Good point. Fixed.

marcinc:
> For my own information which of the MultiLang clients do you use?

I'm using ruby client for the time being.

PR #344, opened 2018-06-06
Added cache updating behavior for GetShard

pfifer:
Customer are occasionally seeing messages about being unable to
retrieve shard information, which is logged as a warning.  This change
will allow the shard map to be updated even when there is no re-shard
operation.

This now triggers a shard list update if there is 1000 cache misses,
or a cache miss occurs when the cache is more than 30 seconds old.
For Kinesis the updates will use ListShards, and for DynamoDB Streams
it will continue to use DescribeStream.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/proxies/KinesisProxy.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/proxies/KinesisProxy.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  *  Licensed under the Amazon Software License (the "License").
  *  You may not use this file except in compliance with the License.
@@ -15,13 +15,19 @@
 package com.amazonaws.services.kinesis.clientlibrary.proxies;
 
 import java.nio.ByteBuffer;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.EnumSet;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazonaws/services/kinesis/clientlibrary/proxies/KinesisProxy.java
```diff
@@ -333,19 +354,73 @@ private ListShardsResult listShards(final String nextToken) {
      */
     @Override
     public Shard getShard(String shardId) {
-        if (this.listOfShardsSinceLastGet.get() == null) {
-            //Update this.listOfShardsSinceLastGet as needed.
-            this.getShardList();
+        if (this.cachedShardMap == null) {
+            synchronized (this) {
+                if (this.cachedShardMap == null) {
```
sahilpalvia: Does it need to be this. ? Can do without it.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/proxies/KinesisProxy.java
```diff
@@ -333,19 +354,73 @@ private ListShardsResult listShards(final String nextToken) {
...
+            if (cacheMisses.incrementAndGet() > MAX_CACHE_MISSES_BEFORE_RELOAD || cacheNeedsTimeUpdate()) {
+                synchronized (this) {
+                    shard = cachedShardMap.get(shardId);
+
+                    //
+                    // If after synchronizing we resolve the shard, it means someone else already got it for us.
+                    //
+                    if (shard == null) {
+                        LOG.info("To many shard map cache misses or cache is out of date -- forcing a refresh");
+                        this.getShardList();
```
sahilpalvia: Same as above can do without this.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/proxies/KinesisProxy.java
```diff
@@ -393,9 +468,12 @@ public Shard getShard(String shardId) {
                 }
             } while (response.getStreamDescription().isHasMoreShards());
         }
-        this.listOfShardsSinceLastGet.set(shardIterationState.getShards());
+        List<Shard> shards = shardIterationState.getShards();
+        this.cachedShardMap = shards.stream().collect(Collectors.toMap(Shard::getShardId, Function.identity()));
+        this.lastCacheUpdateTime = Instant.now();
```
sahilpalvia: Can this be a separate method. The method says getShardList, lets scope it to doing just that.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/proxies/KinesisProxy.java
pfifer: It's a double checked lock to ensure that if someone gets the lock before us, and loads the shard map we don't reload it.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/proxies/KinesisProxy.java
pfifer: It's he same thing as above.  If another processor got in first, don't redo their work.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/proxies/KinesisProxy.java
pfifer: The caching behavior allows the getShard to effectively piggy back on getShardList.  We really want that to prevent additional scans of the shard map.  We also can't change the method as it's part of the interface.

PR #345, opened 2018-06-06
Application didn't checkpoint at end of shard

versemonger:
Issue #211 Application didn't checkpoint at end of shard

*Description of changes:*

Decide whether to checkpoint based on whether `checkpointToRecord` is equal to `lastCheckPointValue`.
- If a shard is open, this makes no difference because `checkpointToRecord` is always equal to `extendedSequenceNumber` .
- If a shard is closed, `sequenceNumberAtShardEnd`  is set, and we are not trying to checkpoint at the last record, this doesn't make a difference, either becuase `checkpointToRecord` is equal to `extendedSequenceNumber`
- When a shard is closed, `sequenceNumberAtShardEnd`  is set, and we are trying to checkpoint at the last record, then  `extendedSequenceNumber` is different from `checkpointToRecord`. If `lastCheckpointValue` it not at the last record, then we could checkpoint value of `checkpointToRecord` because `extendedSequenceNumber` is not equal to `lastCheckpointValue` ; but if `lastCheckpointValue` is at the last reocrd, we couldn't checkpoint becuase it is equal to `extendedSequenceNumber` now.

This patch fixes the issue and this unit test shows a case where the patch helps.


sahilpalvia:
This behavior is dangerous as explained above. Closing the PR, feel free to reopen if needed.

PR #369, opened 2018-08-15
Introduce initialization exception handler in KCL V2

jiaxul:
*Issue #, if available:*

Introduce initialization exception handler in KCL V2.

KCL scheduler does not throw any exceptions that occurs during processing a stream. It just logs it as an error. 

*Description of changes:*
add a CoordinatorExceptionHandler interface
add CoordinatorExceptionHandler as a parameter in CoordinatorConfig 
add a NoOpCoordinatorExceptionHandler to pass in as default so it does not change the behavior of the Scheduler class
The client applications can implement the CoordinatorExceptionHandler and pass in CoordinatorConfig so that they can specify how they want to handle the exceptions during processing the streams. 


amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/NoOpCoordinatorExceptionHandler.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ *  or in the "license" file accompanying this file. This file is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package software.amazon.kinesis.coordinator;
+
+public class NoOpCoordinatorExceptionHandler implements CoordinatorExceptionHandler {
```
sahilpalvia: A better name would be PassThroughCoordinatorExceptionHandler.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/CoordinatorExceptionHandler.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ *  or in the "license" file accompanying this file. This file is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package software.amazon.kinesis.coordinator;
+
+/**
```
sahilpalvia: Could you add documentation

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/CoordinatorExceptionHandler.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ */
+public interface CoordinatorExceptionHandler {
+    void propogate(Exception e);
```
sahilpalvia: Same as above, some documentation.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/NoOpCoordinatorExceptionHandler.java
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * Empty constructor for NoOp Exception Handler
+     */
+    public NoOpCoordinatorExceptionHandler() {
```
sahilpalvia: You could use @NoArgumentConstructor from lombok

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -193,20 +195,26 @@ public void run() {
...
             initialize();
             log.info("Initialization complete. Starting worker loop.");
         } catch (RuntimeException e) {
             log.error("Unable to initialize after {} attempts. Shutting down.", lifecycleConfig.maxInitializationAttempts(), e);
             shutdown();
+            exceptionToThrow = e;
         }
 
         while (!shouldShutdown()) {
+            exceptionToThrow = null;
```
sahilpalvia: Need not set this. If exception is caught from initialize, shutdown is invoked and shouldShutdown will return true and the code should not enter this loop.

jiaxul: Will delete

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/NoOpWorkerStateChangeListener.java
```diff
@@ -27,4 +27,9 @@ public NoOpWorkerStateChangeListener() {
 	public void onWorkerStateChange(WorkerState newState) {
 
 	}
+
+	@Override
+	public void onInitializationError(Exception e) {
```
pfifer: This should be a default on the interface.
Would prefer the name to be onInitializationException

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -193,12 +193,14 @@ public void run() {
             return;
         }
 
+        Exception exceptionToThrow = null;
```
pfifer: This doesn't feel like the right place for this.  Should we instead put onInitializationException at the point that the exception was thrown?
This just reports the last exception, which means a cascaded failure may be misreported.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -207,6 +209,10 @@ public void run() {
 
         finalShutdown();
         log.info("Worker loop is complete. Exiting from worker.");
+
+        if (exceptionToThrow != null) {
```
pfifer: I don't feel this is the correct location to report the exception.  We should report as near as possible to where the exception originally occurred.  This would be important if the user was also attempting to create a stack trace for the original call site.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/WorkerStateChangeListener.java
```diff
@@ -27,4 +26,6 @@
 	}
 
 	void onWorkerStateChange(WorkerState newState);
+
+	void onInitializationError(Exception e);
```
pfifer: This is a breaking change, and must be defaulted.

I also want to think more about how state reporting and actions happen.  Finally this should accept a `Throwable` vs an `Exception`.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -199,6 +199,7 @@ public void run() {
         } catch (RuntimeException e) {
             log.error("Unable to initialize after {} attempts. Shutting down.", lifecycleConfig.maxInitializationAttempts(), e);
             shutdown();
+            workerStateChangeListener.onInitializationFailure(e);
```
sahilpalvia: It would be better to invoke the handler before calling shutdown. The handler is called as a reaction of the exception. It needs to be invoked as soon as the exception is thrown before performing any other action.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/WorkerStateChangeListener.java
```diff
@@ -27,4 +27,7 @@
 	}
 
 	void onWorkerStateChange(WorkerState newState);
+
+	default void onInitializationFailure(Throwable e) {
```
sahilpalvia: Can you rename the method to something like `allInitializationAttemptsFailed(Throwable t)`, indicating that the handler is invoked after all the initialization attempt have failed.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -207,6 +208,7 @@ public void run() {
 
         finalShutdown();
         log.info("Worker loop is complete. Exiting from worker.");
+
```
sahilpalvia: Please avoid whitespace changes

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/NoOpWorkerStateChangeListener.java
```diff
@@ -27,4 +27,5 @@ public NoOpWorkerStateChangeListener() {
 	public void onWorkerStateChange(WorkerState newState) {
 
 	}
+
```
sahilpalvia: Please avoid whitespace changes

PR #376, opened 2018-08-21
Feature/v1.x/proxy support

mscharp:
*Issue #, if available:*

*Description of changes:*
Added proxy support to the 1.x codeline.
Made updates per @sahilpalvia requests

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
+++ b/src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
@@ -126,7 +126,7 @@ public class MultiLangDaemonConfig {
 
             proxyHost = System.getProperty("http.proxyHost");
             proxyPort = Integer.parseInt(System.getProperty("http.proxyPort"));
-        } else if (System.getenv("http_proxy") != null) {
+        } else if (System.getenv("HTTP_PROXY") != null) {
             LOG.debug("Getting proxy info environment settings");
 
             try {
--- a/src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
+++ b/src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
@@ -18,6 +18,7 @@ import com.amazonaws.ClientConfiguration;
 import com.amazonaws.services.kinesis.clientlibrary.config.KinesisClientLibConfigurator;
 import com.amazonaws.services.kinesis.clientlibrary.lib.worker.KinesisClientLibConfiguration;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
@@ -29,7 +30,11 @@ import java.io.FileInputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Properties;
-import java.util.concurrent.*;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 
 /**
  * This class captures the configuration needed to run the MultiLangDaemon.
@@ -108,21 +113,21 @@ public class MultiLangDaemonConfig {
 
     private ClientConfiguration buildClientConfig(Properties properties) {
         ClientConfiguration clientConfig = new ClientConfiguration();
-        String proxyHost = "";
+        String proxyHost = null;
         int proxyPort = 0;
 
         if (properties.getProperty("http.proxyHost") != null) {
-            LOG.info("Getting proxy info from properties file.");
+            LOG.debug("Getting proxy info from properties file.");
 
             proxyHost = properties.getProperty("http.proxyHost");
             proxyPort = Integer.parseInt(properties.getProperty("http.proxyPort"));
         } else if (System.getProperty("http.proxyHost") != null) {
-            LOG.info("Getting proxy info from java system properties");
+            LOG.debug("Getting proxy info from java system properties");
 
             proxyHost = System.getProperty("http.proxyHost");
             proxyPort = Integer.parseInt(System.getProperty("http.proxyPort"));
         } else if (System.getenv("http_proxy") != null) {
-            LOG.info("Getting proxy info environment settings");
+            LOG.debug("Getting proxy info environment settings");
 
             try {
                 URI proxyAddr = new URI(System.getenv("http_proxy"));
@@ -131,10 +136,11 @@ public class MultiLangDaemonConfig {
                 proxyPort = proxyAddr.getPort();
             } catch (URISyntaxException e) {
                 LOG.error("System proxy not set correctly");
+                LOG.error(e);
             }
         }
 
-        if (!proxyHost.equals("") && proxyPort > 0)
+        if (StringUtils.isNotEmpty(proxyHost) && proxyPort > 0)
             clientConfig = clientConfig.withProxyHost(proxyHost).withProxyPort(proxyPort);
         else
             LOG.info("Not configuring proxy as none specified");
--- a/src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
+++ b/src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
@@ -50,6 +50,10 @@ public class MultiLangDaemonConfig {
     private static final String PROP_PROCESSING_LANGUAGE = "processingLanguage";
     private static final String PROP_MAX_ACTIVE_THREADS = "maxActiveThreads";
 
+    public static final String PROXY_HOST_PROP = "http.proxyHost";
+    public static final String PROXY_PORT_PROP = "http.proxyPort";
+    public static final String HTTP_PROXY_ENV_VAR = "HTTP_PROXY";
+
     private KinesisClientLibConfiguration kinesisClientLibConfig;
 
     private ExecutorService executorService;
@@ -116,21 +120,21 @@ public class MultiLangDaemonConfig {
         String proxyHost = null;
         int proxyPort = 0;
 
-        if (properties.getProperty("http.proxyHost") != null) {
+        if (properties.getProperty(PROXY_HOST_PROP) != null) {
             LOG.debug("Getting proxy info from properties file.");
 
-            proxyHost = properties.getProperty("http.proxyHost");
-            proxyPort = Integer.parseInt(properties.getProperty("http.proxyPort"));
-        } else if (System.getProperty("http.proxyHost") != null) {
+            proxyHost = properties.getProperty(PROXY_HOST_PROP);
+            proxyPort = Integer.parseInt(properties.getProperty(PROXY_PORT_PROP));
+        } else if (System.getProperty(PROXY_HOST_PROP) != null) {
             LOG.debug("Getting proxy info from java system properties");
 
-            proxyHost = System.getProperty("http.proxyHost");
-            proxyPort = Integer.parseInt(System.getProperty("http.proxyPort"));
-        } else if (System.getenv("HTTP_PROXY") != null) {
+            proxyHost = System.getProperty(PROXY_HOST_PROP);
+            proxyPort = Integer.parseInt(System.getProperty(PROXY_PORT_PROP));
+        } else if (System.getenv(HTTP_PROXY_ENV_VAR) != null) {
             LOG.debug("Getting proxy info environment settings");
 
             try {
-                URI proxyAddr = new URI(System.getenv("HTTP_PROXY"));
+                URI proxyAddr = new URI(System.getenv(HTTP_PROXY_ENV_VAR));
 
                 proxyHost = proxyAddr.getHost();
                 proxyPort = proxyAddr.getPort();
@@ -141,8 +145,7 @@ public class MultiLangDaemonConfig {
 
         if (StringUtils.isNotEmpty(proxyHost) && proxyPort > 0) {
             clientConfig = clientConfig.withProxyHost(proxyHost).withProxyPort(proxyPort);
-        }
-        else {
+        } else {
             LOG.debug("Not configuring proxy as none specified");
         }
 
--- a/src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
+++ b/src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
@@ -21,7 +21,11 @@ import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
-import java.io.*;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.FileInputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Properties;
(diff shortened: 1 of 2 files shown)


mscharp:
Updated PR with requested changes.

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
```diff
@@ -14,24 +14,22 @@
...
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
-
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
-
-import com.amazonaws.services.kinesis.clientlibrary.config.KinesisClientLibConfigurator;
-import com.amazonaws.services.kinesis.clientlibrary.lib.worker.KinesisClientLibConfiguration;
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import java.util.concurrent.*;
```
sahilpalvia: Avoid star imports

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
```diff
@@ -55,65 +53,102 @@
...
+                URI proxyAddr = new URI(System.getenv("http_proxy"));
+
+                proxyHost = proxyAddr.getHost();
+                proxyPort = proxyAddr.getPort();
+            } catch (URISyntaxException e) {
+                LOG.error("System proxy not set correctly");
+            }
+        }
+
+        if (!proxyHost.equals("") && proxyPort > 0)
```
sahilpalvia: You can use StringUtils instead of equals

```java
StringUtils.isNotEmpty(proxyHost)
```

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
```diff
@@ -55,65 +53,102 @@
...
+                kinesisClientLibConfig);
 
         LOG.info("Running " + kinesisClientLibConfig.getApplicationName() + " to process stream "
                 + kinesisClientLibConfig.getStreamName() + " with executable " + executableName);
         prepare(processingLanguage);
     }
 
+    private ClientConfiguration buildClientConfig(Properties properties) {
+        ClientConfiguration clientConfig = new ClientConfiguration();
+        String proxyHost = "";
```
sahilpalvia: Doesn't need to be empty string could be null

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
```diff
@@ -55,65 +53,102 @@
...
+        int proxyPort = 0;
+
+        if (properties.getProperty("http.proxyHost") != null) {
+            LOG.info("Getting proxy info from properties file.");
```
sahilpalvia: Does not need to be at an info level, debug would be better

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
```diff
@@ -55,65 +53,102 @@
...
+
+            proxyHost = properties.getProperty("http.proxyHost");
+            proxyPort = Integer.parseInt(properties.getProperty("http.proxyPort"));
+        } else if (System.getProperty("http.proxyHost") != null) {
+            LOG.info("Getting proxy info from java system properties");
```
sahilpalvia: Does not need to be at an info level, debug would be better

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
```diff
@@ -55,65 +53,102 @@
...
+
+            proxyHost = System.getProperty("http.proxyHost");
+            proxyPort = Integer.parseInt(System.getProperty("http.proxyPort"));
+        } else if (System.getenv("http_proxy") != null) {
+            LOG.info("Getting proxy info environment settings");
```
sahilpalvia: Does not need to be at an info level, debug would be better

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
```diff
@@ -55,65 +53,102 @@
...
+
+            try {
+                URI proxyAddr = new URI(System.getenv("http_proxy"));
+
+                proxyHost = proxyAddr.getHost();
+                proxyPort = proxyAddr.getPort();
+            } catch (URISyntaxException e) {
+                LOG.error("System proxy not set correctly");
```
sahilpalvia: Could you log the error too

src/test/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfigTest.java
```diff
@@ -88,4 +102,126 @@ public void propertyValidation() {
...
+            MultiLangDaemonConfig config;
+
+            try {
+                config = new MultiLangDaemonConfig(FILENAME, classLoader, buildMockConfigurator());
+
+                // host doesn't have protocol here because URI splits that into a separate piece
+                assertAgainstKclConfig(config.getKinesisClientLibConfiguration(), null, -1);
+            } catch (IllegalArgumentException e) {
+                // Bad
+            } catch (IOException e) {
```
sahilpalvia: No need to catch exceptions if you are going to call fail. You can let the exception be thrown from the test. The test will fail if exception is thrown.

src/test/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfigTest.java
```diff
@@ -88,4 +102,126 @@ public void propertyValidation() {
         }
     }
 
+    @Test public void KinesisClientLibConfiguration_should_get_proxy_info_from_properties_file() {
```
sahilpalvia: Could you keep the annotation before the test method
```java
@Test
public void propertyValidation() {
    // Test goes here
}
```

Could you use the naming standards, use camel case in the method name and not underscores and add test at the start for eg: testProxyInfoFromPropertiesFile.

src/test/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfigTest.java
```diff
@@ -39,23 +43,37 @@ private KinesisClientLibConfigurator buildMockConfigurator() {
...
+                credentialsProvider, "cool-worker");
         KinesisClientLibConfigurator configurator = Mockito.mock(KinesisClientLibConfigurator.class);
         Mockito.doReturn(kclConfig).when(configurator).getConfiguration(Mockito.any(Properties.class));
         return configurator;
     }
 
-    @Test
-    public void constructorTest() throws IOException {
+    // Function to mock ENV variables
+    public static void setEnv(Map<String, String> newenv) throws Exception {
```
sahilpalvia: Make this method private, and doesn't need to be static.

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
mscharp: Will change

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
mscharp: ok

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
mscharp: ok

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
mscharp: ok

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
mscharp: ok

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
mscharp: will do

src/test/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfigTest.java
mscharp: Your formatter is what puts the @Test annotation on the same line as the test...

I'll change the name of the tests

src/test/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfigTest.java
mscharp: I have to catch the exception from the setEnv method, which is why that is there.

src/test/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfigTest.java
mscharp: I will suggest though that "test" as a prefix doesn't add anything, especially with a @Test annotation.  Also, the other tests in there don't follow this convention, nor do the other test names actually tell you what the test is checking...

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
mscharp: ok

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
sahilpalvia: You can set the exception in the same log message,

```java
LOG.error("Message", e)
```

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
```diff
@@ -55,65 +58,103 @@
...
+                LOG.error(e);
+            }
+        }
+
+        if (StringUtils.isNotEmpty(proxyHost) && proxyPort > 0)
```
sahilpalvia: Can you add brackets around the if else condition

```java
if (condition) {
    // do something
} else {
    // do something
}
```

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
```diff
@@ -55,65 +58,103 @@
...
+            clientConfig = clientConfig.withProxyHost(proxyHost).withProxyPort(proxyPort);
+        else
+            LOG.info("Not configuring proxy as none specified");
```
sahilpalvia: Debug level logging would be better.

src/test/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfigTest.java
sahilpalvia: Junit will fail the test if the test throws an exception, and if it is not expected.

```java
@Test
public void testScenario() throws Exception {
    // test goes here
}
```

If the above test throws an exception, Junit will fail the test.

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
```diff
@@ -55,65 +58,103 @@
...
+            LOG.debug("Getting proxy info from properties file.");
+
+            proxyHost = properties.getProperty("http.proxyHost");
+            proxyPort = Integer.parseInt(properties.getProperty("http.proxyPort"));
+        } else if (System.getProperty("http.proxyHost") != null) {
+            LOG.debug("Getting proxy info from java system properties");
+
+            proxyHost = System.getProperty("http.proxyHost");
+            proxyPort = Integer.parseInt(System.getProperty("http.proxyPort"));
+        } else if (System.getenv("http_proxy") != null) {
```
sahilpalvia: Can you change the property to upper case HTTP_PROXY, when reading from System properties to be inline with the example here 

https://docs.aws.amazon.com/cli/latest/userguide/cli-http-proxy.html

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
```diff
@@ -55,65 +58,104 @@
...
+                proxyHost = proxyAddr.getHost();
+                proxyPort = proxyAddr.getPort();
+            } catch (URISyntaxException e) {
+                LOG.error("System proxy not set correctly", e);
+            }
+        }
+
+        if (StringUtils.isNotEmpty(proxyHost) && proxyPort > 0) {
+            clientConfig = clientConfig.withProxyHost(proxyHost).withProxyPort(proxyPort);
+        }
```
sahilpalvia: Keep the else after the closing brackets adding a space.

```java
if (condition) {
    // do something
} else {
    // do something
}
```

src/test/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfigTest.java
```diff
@@ -88,4 +104,87 @@ public void propertyValidation() {
         }
     }
 
+    @Test
+    public void testKinesisClientLibConfigurationShouldGetProxyInfoFromPropertiesFile() throws Exception {
+        String PROPERTIES =
```
sahilpalvia: Static variables are normally uppercase. Could you change this PROPERTIES to lower case properties.

src/test/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfigTest.java
```diff
@@ -67,14 +86,11 @@ public void constructorTest() throws IOException {
 
     @Test
     public void propertyValidation() {
-        String PROPERTIES_NO_EXECUTABLE_NAME =
-                "applicationName = testApp \n" + "streamName = fakeStream \n"
-                        + "AWSCredentialsProvider = DefaultAWSCredentialsProviderChain\n"
-                        + "processingLanguage = malbolge";
+        String PROPERTIES_NO_EXECUTABLE_NAME = "applicationName = testApp \n" + "streamName = fakeStream \n"
```
sahilpalvia: Lowercase variable, i.e. String propertiesNoExecutableName

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
```diff
@@ -55,65 +58,104 @@
...
                 + kinesisClientLibConfig.getStreamName() + " with executable " + executableName);
         prepare(processingLanguage);
     }
 
+    private ClientConfiguration buildClientConfig(Properties properties) {
+        ClientConfiguration clientConfig = new ClientConfiguration();
+        String proxyHost = null;
+        int proxyPort = 0;
+
+        if (properties.getProperty("http.proxyHost") != null) {
```
sahilpalvia: "http.proxyHost" and "http.proxyPort" can be extracted to a String, since it is being used in multiple places.

```java
String proxyHostProperty = "http.proxyHost";
String proxyPortProperty = "http.proxyPort";
String httpProxySystemProperty = "HTTP_PROXY";
```

src/test/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfigTest.java
mscharp: Agreed.  Just copied and pasted from the constructorTest()

src/test/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfigTest.java
mscharp: I didn't write this test but, sure I'll change this and the constructorTest so both are lower cased.

src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangDaemonConfig.java
mscharp: Yeah, I thought about that after.  I'll make them static an public so that I can reference them in the test as well.

PR #413, opened 2018-09-20
Introducing callback for DynamoDB lease table

sahilpalvia:
*Issue #, if available:*
N/A

*Description of changes:*
* Introducing TableCreatorCallback for DynamoDB lease management
* Introducing DoesNothingTableCreatorCallback
* Intoducing TableCreatorCallback config in LeaseManagementConfig, with DoesNothingTableCreatorCallback as the default


amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/TableCreatorCallback.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package software.amazon.kinesis.leases;
+
+import software.amazon.kinesis.leases.dynamodb.TableCreatorCallbackInput;
+
+/**
+ * Callback interface for interacting with the DynamoDB lease table post creation.
+ */
+public interface TableCreatorCallback {
```
pfifer: I would annotate this with `@FunctionalInterface` if you intend it to be used with a lambda.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/NoOpTableCreatorCallback.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package software.amazon.kinesis.leases.dynamodb;
+
+import software.amazon.kinesis.leases.TableCreatorCallback;
+
+/**
+ * This implementation of the TableCreatorCallback does nothing.
+ */
+public class NoOpTableCreatorCallback implements TableCreatorCallback {
```
pfifer: I would probably make this a singleton on the `TableCreatorCallback` interface instead of its own class.

PR #417, opened 2018-09-21
Add a listener to capture task execution in shardConsumer

akhani18:
*Issue #, if available:*

*Description of changes:*
- Add a TaskExecutionListener as part of LifecycleConfig that can be used by all ShardConsumers in a Scheduler to issue notifications about beginning and ending of a task execution
- Provide a default NoOp task execution listener
- Update ShardConsumer constructor to take an instance TaskExecutionListener
- Update executeTask method to issue n onTaskBegin and onTaskEnd notification to any listener providing ShardInfo and the current state the shard is in.
- Update ShardConsumerTest to validate all interactions with taskExecutionListener


amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/NoOpTaskExecutionListener.java
```diff
@@ -0,0 +1,22 @@
+package software.amazon.kinesis.lifecycle;
+
+import software.amazon.kinesis.leases.ShardInfo;
+
+public class NoOpTaskExecutionListener implements TaskExecutionListener {
+    /**
+     * Empty constructor for NoOp Shard Prioritization.
+     */
+    public NoOpTaskExecutionListener() {
```
sahilpalvia: No need to have the empty constructor. Could you remove the empty lines between the brackets.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumer.java
```diff
@@ -95,9 +96,10 @@
     private final InternalSubscriber subscriber;
 
     public ShardConsumer(RecordsPublisher recordsPublisher, ExecutorService executorService, ShardInfo shardInfo,
-                         Optional<Long> logWarningForTaskAfterMillis, ShardConsumerArgument shardConsumerArgument) {
+                         Optional<Long> logWarningForTaskAfterMillis, ShardConsumerArgument shardConsumerArgument,
+                         TaskExecutionListener taskExecutionListener) {
         this(recordsPublisher, executorService, shardInfo, logWarningForTaskAfterMillis, shardConsumerArgument,
-                ConsumerStates.INITIAL_STATE,
+                taskExecutionListener, ConsumerStates.INITIAL_STATE,
```
sahilpalvia: Is there a reason for changing the order for the chained constructor? If not could you add the new parameter to the end of the list.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/NoOpTaskExecutionListener.java
akhani18: done

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumer.java
akhani18: Mainly, I just wanted to avoid adding a reference type (TaskExecutionListener) after a primitive type (int bufferSize) in the constructor as it didn't look aesthetically pleasing.
But I'm fine with adding this new argument to the end of the list.
Why would you not want to use builder pattern for the constructor here?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumer.java
```diff
@@ -392,6 +397,7 @@ private synchronized void executeTask(ProcessRecordsInput input) {
             }
             taskOutcome = resultToOutcome(result);
         }
+        taskExecutionListener.onTaskEnd(currentState, shardInfo);
```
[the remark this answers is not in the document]
akhani18: - Documentation on ConsumerTask.java says that any exception during task.call() will be captured in TaskResult, returned by task.call(). I also looked at all *.Task classes to confirm that, so it would be safe to assume that it won't throw exception and onTaskBegin will always be accompanied by onTaskEnd in all known cases.

- onTaskBegin and onTaskEnd should always be called in the same order.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/TaskExecutionListener.java
```diff
@@ -0,0 +1,27 @@
...
+import software.amazon.kinesis.leases.ShardInfo;
+
+/**
+ * A listener for callbacks on task execution lifecycle for for a shard.
+ */
+public interface TaskExecutionListener {
+
+    void onTaskBegin(ConsumerState state, ShardInfo shardInfo);
+
+    void onTaskEnd(ConsumerState state, ShardInfo shardInfo);
```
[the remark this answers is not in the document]
akhani18: TaskOutcome - Sure, as suggested by @sahilpalvia I'm going to create a data class and I can include things like ShardInfo, TaskType, TaskOutcome in it. Task execution time can be implicitly calculated on the client side by looking at the delta between onTaskBegin and onTaskEnd.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/TaskExecutionListener.java
```diff
@@ -0,0 +1,27 @@
...
+package software.amazon.kinesis.lifecycle;
+
+import software.amazon.kinesis.leases.ShardInfo;
+
+/**
+ * A listener for callbacks on task execution lifecycle for for a shard.
+ */
+public interface TaskExecutionListener {
+
+    void onTaskBegin(ConsumerState state, ShardInfo shardInfo);
```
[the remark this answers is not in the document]
akhani18: Added that to the class header because it applies to both methods of the class.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -571,7 +571,8 @@ protected ShardConsumer buildConsumer(@NonNull final ShardInfo shardInfo,
                 shardDetector,
                 metricsFactory,
                 aggregatorUtil);
-        return new ShardConsumer(cache, executorService, shardInfo, lifecycleConfig.logWarningForTaskAfterMillis(), argument);
+        return new ShardConsumer(cache, executorService, shardInfo, lifecycleConfig.logWarningForTaskAfterMillis(),
+                argument, lifecycleConfig.taskExecutionListener());
```
sahilpalvia: Add the listener to the ShardConsumerArgument. The argument was created for such scenarios where if new parameters need to be added  they can be.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/LifecycleConfig.java
```diff
@@ -46,4 +46,10 @@
      */
     private AggregatorUtil aggregatorUtil = new AggregatorUtil();
 
+    /**
+     * TaskExecutionListener to be used to handle events during task execution lifecycle for a shard.
+     *
+     * <p>Default value: {@link NoOpTaskExecutionListener}</p>
+     */
+    private TaskExecutionListener taskExecutionListener = new NoOpTaskExecutionListener();
```
sahilpalvia: Is the listener the same for all ShardConsumers? It would be better to have a factory and create a listener per ShardConsumer.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/NoOpTaskExecutionListener.java
```diff
@@ -0,0 +1,31 @@
...
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+package software.amazon.kinesis.lifecycle;
+
+import software.amazon.kinesis.lifecycle.events.TaskExecutionListenerInput;
+
+/**
+ * A default implementation of TaskExecutionListener that takes no action on task execution.
```
sahilpalvia: Avoid using `default implementation`. 

```java
/**
* NoOp implementation of the TaskExecutionListener, no actions are performed on task execution.
*/
```

Something like this could be better.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/TaskExecutionListener.java
```diff
@@ -0,0 +1,28 @@
...
+
+import software.amazon.kinesis.lifecycle.events.TaskExecutionListenerInput;
+
+/**
+ * A listener for callbacks on task execution lifecycle for for a shard. The implementation for this interface should
+ * not perform any blocking operation like remote service calls, because that would block subsequent task executions.
+ */
+public interface TaskExecutionListener {
+
+    void onTaskBegin(TaskExecutionListenerInput input);
```
sahilpalvia: Missing javadoc. Javadoc is helpful for any publicly exposed interfaces.

I think `beforeTaskExecution` and `afterTaskExecution` are better names, since that's when the methods are called in the ShardConsumer implementation.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumer.java
```diff
@@ -379,6 +384,11 @@ private synchronized void processData(ProcessRecordsInput input) {
     }
 
     private synchronized void executeTask(ProcessRecordsInput input) {
+        TaskExecutionListenerInput taskExecutionListenerInput = TaskExecutionListenerInput.builder()
```
sahilpalvia: Do not share the input objects between the methods.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/events/TaskExecutionListenerInput.java
```diff
@@ -0,0 +1,52 @@
...
+@Builder(toBuilder = true)
+@Getter
+@Accessors(fluent = true)
+@EqualsAndHashCode
+@ToString
+public class TaskExecutionListenerInput {
+    /**
+     * Detailed information about the shard whose progress is monitored by TaskExecutionListener.
+     */
+    private ShardInfo shardInfo;
```
sahilpalvia: Could you make all variables final

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/events/TaskExecutionListenerInput.java
```diff
@@ -0,0 +1,52 @@
...
+import software.amazon.kinesis.lifecycle.TaskType;
+import software.amazon.kinesis.lifecycle.TaskExecutionListener;
+
+/**
+ * Container for the parameters to the TaskExecutionListener's
+ * {@link TaskExecutionListener#onTaskBegin(TaskExecutionListenerInput)} method.
+ * {@link TaskExecutionListener#onTaskEnd(TaskExecutionListenerInput)} method.
+ */
+@Builder(toBuilder = true)
+@Getter
```
pfifer: Instead of all (\@Getter, \@EqualsAndHashCode, \@ToString) just use \@Data, this does work with \@Builder.  If the variables are all final \@Data will not create setters.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumer.java
pfifer: As long as the fields are final sharing this way is fine.  `toBuilder` will trigger the creation of a new object.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/LifecycleConfig.java
pfifer: I think having a single instance of the listener is fine.  The implementations of listener will need to handle that their singleton state.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
pfifer: The `ShardConsumerArgument` was primarily added for things that are passed downstream to the tasks themselves, and not used by the `ShardConsumer`.  I think it's fine that the `TaskExecutionListener` is passed in, as it's used primarily by the ShardConsumer.

pfifer:
Thanks for providing this.

Can please resolve the conflicts and push and update to the PR?

akhani18:
I fixed the merge conflict here itself in the github UI and looks like the checks have passed. Is that sufficient?

Issue #404, opened 2018-09-19
MultiLangDaemon broken in 2.0.2? (credentials provider issues)


zerth:
I see the following error and stack trace when specifying `software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider` (the new class replacing `DefaultAWSCredentialsProviderChain`) as the value of `AWSCredentialsProvider` in a MultiLangDaemon properties file (2.0.2):

Error:
```
2018-09-18 18:00:01,105 [main] DEBUG c.a.s.k.m.c.AWSCredentialsProviderPropertyValueDecoder [NONE] - Can't find any credentials provider matching software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider.
```

Stack trace:
```
java.lang.InstantiationException: software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider 
java.lang.InstantiationException: software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider
	at java.lang.Class.newInstance(Class.java:423)
	at com.amazonaws.services.kinesis.multilang.config.AWSCredentialsProviderPropertyValueDecoder.getValidCredentialsProviders(AWSCredentialsProviderPropertyValueDecoder.java:91)
	at com.amazonaws.services.kinesis.multilang.config.AWSCredentialsProviderPropertyValueDecoder.decodeValue(AWSCredentialsProviderPropertyValueDecoder.java:52)
	at com.amazonaws.services.kinesis.multilang.config.AWSCredentialsProviderPropertyValueDecoder.decodeValue(AWSCredentialsProviderPropertyValueDecoder.java:29)
	at com.amazonaws.services.kinesis.multilang.config.KinesisClientLibConfigurator.getConfiguration(KinesisClientLibConfigurator.java:109)
	at com.amazonaws.services.kinesis.multilang.MultiLangDaemonConfig.<init>(MultiLangDaemonConfig.java:101)
	at com.amazonaws.services.kinesis.multilang.MultiLangDaemonConfig.<init>(MultiLangDaemonConfig.java:74)
	at com.amazonaws.services.kinesis.multilang.MultiLangDaemonConfig.<init>(MultiLangDaemonConfig.java:58)
	at com.amazonaws.services.kinesis.multilang.MultiLangDaemon.main(MultiLangDaemon.java:128)
Caused by: java.lang.NoSuchMethodException: software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider.<init>()
[...]

PR #443, opened 2018-10-14
refs #404, make multilang work in v2

wwwjfy:
*Issue #, if available:* #404

*Description of changes:*
I have little knowledge about this library, so I fixed the issues I met when starting a multilang v2 daemon. I'm sure there are things I missed, but with this PR, I can successfully run the daemon and consume messages from it.


pfifer:
Thank you for providing this, but unforunately we won't be able to accept the changes.  

To everyone we're currently working on updating the multilang daemon for KCL 2.0, and hope to share more details in the near future.

PR #490, opened 2019-01-30
separated out some lease taker logic as interface implementations to …

achitojha:
…create abstraction

Separated out the leaseTaker logic :
1) to fetch leases from expired leases
2) to fetch the number of leases to take

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibLeaseCoordinator.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibLeaseCoordinator.java
@@ -22,6 +22,8 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
 
+import com.amazonaws.services.kinesis.leases.impl.GenericLeaseSelector;
+import com.amazonaws.services.kinesis.leases.interfaces.LeaseSelector;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
@@ -50,6 +52,7 @@ class KinesisClientLibLeaseCoordinator extends LeaseCoordinator<KinesisClientLea
 
     private static final long DEFAULT_INITIAL_LEASE_TABLE_READ_CAPACITY = 10L;
     private static final long DEFAULT_INITIAL_LEASE_TABLE_WRITE_CAPACITY = 10L;
+    private static final LeaseSelector<KinesisClientLease> DEFAULT_LEASE_SELECTOR = new GenericLeaseSelector<KinesisClientLease>();
 
     private final ILeaseManager<KinesisClientLease> leaseManager;
 
@@ -61,12 +64,14 @@ class KinesisClientLibLeaseCoordinator extends LeaseCoordinator<KinesisClientLea
      * @param workerIdentifier Used to identify this worker process
      * @param leaseDurationMillis Duration of a lease in milliseconds
      * @param epsilonMillis Delta for timing operations (e.g. checking lease expiry)
+     * @param leaseSelector Lease selector which decides which leases to take
      */
     public KinesisClientLibLeaseCoordinator(ILeaseManager<KinesisClientLease> leaseManager,
             String workerIdentifier,
             long leaseDurationMillis,
-            long epsilonMillis) {
-        super(leaseManager, workerIdentifier, leaseDurationMillis, epsilonMillis);
+            long epsilonMillis,
+            LeaseSelector<KinesisClientLease> leaseSelector) {
+        super(leaseManager, leaseSelector, workerIdentifier, leaseDurationMillis, epsilonMillis);
         this.leaseManager = leaseManager;
     }
 
@@ -75,19 +80,35 @@ class KinesisClientLibLeaseCoordinator extends LeaseCoordinator<KinesisClientLea
      * @param workerIdentifier Used to identify this worker process
      * @param leaseDurationMillis Duration of a lease in milliseconds
      * @param epsilonMillis Delta for timing operations (e.g. checking lease expiry)
-     * @param metricsFactory Metrics factory used to emit metrics
      */
     public KinesisClientLibLeaseCoordinator(ILeaseManager<KinesisClientLease> leaseManager,
             String workerIdentifier,
             long leaseDurationMillis,
+            long epsilonMillis) {
+        this(leaseManager, workerIdentifier, leaseDurationMillis, epsilonMillis, DEFAULT_LEASE_SELECTOR);
+    }
+
+    /**
+     * @param leaseManager Lease manager which provides CRUD lease operations.
+     * @param leaseSelector Lease selector which decides which leases to take
+     * @param workerIdentifier Used to identify this worker process
+     * @param leaseDurationMillis Duration of a lease in milliseconds
+     * @param epsilonMillis Delta for timing operations (e.g. checking lease expiry)
+     * @param metricsFactory Metrics factory used to emit metrics
+     */
+    public KinesisClientLibLeaseCoordinator(ILeaseManager<KinesisClientLease> leaseManager,
+            LeaseSelector<KinesisClientLease> leaseSelector,
+            String workerIdentifier,
+            long leaseDurationMillis,
             long epsilonMillis,
             IMetricsFactory metricsFactory) {
-        super(leaseManager, workerIdentifier, leaseDurationMillis, epsilonMillis, metricsFactory);
+        super(leaseManager, leaseSelector, workerIdentifier, leaseDurationMillis, epsilonMillis, metricsFactory);
         this.leaseManager = leaseManager;
     }
 
     /**
      * @param leaseManager Lease manager which provides CRUD lease operations.
+     * @param leaseSelector Lease selector which decides which leases to take
      * @param workerIdentifier Used to identify this worker process
      * @param leaseDurationMillis Duration of a lease in milliseconds
      * @param epsilonMillis Delta for timing operations (e.g. checking lease expiry)
@@ -96,6 +117,7 @@ class KinesisClientLibLeaseCoordinator extends LeaseCoordinator<KinesisClientLea
      * @param metricsFactory Metrics factory used to emit metrics
      */
     public KinesisClientLibLeaseCoordinator(ILeaseManager<KinesisClientLease> leaseManager,
+            LeaseSelector<KinesisClientLease> leaseSelector,
             String workerIdentifier,
             long leaseDurationMillis,
             long epsilonMillis,
@@ -103,7 +125,7 @@ class KinesisClientLibLeaseCoordinator extends LeaseCoordinator<KinesisClientLea
             int maxLeasesToStealAtOneTime,
             int maxLeaseRenewerThreadCount,
             IMetricsFactory metricsFactory) {
-        super(leaseManager, workerIdentifier, leaseDurationMillis, epsilonMillis, maxLeasesForWorker,
+        super(leaseManager, leaseSelector, workerIdentifier, leaseDurationMillis, epsilonMillis, maxLeasesForWorker,
                 maxLeasesToStealAtOneTime, maxLeaseRenewerThreadCount, metricsFactory);
         this.leaseManager = leaseManager;
     }
@@ -136,19 +158,19 @@ class KinesisClientLibLeaseCoordinator extends LeaseCoordinator<KinesisClientLea
 
     /**
      * Sets the checkpoint for a shard and updates ownerSwitchesSinceCheckpoint.
-     * 
+     *
      * @param shardId shardId to update the checkpoint for
      * @param checkpoint checkpoint value to set
      * @param concurrencyToken obtained by calling Lease.getConcurrencyToken for a currently held lease
-     * 
+     *
      * @return true if checkpoint update succeeded, false otherwise
-     * 
+     *
      * @throws InvalidStateException if lease table does not exist
      * @throws ProvisionedThroughputException if DynamoDB update fails due to lack of capacity
      * @throws DependencyException if DynamoDB update fails in an unexpected way
      */
     boolean setCheckpoint(String shardId, ExtendedSequenceNumber checkpoint, UUID concurrencyToken)
-        throws DependencyException, InvalidStateException, ProvisionedThroughputException {
+            throws DependencyException, InvalidStateException, ProvisionedThroughputException {
         KinesisClientLease lease = getCurrentlyHeldLease(shardId);
         if (lease == null) {
             LOG.info(String.format(
@@ -170,7 +192,7 @@ class KinesisClientLibLeaseCoordinator extends LeaseCoordinator<KinesisClientLea
      */
     @Override
     public void setCheckpoint(String shardId, ExtendedSequenceNumber checkpointValue, String concurrencyToken)
-        throws KinesisClientLibException {
+            throws KinesisClientLibException {
         try {
             boolean wasSuccessful = setCheckpoint(shardId, checkpointValue, UUID.fromString(concurrencyToken));
             if (!wasSuccessful) {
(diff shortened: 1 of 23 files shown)


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/InitialPositionInStreamExtended.java
```diff
@@ -20,7 +20,7 @@
  * Class that houses the entities needed to specify the position in the stream from where a new application should
  * start.
  */
-class InitialPositionInStreamExtended {
+public class InitialPositionInStreamExtended {
```
sahilpalvia: Does this need to be changed to `public`?

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisLeaseCleanupValidator.java
```diff
@@ -0,0 +1,50 @@
...
+import com.amazonaws.services.kinesis.leases.impl.KinesisClientLease;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import java.util.Set;
+
+/**
+ * Represents the class that decides if a lease is eligible for cleanup.
+ */
+public class KinesisLeaseCleanupValidator implements LeaseCleanupValidator {
```
sahilpalvia: Doesn't need to be `public` can be default.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisShardSyncer.java
```diff
@@ -0,0 +1,844 @@
...
+import com.amazonaws.services.kinesis.metrics.interfaces.MetricsLevel;
+import com.amazonaws.services.kinesis.model.Shard;
+
+/**
+ * KinesisShardSyncer syncs leases with shards of the Kinesis stream.
+ * It will create new leases/activities when it discovers new Kinesis shards (bootstrap/resharding).
+ * It deletes leases for shards that have been trimmed from Kinesis, or if we've completed processing it
+ * and begun processing it's child shards.
+ */
+public class KinesisShardSyncer implements ShardSyncer {
```
sahilpalvia: Same thing, can be made default doesn't need to be `public`

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardConsumer.java
```diff
@@ -388,9 +398,17 @@ ShutdownReason getShutdownReason() {
         return shutdownReason;
     }
 
+    /**
+     * @return the shardSyncer
+     */
+    public ShardSyncer getShardSyncer() {
```
sahilpalvia: Just a tip: KCL uses lombok, and you can create this method by adding the `@Getter` annotation.

```java
@Getter
private final ShardSyncer shardSyncer;
```

src/main/java/com/amazonaws/services/kinesis/leases/impl/KinesisLeaseSelector.java
```diff
@@ -0,0 +1,40 @@
+package com.amazonaws.services.kinesis.leases.impl;
+
+import com.amazonaws.services.kinesis.leases.interfaces.ILeaseSelector;
+
+import java.util.*;
```
sahilpalvia: Avoid `*` imports

src/main/java/com/amazonaws/services/kinesis/leases/interfaces/ILeaseSelector.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * ILeaseSelector abstracts away the lease selection logic from the application code that's using leasing.
+ * It owns filtering of the leases to be taken.
+ */
+public interface ILeaseSelector<T extends Lease> {
```
sahilpalvia: Could you rename this to `LeaseSelector` instead of `ILeaseSelector`

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncer.java
```diff
@@ -1,892 +1,39 @@
...
-    static synchronized void bootstrapShardLeases(IKinesisProxy kinesisProxy,
-            ILeaseManager<KinesisClientLease> leaseManager,
-            InitialPositionInStreamExtended initialPositionInStream,
-            boolean cleanupLeasesOfCompletedShards,
-            boolean ignoreUnexpectedChildShards)
-        throws DependencyException, InvalidStateException, ProvisionedThroughputException, KinesisClientLibIOException {
-        syncShardLeases(kinesisProxy, leaseManager, initialPositionInStream, cleanupLeasesOfCompletedShards,
-                        ignoreUnexpectedChildShards);
-    }
+public interface ShardSyncer {
```
sahilpalvia: Doesn't need to be changed to an interface unless implementing a different `ShardSyncer`. The only reason I could see for `KinesisShardSyncer` to be created is because this class has a private method. If that is the only reason, I would recommend changing the constructor on the `ShardSyncer` rather than changing it to an interface.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/InitialPositionInStreamExtended.java
achitojha: This is required in the DDB streams adapter and hence it had to be made public

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisLeaseCleanupValidator.java
achitojha: Done

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardConsumer.java
achitojha: Using lombok for the getter now.

src/main/java/com/amazonaws/services/kinesis/leases/impl/KinesisLeaseSelector.java
achitojha: Using explicit imports now.

src/main/java/com/amazonaws/services/kinesis/leases/interfaces/ILeaseSelector.java
achitojha: All other interfaces in this class have their names starting with 'I'.

sahilpalvia: Yeah, that was done earlier. Going forward we are trying to avoid that.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncer.java
```diff
@@ -210,28 +201,24 @@ private static void assertAllParentShardsAreClosed(Set<String> inconsistentShard
...
-     * @param shardIdsOfClosedShards Id of the shard which is expected to be closed
-     * @return ShardIds of child shards (children of the expectedClosedShard)
-     * @throws KinesisClientLibIOException
+     * is covered by its child shards.
      */
-    static synchronized void assertClosedShardsAreCoveredOrAbsent(Map<String, Shard> shardIdToShardMap,
-            Map<String, Set<String>> shardIdToChildShardIdsMap,
-            Set<String> shardIdsOfClosedShards) throws KinesisClientLibIOException {        
+    synchronized void assertClosedShardsAreCoveredOrAbsent(Map<String, Shard> shardIdToShardMap,
+                                                           Map<String, Set<String>> shardIdToChildShardIdsMap,
```
sahilpalvia: Could you use the formatter from: https://github.com/awslabs/amazon-kinesis-client/blob/master/formatter/formatter.xml

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -465,7 +468,7 @@ public Worker(
         this(applicationName, recordProcessorFactory, config, streamConfig, initialPositionInStream, parentShardPollIntervalMillis,
                 shardSyncIdleTimeMillis, cleanupLeasesUponShardCompletion, checkpoint, leaseCoordinator, execService,
                 metricsFactory, taskBackoffTimeMillis, failoverTimeMillis, skipShardSyncAtWorkerInitializationIfLeasesExist,
-                shardPrioritization, Optional.empty(), Optional.empty(), DEFAULT_WORKER_STATE_CHANGE_LISTENER);
+                shardPrioritization, Optional.empty(), Optional.empty(), DEFAULT_WORKER_STATE_CHANGE_LISTENER, new ShardSyncer(new KinesisLeaseCleanupValidator()));
```
sahilpalvia: Seems like the `ShardSyncer` is being created multiple times. Can you make sure to create it only once. Each `Worker` should have only 1 `ShardSyncer`.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -1272,6 +1297,10 @@ public Worker build() {
                 workerStateChangeListener = DEFAULT_WORKER_STATE_CHANGE_LISTENER;
             }
 
+            if(shardSyncer == null) {
+                shardSyncer = new ShardSyncer(new KinesisLeaseCleanupValidator());
```
sahilpalvia: Don't take `ShardSyncer` as a option, take the `LeaseCleanupValidator` as an option.

src/main/java/com/amazonaws/services/kinesis/leases/impl/KinesisLeaseSelector.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * KinesisLeaseSelector abstracts away the lease selection logic from the application code that's using leasing.
+ * It owns filtering of the leases to be taken.
+ */
+public class KinesisLeaseSelector<T extends Lease> implements ILeaseSelector<T> {
```
sahilpalvia: Doesn't need to be public. At the creation of the `LeaseCoordinator` if `LeaseSelector` is passed use the one that is passed otherwise create this as the default one.

src/test/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/WorkerTest.java
```diff
@@ -66,6 +66,8 @@
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import com.amazonaws.services.kinesis.leases.impl.*;
```
sahilpalvia: Avoid `*` imports

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
achitojha: The ShardSyncer has only been instantiated in the constructors, so each worker has only 1 shard syncer. I have created a private static final instance instead now.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
achitojha: That would create multiple shardSyncer instances, I'll stick to the current state.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -1201,6 +1215,19 @@ public Builder recordProcessorFactory(IRecordProcessorFactory recordProcessorFac
...
 
+        /**
+         * Provide a {@link LeaseSelector
+         * LeaseSelector}.
+         *
+         * @param leaseSelector
+         *            Used to get fetch the list of leases that can be taken
+         * @return A reference to this updated object so that method calls can be chained together.
+         */
+        public Builder leaseSelector(LeaseSelector<KinesisClientLease> leaseSelector) {
```
sahilpalvia: You have the `@Setter` annotation on the variable. You should not need to create this method.

src/test/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncTaskIntegrationTest.java
```diff
@@ -125,7 +126,7 @@ public final void testCall() throws DependencyException, InvalidStateException,
                 InitialPositionInStreamExtended.newInitialPosition(InitialPositionInStream.LATEST),
                 false,
                 false,
-                0L);
+                0L, shardSyncer);
```
sahilpalvia: This looks weird, could you add `shardSyncer` to a new line.

src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseCoordinator.java
```diff
@@ -89,32 +90,36 @@
      * Constructor.
      *
      * @param leaseManager LeaseManager instance to use
+     * @param leaseSelector LeaseSelector instance to use
      * @param workerIdentifier Identifies the worker (e.g. useful to track lease ownership)
      * @param leaseDurationMillis Duration of a lease
      * @param epsilonMillis Allow for some variance when calculating lease expirations
      */
     public LeaseCoordinator(ILeaseManager<T> leaseManager,
+            LeaseSelector<T> leaseSelector,
```
pfifer: Adding this parameter will break custom implementations of the LeaseCoordinator.  We need to determine a way to add the parameter without breaking existing implementations.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
achitojha: Done

src/test/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncTaskIntegrationTest.java
achitojha: Modified

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibLeaseCoordinator.java
```diff
@@ -58,36 +59,41 @@
 
     /**
      * @param leaseManager Lease manager which provides CRUD lease operations.
+     * @param leaseSelector Lease selector which decides which leases to take
      * @param workerIdentifier Used to identify this worker process
      * @param leaseDurationMillis Duration of a lease in milliseconds
      * @param epsilonMillis Delta for timing operations (e.g. checking lease expiry)
      */
     public KinesisClientLibLeaseCoordinator(ILeaseManager<KinesisClientLease> leaseManager,
```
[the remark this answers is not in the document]
achitojha: I've added a new method that has the same parameters as earlier, and calls the new constructor with DEFAULT_LEASE_SELECTOR. I believe that'll not break any existing customers.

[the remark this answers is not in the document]
pfifer: You also changed `LeaseCoordinator<T>` which could break implementations that derive from it.

Issue #537, opened 2019-04-07
Kinesis consumer cannot checkpoint because it cannot renew the lease


BobbyJohansen:
`[shardId-000000000011] Worker 885e7e2d-c00e-4145-8f3d-b4fce1f1c46b refusing to update lease with key shardId-000000000011 because concurrency tokens don't match`

Our consumer will sometimes fire this error out of the blue because it thinks it lost the lease and starts lagging behind our stream. I recently upgraded to 

`<groupId>software.amazon.kinesis</groupId>
  <artifactId>amazon-kinesis-client-pom</artifactId>
  <version>2.1.3</version>`

but still this error occurs. We get stuck in a process loop since we do not system.exit when a shutdown error occurs. Please help.

BobbyJohansen:
It looks like the lease coordinator does not mark shard record processors for shutdown when it cannot renew the lease due to this error. Which means that when I checkpoint in the software.amazon.kinesis.lifecycle.ProcessTask#callProcessRecords() where processRecords() is called to my code I end not being able to checkpoint and I throw a Java Error. Even more confusing is that software.amazon.kinesis.lifecycle.ProcessTask#callProcessRecords does not catch Error and cause a shutdown. I think that there should be an error type that I can bubble up to software.amazon.kinesis.lifecycle.ProcessTask#callProcessRecords that will cause the ShardRecordProcessor to perform a shutdown or the lease coordinator that gets the concurrency token exception should mark for shutdown.

BobbyJohansen:
I have done some more analysis with KCL debug on here are the results:


`
I have reduced the logs to just shard 17 which is the one that had the problem occur. It looks like the leaseCounter was off by one in these logs. There are 16 applications running and balancing leases across the stream, it looks like in the screenshot provided that the other consumers thought the current lease leaseCounter was one higher than what the shard consumer consuming the shard thought it was. 

i.e when the shard consumer for shard 17 tried to prepare the checkpoint we get this error: DEBUG [ShardRecordProcessor-0001] [software.amazon.kinesis.leases.dynamodb.DynamoDBLeaseRefresher] [shardId-000000000017] Lease update failed for lease with key shardId-000000000017 because the lease counter was not 15505988

and a previous scan of the dynamo table revealed 20 milliseconds earlier that the lease looked like this (where the leaseCounter is 15505989) :
[...]

PR #538, opened 2019-04-07
Allow RecordProcessors To bubble up shutdown exception

BobbyJohansen:
*Issue #, if available:*
https://github.com/awslabs/amazon-kinesis-client/issues/537
*Description of changes:*
This change is aimed to allow the ShardRecordProcessor bubble up an unrecoverable exception given the current consumer state i.e issues with checkpointing or bugs around the lease coordinator. The goal of the unrecoverable exception is to make the scheduler re-create a shard consumer and refresh its concurrency tokens and lease. Fair warning I am not very versed in this library so i may be off basis by allowing this to happen. However it seems like it would be a fair ability to give to the implementer.

The problem listed above may be another root problem that this works around but I am unaware of what is really causing that issue hence the workaround.


BobbyJohansen:
ping. this is still happnening in latest version of the kcl, it also happens with dedicated resources to each consumer application.

BobbyJohansen:
Moved to Kafka. Closing because I don't think it will be taken in.

Issue #476, opened 2018-12-13
KCL1.8.10 - NPE in KinesisClientLibLeaseCoordinator.java:265


fstrudel:
Hello,
Sorry for the lack of reproducible steps, we have not found when this happens, but we recently got a large number of NullPointerException from the library.
The machine running the software did not reboot or anything.

```
2018-12-13 16:38:30.083 ERROR 3664 --- [dProcessor-0002] c.a.s.k.c.lib.worker.InitializeTask      : Caught exception: 

java.lang.NullPointerException: null
	at com.amazonaws.services.kinesis.clientlibrary.lib.worker.KinesisClientLibLeaseCoordinator.getCheckpointObject(KinesisClientLibLeaseCoordinator.java:265) ~[amazon-kinesis-client-1.8.10.jar!/:na]
	at com.amazonaws.services.kinesis.clientlibrary.lib.worker.InitializeTask.call(InitializeTask.java:82) ~[amazon-kinesis-client-1.8.10.jar!/:na]
	at com.amazonaws.services.kinesis.clientlibrary.lib.worker.MetricsCollectingTaskDecorator.call(MetricsCollectingTaskDecorator.java:49) [amazon-kinesis-client-1.8.10.jar!/:na]
	at com.amazonaws.services.kinesis.clientlibrary.lib.worker.MetricsCollectingTaskDecorator.call(MetricsCollectingTaskDecorator.java:24) [amazon-kinesis-client-1.8.10.jar!/:na]
	at java.util.concurrent.FutureTask.run(FutureTask.java:266) [na:1.8.0_191]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [na:1.8.0_191]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [na:1.8.0_191]
	at java.lang.Thread.run(Thread.java:748) [na:1.8.0_191]
```

sahilpalvia:
Are you trying to checkpoint in the `initialize` method in the IRecordProcessor implementation?

anoopgarlapati:
@sahilpalvia No, we are not checkpointing in the `initialize` method of our IRecordProcessor implementation.

sahilpalvia:
Do you have debug logs enabled? Can you see `No lease found with key <SHARD_ID>, returning null` in your logs?

[...]

PR #550, opened 2019-05-02
logging message instead of throwing NPE when lease is not found during shard consumer initialization task

aggarwal:
Issue #476

Several users are experiencing an issue where KCL logs an NPE during `InitializingTask` execution for a `ShardConsumer` object. This code change explicitly checks for the `lease` field being `null`, and throws a `KinesisClientLibIOException` instead of `NullPointerException` when it is `null`. The extra log message will help debug this issue further. The current messaging is inadequate.


Cory-Bradshaw:
Closing this PR, as it is a duplicate of an already merged PR #558

PR #558, opened 2019-06-20
Better handling of no lease in getCheckpointObject in KinesisClientLibLeaseCoordinator

anoopgarlapati:
Issue #476  

Description of changes:
Improved the handling of get checkpoint object in KinesisClientLibLeaseCoordinator
by no longer triggering a null pointer exception when lease is null.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibLeaseCoordinator.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibLeaseCoordinator.java
@@ -282,13 +282,16 @@ class KinesisClientLibLeaseCoordinator extends LeaseCoordinator<KinesisClientLea
      */
     @Override
     public Checkpoint getCheckpointObject(String shardId) throws KinesisClientLibException {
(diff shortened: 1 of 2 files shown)


anoopgarlapati:
@sahilpalvia Do you think this is a right approach to solve #476 ? 
We are seeing this often nowadays and the only measure we can take he is deploy a new EC2 instance and kill the existing one on which the worker is causing the problem.

anoopgarlapati:
@micah-jaffe Any timeline/milestone when the next release on 1.x branch would be shipped?

micah-jaffe:
> @micah-jaffe Any timeline/milestone when the next release on 1.x branch would be shipped?

The next release (1.11.0) is now publicly available. We hope this fixes your issue!

abhijit91:
@micah-jaffe I see that this fix is mentioned in the release notes, but the change itself is not included.

aggarwal:
@abhijit91 It's included: https://github.com/awslabs/amazon-kinesis-client/blob/f1a50cd14bc50b1edf98ecac9c06c312e06437b7/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibLeaseCoordinator.java#L280-L293

PR #559, opened 2019-06-21
Add diagnostic events for logging visibility

micah-jaffe:
Added logging into the executor service for visibility and to help debug complex thread-related issues. Also added an error handler for rejected task exceptions in the RxJava layer to propagate those exceptions up to the KCL layer and log them accordingly.

Per discussion with team we are not providing the flexibility to customize logging or add additional handlers as of now. As a result all new classes in this PR are package-private access.

Example of regular executor state diagnostic:
```
16:20:31.746 [Thread-1] INFO  s.a.k.c.DiagnosticEventLogger - Current thread pool executor state: ExecutorStateEvent(executorName=SchedulerThreadPoolExecutor, currentQueueSize=0, activeThreads=1, coreThreads=0, leasesOwned=1, largestPoolSize=2, maximumPoolSize=2147483647)
```

Example of rejected task logging:
```
2019-06-26 16:18:35,044 [main] ERROR s.a.k.c.DiagnosticEventLogger [NONE] - Review your thread configuration to prevent task rejections. Until next release, KCL will not be resilient to task rejections. Current thread pool executor state: ExecutorStateEvent(executorName=SchedulerThreadPoolExecutor, currentQueueSize=0, activeThreads=0, coreThreads=0, leasesOwned=0, largestPoolSize=0, maximumPoolSize=2147483647) 
io.reactivex.exceptions.UndeliverableException: java.util.concurrent.RejectedExecutionException: Test exception.
	at io.reactivex.plugins.RxJavaPlugins.onError(RxJavaPlugins.java:367)
	at software.amazon.kinesis.coordinator.SchedulerTest.lambda$testErrorHandlerForUndeliverableAsyncTaskExceptions$1(SchedulerTest.java:286)
	at org.mockito.internal.stubbing.StubbedInvocationMatcher.answer(StubbedInvocationMatcher.java:34)
	at org.mockito.internal.handler.MockHandlerImpl.handle(MockHandlerImpl.java:91)
	at org.mockito.internal.handler.NullResultGuardian.handle(NullResultGuardian.java:29)
	at org.mockito.internal.handler.InvocationNotifierHandler.handle(InvocationNotifierHandler.java:38)
	at org.mockito.internal.creation.cglib.MethodInterceptorFilter.intercept(MethodInterceptorFilter.java:59)
	at software.amazon.kinesis.coordinator.Scheduler$$EnhancerByMockitoWithCGLIB$$1999942.runProcessLoop(<generated>)
	at software.amazon.kinesis.coordinator.SchedulerTest.testErrorHandlerForUndeliverableAsyncTaskExceptions(SchedulerTest.java:291)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:70)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:50)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:238)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:63)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:236)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:53)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:229)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:309)
	at org.mockito.internal.runners.JUnit45AndHigherRunnerImpl.run(JUnit45AndHigherRunnerImpl.java:37)
	at org.mockito.runners.MockitoJUnitRunner.run(MockitoJUnitRunner.java:62)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:160)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
Caused by: java.util.concurrent.RejectedExecutionException: Test exception.
	... 33 common frames omitted
```

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/CoordinatorConfig.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/CoordinatorConfig.java
@@ -89,11 +89,4 @@ public class CoordinatorConfig {
     private GracefulShutdownCoordinator gracefulShutdownCoordinator = new GracefulShutdownCoordinator();
 
     private CoordinatorFactory coordinatorFactory = new SchedulerCoordinatorFactory();
-
-    /**
-     * The number of milliseconds to sleep in between logging the executor state.
-     *
-     * <p>Default value: 30,000 milliseconds</p>
-     */
-    private long executorDiagnosticsDaemonSleepTimeMillis = 30000L;
 }
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/CoordinatorConfig.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/CoordinatorConfig.java
@@ -90,4 +90,10 @@ public class CoordinatorConfig {
 
     private CoordinatorFactory coordinatorFactory = new SchedulerCoordinatorFactory();
 
+    /**
+     * The number of milliseconds to sleep in between logging the executor state.
+     *
+     * <p>Default value: 30,000 milliseconds</p>
+     */
+    private long executorDiagnosticsDaemonSleepTimeMillis = 30000L;
 }
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/CoordinatorConfig.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/CoordinatorConfig.java
@@ -89,4 +89,5 @@ public class CoordinatorConfig {
     private GracefulShutdownCoordinator gracefulShutdownCoordinator = new GracefulShutdownCoordinator();
 
     private CoordinatorFactory coordinatorFactory = new SchedulerCoordinatorFactory();
+
 }
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DefaultDiagnosticEventHandler.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DefaultDiagnosticEventHandler.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates.
+ * Licensed under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package software.amazon.kinesis.coordinator;
+
+import lombok.NoArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import software.amazon.kinesis.annotations.KinesisClientInternalApi;
+
+@NoArgsConstructor
+@Slf4j
+@KinesisClientInternalApi
+public class DefaultDiagnosticEventHandler implements DiagnosticEventHandler {
+    @Override
+    public void visit(ExecutorStateEvent event) {
+        log.info("Current state of thread pool executor: {}", event);
+    }
+
+    @Override
+    public void visit(RejectedTaskEvent event) {
+        log.info("Undeliverable task exception: {}", event);
+    }
+}
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEvent.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEvent.java
@@ -17,7 +17,7 @@ package software.amazon.kinesis.coordinator;
 /**
  * An interface to implement various types of stateful events that can be used for diagnostics throughout the KCL.
  */
-public interface DiagnosticEvent {
+interface DiagnosticEvent {
     /**
      * DiagnosticEvent is part of a visitor pattern and it accepts DiagnosticEventHandler visitors.
      *
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEvent.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEvent.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates.
+ * Licensed under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package software.amazon.kinesis.coordinator;
+
+import software.amazon.kinesis.annotations.KinesisClientInternalApi;
+
+/**
+ * An interface to implement various types of stateful events that can be used for diagnostics throughout the KCL.
+ */
+@KinesisClientInternalApi
+public interface DiagnosticEvent {
+    /**
+     * DiagnosticEvent is part of a visitor pattern along with DiagnosticEventHandler visitors.
+     *
+     * @param visitor A handler that that controls the behavior of the DiagnosticEvent when invoked.
+     */
+    void accept(DiagnosticEventHandler visitor);
+}
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEvent.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEvent.java
@@ -12,6 +12,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package software.amazon.kinesis.coordinator;
 
 /**
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEvent.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEvent.java
@@ -14,17 +14,19 @@
  */
 package software.amazon.kinesis.coordinator;
 
-import software.amazon.kinesis.annotations.KinesisClientInternalApi;
-
 /**
  * An interface to implement various types of stateful events that can be used for diagnostics throughout the KCL.
  */
-@KinesisClientInternalApi
 public interface DiagnosticEvent {
     /**
-     * DiagnosticEvent is part of a visitor pattern along with DiagnosticEventHandler visitors.
+     * DiagnosticEvent is part of a visitor pattern and it accepts DiagnosticEventHandler visitors.
      *
      * @param visitor A handler that that controls the behavior of the DiagnosticEvent when invoked.
      */
     void accept(DiagnosticEventHandler visitor);
+
+    /**
+     * The string to output to logs when a DiagnosticEvent occurs.
+     */
+    String message();
 }
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEventLogger.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEventLogger.java
@@ -18,17 +18,27 @@ import lombok.NoArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import software.amazon.kinesis.annotations.KinesisClientInternalApi;
 
+/**
+ * Internal implementation of {@link DiagnosticEventHandler} used by {@link Scheduler} to log executor state both
+ * 1) in normal conditions periodically, and 2) in reaction to rejected task exceptions.
+ */
 @NoArgsConstructor
 @Slf4j
 @KinesisClientInternalApi
-public class DefaultDiagnosticEventHandler implements DiagnosticEventHandler {
+public class DiagnosticEventLogger implements DiagnosticEventHandler {
+    /**
+     * {@inheritDoc}
+     */
     @Override
     public void visit(ExecutorStateEvent event) {
-        log.info("Current state of thread pool executor: {}", event);
+        log.info(event.message());
     }
 
+    /**
+     * {@inheritDoc}
+     */
     @Override
     public void visit(RejectedTaskEvent event) {
-        log.info("Undeliverable task exception: {}", event);
+        log.error(event.message(), event.getThrowable());
     }
 }
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEventLogger.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEventLogger.java
@@ -12,6 +12,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package software.amazon.kinesis.coordinator;
 
 import lombok.NoArgsConstructor;
@@ -26,8 +27,8 @@ import software.amazon.kinesis.annotations.KinesisClientInternalApi;
 @Slf4j
 @KinesisClientInternalApi
 class DiagnosticEventLogger implements DiagnosticEventHandler {
-    private static final long executorLogIntervalMillis = 30000L;
-    private long nextExecutorLogTime = System.currentTimeMillis() + executorLogIntervalMillis;
+    private static final long EXECUTOR_LOG_INTERVAL_MILLIS = 30000L;
+    private long nextExecutorLogTime = System.currentTimeMillis() + EXECUTOR_LOG_INTERVAL_MILLIS;
 
     /**
      * {@inheritDoc}
@@ -38,7 +39,7 @@ class DiagnosticEventLogger implements DiagnosticEventHandler {
     public void visit(ExecutorStateEvent event) {
         if (System.currentTimeMillis() >= nextExecutorLogTime) {
             log.info(event.message());
-            nextExecutorLogTime = System.currentTimeMillis() + executorLogIntervalMillis;
+            nextExecutorLogTime = System.currentTimeMillis() + EXECUTOR_LOG_INTERVAL_MILLIS;
         } else {
             log.debug(event.message());
         }
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEventLogger.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEventLogger.java
@@ -25,13 +25,23 @@ import software.amazon.kinesis.annotations.KinesisClientInternalApi;
 @NoArgsConstructor
 @Slf4j
 @KinesisClientInternalApi
-public class DiagnosticEventLogger implements DiagnosticEventHandler {
+class DiagnosticEventLogger implements DiagnosticEventHandler {
+    private static final long executorLogIntervalMillis = 30000L;
+    private long nextExecutorLogTime = System.currentTimeMillis() + executorLogIntervalMillis;
+
     /**
      * {@inheritDoc}
+     *
+     * Only log at info level every 30s to avoid over-logging, else log at debug level
      */
     @Override
     public void visit(ExecutorStateEvent event) {
-        log.info(event.message());
+        if (System.currentTimeMillis() >= nextExecutorLogTime) {
+            log.info(event.message());
+            nextExecutorLogTime = System.currentTimeMillis() + executorLogIntervalMillis;
+        } else {
+            log.debug(event.message());
+        }
     }
 
     /**
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
@@ -24,10 +24,12 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 
 @Getter
-@ToString
+@ToString(exclude = "MESSAGE")
 @Slf4j
 @KinesisClientInternalApi
 public class ExecutorStateEvent implements DiagnosticEvent {
+    private final String MESSAGE = "Current thread pool executor state: ";
+
     private int currentQueueSize;
     private int activeThreads;
     private int coreThreads;
@@ -51,5 +53,10 @@ public class ExecutorStateEvent implements DiagnosticEvent {
     @Override
     public void accept(DiagnosticEventHandler visitor) {
         visitor.visit(this);
+    }
+
+    @Override
+    public String message() {
+        return MESSAGE + this.toString();
     }
 }
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
@@ -12,6 +12,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package software.amazon.kinesis.coordinator;
 
 import lombok.Getter;
@@ -31,6 +32,7 @@ class ExecutorStateEvent implements DiagnosticEvent {
     private final String MESSAGE = "Current thread pool executor state: ";
 
     private boolean isThreadPoolExecutor;
+    private String executorName;
     private int currentQueueSize;
     private int activeThreads;
     private int coreThreads;
@@ -43,6 +45,7 @@ class ExecutorStateEvent implements DiagnosticEvent {
             this.isThreadPoolExecutor = true;
 
             ThreadPoolExecutor ex = (ThreadPoolExecutor) executor;
+            this.executorName = ex.getClass().getSimpleName();
             this.currentQueueSize = ex.getQueue().size();
             this.activeThreads = ex.getActiveCount();
             this.coreThreads = ex.getCorePoolSize();
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
@@ -56,7 +56,6 @@ class ExecutorStateEvent implements DiagnosticEvent {
         this.leasesOwned = leaseCoordinator.getAssignments().size();
     }
 
-
     @Override
     public void accept(DiagnosticEventHandler visitor) {
         // logging is only meaningful for a ThreadPoolExecutor executor service (default config)
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
@@ -20,15 +20,17 @@ import lombok.extern.slf4j.Slf4j;
 import software.amazon.kinesis.annotations.KinesisClientInternalApi;
 import software.amazon.kinesis.leases.LeaseCoordinator;
 
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 
 @Getter
-@ToString(exclude = "MESSAGE")
+@ToString(exclude = {"MESSAGE", "isThreadPoolExecutor"})
 @Slf4j
 @KinesisClientInternalApi
 class ExecutorStateEvent implements DiagnosticEvent {
     private final String MESSAGE = "Current thread pool executor state: ";
 
+    private boolean isThreadPoolExecutor;
     private int currentQueueSize;
     private int activeThreads;
     private int coreThreads;
@@ -36,18 +38,28 @@ class ExecutorStateEvent implements DiagnosticEvent {
     private int largestPoolSize;
     private int maximumPoolSize;
 
-    ExecutorStateEvent(ThreadPoolExecutor executor, LeaseCoordinator leaseCoordinator) {
-        this.currentQueueSize = executor.getQueue().size();
-        this.activeThreads = executor.getActiveCount();
-        this.coreThreads = executor.getCorePoolSize();
-        this.largestPoolSize = executor.getLargestPoolSize();
-        this.maximumPoolSize = executor.getMaximumPoolSize();
+    ExecutorStateEvent(ExecutorService executor, LeaseCoordinator leaseCoordinator) {
+        if (executor instanceof ThreadPoolExecutor) {
+            this.isThreadPoolExecutor = true;
+
+            ThreadPoolExecutor ex = (ThreadPoolExecutor) executor;
+            this.currentQueueSize = ex.getQueue().size();
+            this.activeThreads = ex.getActiveCount();
+            this.coreThreads = ex.getCorePoolSize();
+            this.largestPoolSize = ex.getLargestPoolSize();
+            this.maximumPoolSize = ex.getMaximumPoolSize();
+        }
+
         this.leasesOwned = leaseCoordinator.getAssignments().size();
     }
 
+
     @Override
     public void accept(DiagnosticEventHandler visitor) {
-        visitor.visit(this);
+        // logging is only meaningful for a ThreadPoolExecutor executor service (default config)
+        if (isThreadPoolExecutor) {
+            visitor.visit(this);
+        }
     }
 
     @Override
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates.
+ * Licensed under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package software.amazon.kinesis.coordinator;
+
+import lombok.Getter;
+import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
+import software.amazon.kinesis.annotations.KinesisClientInternalApi;
+import software.amazon.kinesis.leases.LeaseCoordinator;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ThreadPoolExecutor;
+
+@Getter
+@ToString
+@Slf4j
+@KinesisClientInternalApi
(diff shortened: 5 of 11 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/DiagnosticEventInput.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Container for the parameters to a {@link DiagnosticEvent}. Not all parameters may be applicable, depending on the
+ * type of DiagnosticEvent being instantiated.
+ */
+@Builder
+@Getter
+@Accessors(fluent = true)
+@EqualsAndHashCode
+@ToString
+public class DiagnosticEventInput {
```
ashwing: Can we have an event input for each of the event like`ExecutorStateEventInput` and `RejectedTaskEventInput extends ExecutorStateEventInput`

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/DiagnosticEvent.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.commons.lang3.NotImplementedException;
+
+import java.util.function.Consumer;
+
+/**
+ * This class is used to log state information periodically via a background thread in the KCL.
+ */
+@Slf4j
+@ToString
+public abstract class DiagnosticEvent<T extends DiagnosticEvent> {
```
ashwing: We can have a member variable "message" as part of this class.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/DiagnosticEventInput.java
```diff
@@ -0,0 +1,48 @@
...
+    private final ThreadPoolExecutor executor;
+    /**
+     * @param leaseCoordinator contains state information about leases owned by the KCL.
+     */
+    private final LeaseCoordinator leaseCoordinator;
+    /**
+     * @param Throwable allows for hooks that respond to exceptions in {@link RejectedTaskEvent}s.
+     */
+    private final Throwable throwable;
+}
```
ashwing: Need new line

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/ExecutorStateEvent.java
```diff
@@ -0,0 +1,39 @@
...
+
+@Getter
+@ToString
+public class ExecutorStateEvent extends DiagnosticEvent {
+    private int currentQueueSize;
+    private int activeThreads;
+    private int coreThreads;
+    private int leasesOwned;
+
+    protected ExecutorStateEvent(DiagnosticEventInput input) {
```
ashwing: Actually we can add more state information here,
getLargestPoolSize -> Returns the largest number of threads that have ever simultaneously been in the pool.
getMaximumPoolSize -> the maximum allowed number of threads

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/DiagnosticEvent.java
sahilpalvia: There are too many levels of inheritance. Avoid using abstract classes if possible. This could be done as an interface and a implementation.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/ExecutorStateEvent.java
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+package software.amazon.kinesis.coordinator.logging;
+
+import lombok.Getter;
+import lombok.ToString;
+
+@Getter
+@ToString
+public class ExecutorStateEvent extends DiagnosticEvent {
```
sahilpalvia: Could you also add the annotation `@KinesisInternalApi` to this class.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/RejectedTaskEvent.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+package software.amazon.kinesis.coordinator.logging;
+
+import lombok.Getter;
+import lombok.ToString;
+
+@Getter
+@ToString
+public class RejectedTaskEvent extends ExecutorStateEvent {
```
sahilpalvia: Where is this class being used? How is it being set within the KCL?

sahilpalvia: Also would recommend naming classes based on what they are doing. `RejectedTaskEvent` mentions nothing on what the class is doing.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/DiagnosticEvent.java
micah-jaffe: @ashwing do you have any thoughts on this note? The design we originally discussed used inheritance but after talking to Sahil it sounds like the KCL 2.X strongly prefers interfaces

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/RejectedTaskEvent.java
micah-jaffe: This class will be used by an error handler in the Scheduler to react to rejected tasks, but that feature is currently being worked on by @ashwing. We will regroup to consolidate work

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/DiagnosticEventInput.java
micah-jaffe: Per Sahil's comments below we may want to rethink the inheritance design here and use interfaces instead

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/DiagnosticEvent.java
ashwing: We can use interfaces and have a visitor pattern to enforce customer to handle all the events we generate.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/DiagnosticEventInput.java
micah-jaffe: For the sake of simplicity, just left simple constructors for now. Can refactor to an input wrapper if and when necessary

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/ExecutorStateEvent.java
micah-jaffe: Added new state information for getLargestPoolSize and getMaximumPoolSize

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/DiagnosticEvent.java
micah-jaffe: Refactored to use visitor pattern for logging behavior handling, rather than inheritance from abstract class

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/ExecutorStateEvent.java
micah-jaffe: Added to all internal classes in this PR

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/RejectedTaskEvent.java
micah-jaffe: All calls to these classes are now contained within this PR

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DefaultDiagnosticEventHandler.java
```diff
@@ -0,0 +1,34 @@
...
+package software.amazon.kinesis.coordinator;
+
+import lombok.NoArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import software.amazon.kinesis.annotations.KinesisClientInternalApi;
+
+@NoArgsConstructor
+@Slf4j
+@KinesisClientInternalApi
+public class DefaultDiagnosticEventHandler implements DiagnosticEventHandler {
```
sahilpalvia: Avoid using the terms `Default`, `Basic`, etc. It is not clear on what the behavior of the class is. Also could you add javadocs.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/RejectedTaskEvent.java
```diff
@@ -0,0 +1,39 @@
...
+import software.amazon.kinesis.annotations.KinesisClientInternalApi;
+import software.amazon.kinesis.leases.LeaseCoordinator;
+
+import java.util.concurrent.ExecutorService;
+
+@Getter
+@ToString
+@Slf4j
+@KinesisClientInternalApi
+public class RejectedTaskEvent extends ExecutorStateEvent {
```
sahilpalvia: This seems like it could be more of a composition rather than inheritance.

```
public RejectedTaskEvent(ExecutorStateEvent executorStateEvent, Throwable throwable) {
    ...
}
```

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -167,6 +170,9 @@ public Scheduler(@NonNull final CheckpointConfig checkpointConfig,
         this.shardConsumerDispatchPollIntervalMillis = this.coordinatorConfig.shardConsumerDispatchPollIntervalMillis();
         this.parentShardPollIntervalMillis = this.coordinatorConfig.parentShardPollIntervalMillis();
         this.executorService = this.coordinatorConfig.coordinatorFactory().createExecutorService();
+        this.diagnosticEventHandler = new DefaultDiagnosticEventHandler();
```
sahilpalvia: This could come from the config.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -612,6 +620,32 @@ void cleanupShardConsumers(Set<ShardInfo> assignedShards) {
...
+                try {
+                    Thread.sleep(executorDiagnosticsDaemonSleepTimeMillis);
+                } catch (InterruptedException e) {
+                    log.error("Executor diagnostics thread interrupted", e);
+                }
+            }
+        });
+
+        diagnosticsThread.setDaemon(true);
+        diagnosticsThread.start();
```
sahilpalvia: Why not run this in a ScheduledExecutorService? That way it can be made sure that it cleanly shuts down.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/DiagnosticEventsTest.java
```diff
@@ -0,0 +1,152 @@
...
+    private int activeThreadCount;
+    private int corePoolSize;
+    private int largestPoolSize;
+    private int maximumPoolSize;
+
+    private SynchronousQueue<Runnable> executorQueue;
+    private Collection<Lease> leaseAssignments;
+
+    @Before
+    public final void setup() {
```
sahilpalvia: No need for final methods in tests.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/DiagnosticEventsTest.java
```diff
@@ -0,0 +1,152 @@
...
+        customHandler = new CustomHandler();
+        wasCustomHandlerInvoked = false;
+
+        throwable = new TestRejectedTaskException();
+
+        activeThreadCount = 2;
```
sahilpalvia: These values don't seem to change, they can be made finals

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/DiagnosticEventsTest.java
```diff
@@ -0,0 +1,152 @@
...
+        verify(defaultHandler, times(1)).visit(event);
+    }
+
+    @Test
+    public final void testExecutorStateEventWithCustomHandler() {
+        ExecutorStateEvent event = new ExecutorStateEvent(executor, leaseCoordinator);
+        event.accept(customHandler);
+
+        assertTrue(wasCustomHandlerInvoked);
+        wasCustomHandlerInvoked = false;
```
sahilpalvia: Not sure why this statement is the last statement. It doesn't seem to do anything.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEvent.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package software.amazon.kinesis.coordinator;
+
+import software.amazon.kinesis.annotations.KinesisClientInternalApi;
+
+/**
+ * An interface to implement various types of stateful events that can be used for diagnostics throughout the KCL.
+ */
+@KinesisClientInternalApi
+public interface DiagnosticEvent {
```
sahilpalvia: Do not mark interfaces as `@KinesisClientInternalApi`, that is supposed to be for implementations.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/CoordinatorConfig.java
```diff
@@ -90,4 +90,10 @@
 
     private CoordinatorFactory coordinatorFactory = new SchedulerCoordinatorFactory();
 
+    /**
+     * The number of milliseconds to sleep in between logging the executor state.
+     *
+     * <p>Default value: 30,000 milliseconds</p>
+     */
+    private long executorDiagnosticsDaemonSleepTimeMillis = 30000L;
```
sahilpalvia: Not needed: Seems to be a long name, if it can be made shorter would be better.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/DiagnosticEventsTest.java
micah-jaffe: This resets wasCustomHandlerInvoked so that it doesn't start out as true on subsequent tests and give false positives. Will add a comment for this

sahilpalvia: The `@Before` method gets run before every test. The value is being set as false in the before.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
micah-jaffe: Given that we don't want the customer to modify this handler, should this still go in the config?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
micah-jaffe: Per discussion moved this functionality to runProcessLoop() method in Scheduler

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/RejectedTaskEvent.java
micah-jaffe: Changed to use composition rather than inheritance

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DefaultDiagnosticEventHandler.java
micah-jaffe: Changed name to DiagnosticEventLogger and added javadocs

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/DiagnosticEventInput.java
micah-jaffe: Fixed

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/logging/DiagnosticEvent.java
micah-jaffe: Added message member variable

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/DiagnosticEventsTest.java
micah-jaffe: Removed final methods from tests, though there's some inconsistency in the KCL in that some tests use final methods and some do not

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/DiagnosticEventsTest.java
micah-jaffe: You're right, redundant line. Removed it from both tests.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEvent.java
micah-jaffe: Removed `@KinesisClientInternalApi` annotation

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/CoordinatorConfig.java
micah-jaffe: Renamed

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/DiagnosticEventsTest.java
micah-jaffe: Made these values final

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
```diff
@@ -0,0 +1,62 @@
...
+import software.amazon.kinesis.leases.LeaseCoordinator;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ThreadPoolExecutor;
+
+@Getter
+@ToString(exclude = "MESSAGE")
+@Slf4j
+@KinesisClientInternalApi
+public class ExecutorStateEvent implements DiagnosticEvent {
```
yasemin-amzn: ExecutorServiceStateEvent?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private int currentQueueSize;
+    private int activeThreads;
+    private int coreThreads;
+    private int leasesOwned;
+    private int largestPoolSize;
+    private int maximumPoolSize;
+
+    public ExecutorStateEvent(ExecutorService executor, LeaseCoordinator leaseCoordinator) {
+        if (executor instanceof ThreadPoolExecutor) {
```
yasemin-amzn: this method seem to work only for ThreadPoolExecutor, can we take the specific type in the arguments?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/RejectedTaskEvent.java
```diff
@@ -0,0 +1,47 @@
...
+import software.amazon.kinesis.annotations.KinesisClientInternalApi;
+
+@Getter
+@ToString(exclude = "MESSAGE")
+@Slf4j
+@KinesisClientInternalApi
+public class RejectedTaskEvent implements DiagnosticEvent {
+    private final String MESSAGE = "Unexpected task rejection occurred. This could possibly " +
+            "be an issue or a bug. Please search for the exception/error online to check what is " +
+            "going on. If the issue persists or is a recurring problem, feel free to open an issue " +
```
yasemin-amzn: This is a misconfiguration for the thread pool. We currently only support the default thread pool configuration, until the next release.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -126,6 +131,9 @@
 
     private final Object lock = new Object();
 
+    private final long executorLogIntervalMillis = TimeUnit.SECONDS.toMillis(30);
```
yasemin-amzn: let's make this static variable.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -612,6 +623,51 @@ void cleanupShardConsumers(Set<ShardInfo> assignedShards) {
...
 
+    /**
+     * Exceptions in the RxJava layer can fail silently unless an error handler is set to propagate these exceptions
+     * back to the KCL, as is done below. This method is internal to the class and has package access solely
+     * for testing.
+     *
+     * @param handler This method accepts a handler param solely for testing. All non-test calls to this method will
+     *                have a null input.
+     */
+    @VisibleForTesting
```
yasemin-amzn: Can this be tested as part of runProcessLoop() method? 
We want to make sure this is invoked from runProcessLoop.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -612,6 +623,51 @@ void cleanupShardConsumers(Set<ShardInfo> assignedShards) {
...
+               rejectedTaskEvent.accept(diagnosticEventHandler);
+           });
+       } else {
+           RxJavaPlugins.setErrorHandler(t -> handler.accept(t));
+       }
+    }
+
+    private void logExecutorState() {
+        ExecutorStateEvent executorStateEvent = new ExecutorStateEvent(executorService, leaseCoordinator);
+        DiagnosticEventHandler debugHandler = new DiagnosticEventHandler() {
```
yasemin-amzn: we should move all this logic into DiagnosticEventLogger. That's the class responsible for logging details. DiagnosticEventLogger should be the one deciding when to log debug/info at what frequency.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/RejectedTaskEvent.java
micah-jaffe: Changed message to the above

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
micah-jaffe: We can do that. Because the Scheduler's executor service member is typed more generally as ExecutorService, we'll get a compilation error unless we cast it when we create the ExecutorStateEvent, like so:

`new ExecutorStateEvent((ThreadPoolExecutor) executorService, leaseCoordinator);`

Looks a little less pretty to me but I agree that it's probably better to restrict the input type rather than having a bunch of null and meaningless members

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
micah-jaffe: If we restrict the input type per below comment then I think it makes sense to keep this as ExecutorStateEvent or perhaps ThreadPoolExecutorStateEvent (this is a little verbose in my opinion)

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
micah-jaffe: Great call. Moved all timing and level logic into DiagnosticEventLogger

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
micah-jaffe: Changed the unit test for this method to call `RxJavaPlugins.onError()` in runProcessLoop() on a scheduler spy to verify that the error handler works as expected

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
micah-jaffe: Made variable static and moved to DiagnosticEventLogger

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
micah-jaffe: Actually, we can't do that - as discussed force casting a generic ExecutorService into a ThreadPoolExecutor could cause crashes. We can keep the ExecutorService as the input type but only have it take meaningful action (i.e. logging) if that ExecutorService is a ThreadPoolExecutor

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEventLogger.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Internal implementation of {@link DiagnosticEventHandler} used by {@link Scheduler} to log executor state both
+ * 1) in normal conditions periodically, and 2) in reaction to rejected task exceptions.
+ */
+@NoArgsConstructor
+@Slf4j
+@KinesisClientInternalApi
+class DiagnosticEventLogger implements DiagnosticEventHandler {
+    private static final long executorLogIntervalMillis = 30000L;
```
ashwing: CAPS for static final variables

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ThreadPoolExecutor;
+
+@Getter
+@ToString(exclude = {"MESSAGE", "isThreadPoolExecutor"})
+@Slf4j
+@KinesisClientInternalApi
+class ExecutorStateEvent implements DiagnosticEvent {
+    private final String MESSAGE = "Current thread pool executor state: ";
```
ashwing: We need to have the name of the executorService

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/RejectedTaskEvent.java
```diff
@@ -0,0 +1,47 @@
...
+import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
+import software.amazon.kinesis.annotations.KinesisClientInternalApi;
+
+@Getter
+@ToString(exclude = "MESSAGE")
+@Slf4j
+@KinesisClientInternalApi
+class RejectedTaskEvent implements DiagnosticEvent {
+    private final String MESSAGE = "This is a misconfiguration of the thread pool. We currently " +
```
ashwing: This may not be due to misconfiguration. The MESSAGE can be "Review your thread configuration to prevent task rejections. Until next release, KCL will not be resilient to task rejections."

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/SchedulerTest.java
```diff
@@ -269,6 +274,24 @@ public final void testSchedulerShutdown() {
...
     }
 
+    @Test
+    public final void testErrorHandlerForUndeliverableAsyncTaskExceptions() {
+        AtomicBoolean wasHandlerInvoked = new AtomicBoolean(false);
+        Consumer<Throwable> testHandler = t -> wasHandlerInvoked.compareAndSet(false, true);
+
+        Scheduler schedulerSpy = spy(scheduler);
+
+        doAnswer(invocation -> {
```
ashwing: This is same as directly calling RxJavaPlugins.onError(). To address Yasemin's feedback, you need to mock the executor service to reject the task, when ops like subscribe() or onNext() is called. Similar to unit tests in my PR.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/RejectedTaskEvent.java
micah-jaffe: Updated message

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java
micah-jaffe: Updated to include name of executorService

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DiagnosticEventLogger.java
micah-jaffe: Fixed

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/SchedulerTest.java
micah-jaffe: Even when mocking the executor service to reject the tasks, it seems like the exceptions are occurring at the KCL layer still. In your unit tests, the exception from a mocked executor service is:
```java.util.concurrent.RejectedExecutionException```

Whereas using RxJavaPlugins.onError() gives what we're trying to trigger:
```io.reactivex.exceptions.UndeliverableException: java.util.concurrent.RejectedExecutionException: Test exception.```

I set up a failing executor service and got the same result as your tests, with the exception still happening in the KCL layer. I'm not sure it's possible to trigger an exception in the RxJava layer with mocks, or at least not easily. Curious to hear your thoughts?

File: amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java (hunk 20 above)
Comment (yasemin-amzn) [reply amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java]: If we get the scheduler take an EventFactory, then we can mock the event vended to the scheduler, and assert that .accept is invoked on the event upon failure. 

That way, this method can be marked private, and no need to support custom handler as an argument to this method.

File: amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/ExecutorStateEvent.java [hunk 26]
```diff
@@ -0,0 +1,72 @@
...
```
sahilpalvia: Not a blocker: Seems like this is only used in one place, so doesn't need to be a class variable.

```java
public String message() {
    return String.format("Current threadpool executor state: %s", toString());
}
```

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/RejectedTaskEvent.java
```diff
@@ -0,0 +1,48 @@
...
+import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
+import software.amazon.kinesis.annotations.KinesisClientInternalApi;
+
+@Getter
+@ToString(exclude = "MESSAGE")
+@Slf4j
+@KinesisClientInternalApi
+class RejectedTaskEvent implements DiagnosticEvent {
+    private final String MESSAGE = "Review your thread configuration to prevent task rejections. " +
```
sahilpalvia: Same as above. And if a variable is not static and final it doesn't need to be upper case. Seems like you wanted it to be a static final variable.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/DiagnosticEventsTest.java
```diff
@@ -0,0 +1,144 @@
...
+        when(executor.getQueue()).thenReturn(executorQueue);
+        when(executor.getActiveCount()).thenReturn(activeThreadCount);
+        when(executor.getCorePoolSize()).thenReturn(corePoolSize);
+        when(executor.getLargestPoolSize()).thenReturn(largestPoolSize);
+        when(executor.getMaximumPoolSize()).thenReturn(maximumPoolSize);
+        when(leaseCoordinator.getAssignments()).thenReturn(leaseAssignments);
+    }
+
+    @Test
+    public void testExecutorStateEventWithDefaultHandler() {
```
sahilpalvia: Avoid the term default. `testExecutorStateEvent()` should be fine as you are passing in a handler that is not a custom handler.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/RejectedTaskEvent.java
micah-jaffe: Good catch, made static and final

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/DiagnosticEventsTest.java
micah-jaffe: Renamed without 'default'

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/SchedulerTest.java
micah-jaffe: Leaving as is per in-person discussion on scope

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
micah-jaffe: To be able to mock the EventFactory for testability, we would need a public interface for setting/modifying the EventFactory (either in configs or a setter). Currently we don't want users to be able to modify this factory so this would only solve the problem if we introduced a 'backdoor' non-public setter for testing, which still isn't really designing for testability. Unless there's another way to inject a mock that doesn't rely on a public gateway?

micah-jaffe: Refactored to use a factory by overloading with a protected constructor for testing

PR #562, opened 2019-06-27
Fixing bug where initial subscription failure causes shard consumer to get stuck.

yatins47:
Fixing bug where initial subscription failure causes shard consumer to get stuck.

*Issue #, if available:*

*Description of changes:*
This change is to fix the scenario where the initial SubscribeToShard call failing silently in the SDK layer during the state transition and KCL will be not be able to set the variables which causes the health check to retry the failure. The shard consumer will hold the lease and get stuck without making any progress. 
The fix is to update the lastRequestTime variable when we start the subscription to get the health check work in case the initial SubscribeToShard call fail and we don't make progress. 

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriber.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriber.java
@@ -44,6 +44,7 @@ class ShardConsumerSubscriber implements Subscriber<RecordsRetrieved> {
 
     @VisibleForTesting
     final Object lockObject = new Object();
+    // This holds the last time a request to upstream service is made including the first try to establish subscription.
     private Instant lastRequestTime = null;
     private RecordsRetrieved lastAccepted = null;
 
@@ -73,6 +74,8 @@ class ShardConsumerSubscriber implements Subscriber<RecordsRetrieved> {
 
     void startSubscriptions() {
         synchronized (lockObject) {
+            // Setting the lastRequestTime to allow for health checks to restart subscriptions if they failed to
+            // during initial try.
             lastRequestTime = Instant.now();
             if (lastAccepted != null) {
                 recordsPublisher.restartFrom(lastAccepted);
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriber.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriber.java
@@ -44,7 +44,8 @@ class ShardConsumerSubscriber implements Subscriber<RecordsRetrieved> {
 
     @VisibleForTesting
     final Object lockObject = new Object();
-    // This holds the last time a request to upstream service is made including the first try to establish subscription.
+    // This holds the last time an attempt of request to upstream service was made including the first try to
+    // establish subscription.
     private Instant lastRequestTime = null;
     private RecordsRetrieved lastAccepted = null;
 
@@ -131,12 +132,8 @@ class ShardConsumerSubscriber implements Subscriber<RecordsRetrieved> {
                             "{}: Last request was dispatched at {}, but no response as of {} ({}).  Cancelling subscription, and restarting.",
                             shardConsumer.shardInfo().shardId(), lastRequestTime, now, timeSinceLastResponse);
                     cancel();
-                    //
-                    // Set the last request time to now, we specifically don't null it out since we want it to
-                    // trigger a
-                    // restart if the subscription still doesn't start producing.
-                    //
-                    lastRequestTime = Instant.now();
+
+                    // Start the subscription again which will update the lastRequestTime as well.
                     startSubscriptions();
                 }
             }
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriber.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriber.java
@@ -73,6 +73,7 @@ class ShardConsumerSubscriber implements Subscriber<RecordsRetrieved> {
 
     void startSubscriptions() {
         synchronized (lockObject) {
+            lastRequestTime = Instant.now();
             if (lastAccepted != null) {
                 recordsPublisher.restartFrom(lastAccepted);
             }
--- a/amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriberTest.java
+++ b/amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriberTest.java
@@ -22,6 +22,7 @@ import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.argThat;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
@@ -315,12 +316,11 @@ public class ShardConsumerSubscriberTest {
         executorService = Executors.newFixedThreadPool(1, new ThreadFactoryBuilder()
                 .setNameFormat("test-" + testName.getMethodName() + "-%04d").setDaemon(true).build());
 
+        // Mock record publisher which doesn't publish any records on first try which simulates any scenario which
+        // causes first subscription try to fail.
         recordsPublisher = new RecordPublisherWithInitialFailureSubscription();
         subscriber = new ShardConsumerSubscriber(recordsPublisher, executorService, bufferSize, shardConsumer, 0);
         addUniqueItem(1);
-        addTerminalMarker(1);
-
-        CyclicBarrier barrier = new CyclicBarrier(2);
 
         List<ProcessRecordsInput> received = new ArrayList<>();
         doAnswer(a -> {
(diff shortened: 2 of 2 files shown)


amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriberTest.java
```diff
@@ -448,6 +520,31 @@ public void cancel() {
...
+        private int subscriptionTryCount = 0;
+
+        @Override
+        public void subscribe(Subscriber<? super RecordsRetrieved> s) {
+            subscriber = s;
+            ++subscriptionTryCount;
+            s.onSubscribe(new Subscription() {
+                @Override
+                public void request(long n) {
+                    if (subscriptionTryCount == 1) {
```
micah-jaffe: Would it be simpler to rewrite this block as just `if (subscriptionTryCount != 1)`?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriberTest.java
```diff
@@ -311,6 +309,80 @@ public void restartAfterRequestTimerExpiresTest() throws Exception {
...
+        synchronized (processedNotifier) {
+            subscriber.startSubscriptions();
+            processedNotifier.wait(1000);
+        }
+
+        Stream.iterate(2, i -> i + 1).limit(97).forEach(this::addUniqueItem);
+
+        addTerminalMarker(2);
+
+        verify(shardConsumer, times(0)).handleInput(argThat(eqProcessRecordsInput(processRecordsInput)),
```
micah-jaffe: If you're trying to verify that the shardConsumer doesn't initialize successfully, might it be clearer to use the `never()` matcher rather than `timer(0)`?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriberTest.java
```diff
@@ -311,6 +309,80 @@ public void restartAfterRequestTimerExpiresTest() throws Exception {
...
+            processedNotifier.wait(5000);
+        }
+
+        barrier.await(500, TimeUnit.MILLISECONDS);
+
+        verify(shardConsumer, times(100)).handleInput(argThat(eqProcessRecordsInput(processRecordsInput)),
+                any(Subscription.class));
+
+        assertThat(received.size(), equalTo(recordsPublisher.responses.size()));
+        Stream.iterate(0, i -> i + 1).limit(received.size()).forEach(i -> assertThat(received.get(i),
```
micah-jaffe: Could you add a comment for what this test is verifying? It's a little hard to follow

amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriberTest.java
```diff
@@ -311,6 +309,80 @@ public void restartAfterRequestTimerExpiresTest() throws Exception {
 
     }
 
+    @Test
+    public void restartAfterRequestTimerExpiresAfterInitialSubscriptionFailsTest() throws Exception {
```
micah-jaffe: How long does this test take to run? Generally we want to avoid timing-related tests as it increases the build time, but in this case that might be the best way to go about it (not close enough to the issue to say)

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriber.java
```diff
@@ -73,6 +73,7 @@
 
     void startSubscriptions() {
         synchronized (lockObject) {
+            lastRequestTime = Instant.now();
```
ashwing: 1. Can you add a comment on why we setting this here? 
2. Also add a variable comment on lastRequestTime (Last time an upstream request(N) call made to the service)
3. Refactor the places where lastRequestTime = Instant.now() followed by startSubscriptions()
4. Check why shouldn't we set this variable on call request(N) calls made from this class.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriberTest.java
ashwing: The test case is hard to follow and has lot of waits. We need to avoid sleeps. adding comments would be helpful. We can sync up offline to check if the test can be shortened.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriberTest.java
yatins47: Changed to what u suggested. Agree it seems simpler to read.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriberTest.java
yatins47: i didn't knew about never() , new to mockito always been a EasyMock user :)

amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriberTest.java
yatins47: Added comments and simplifying the test too.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriberTest.java
yatins47: I simplified the test. It takes around 60 to 80 ms to run in general which is in line with other tests in the class.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriber.java
yatins47: Added comments. Other points as discussed offline.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriberTest.java
yatins47: I have refactored the unit test and added comments to made it easier to read. It takes for it to 60 to 80 ms to run.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriberTest.java
```diff
@@ -311,6 +310,64 @@ public void restartAfterRequestTimerExpiresTest() throws Exception {
...
+
+        addTerminalMarker(2);
+
+        // Doing the health check to allow the subscription to restart.
+        assertThat(subscriber.healthCheck(1), nullValue());
+
+        // Allow time for processing of the records to end in the executor thread which call notifyAll as it gets the
+        // terminal record.
+        synchronized (processedNotifier) {
+            processedNotifier.wait(500);
```
micah-jaffe: Is 500 millis guaranteed to be enough time to reach the terminal record, even under conditions of high CPU usage or other potential slowdowns? We don't want a test failure based on internals we have no control over. 

Would mocking your executorService and telling it explicitly what to do on each call let you avoid this wait?

yatins47: The test here is actually about sending records from RecordPublishers to the Subscriber through the executor service. The avg time taken to receive all the records from the record publisher is around 60-80 ms hence 500 seems to be a good general timeout. I can update it to a higher value since this is a timeout and having it higher should not affect our test run time.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriber.java
```diff
@@ -44,6 +44,7 @@
 
     @VisibleForTesting
     final Object lockObject = new Object();
+    // This holds the last time a request to upstream service is made including the first try to establish subscription.
```
ashwing: actually it is an attempt to request

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriber.java
ashwing: Line 139 needs to be removed.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriberTest.java
```diff
@@ -311,6 +310,64 @@ public void restartAfterRequestTimerExpiresTest() throws Exception {
...
 
+    @Test
+    public void restartAfterRequestTimerExpiresAfterInitialSubscriptionFailsTest() throws Exception {
+
+        executorService = Executors.newFixedThreadPool(1, new ThreadFactoryBuilder()
+                .setNameFormat("test-" + testName.getMethodName() + "-%04d").setDaemon(true).build());
+
+        // Mock record publisher which doesn't publish any records on first try which simulates any scenario which
+        // causes first subscription try to fail.
+        recordsPublisher = new RecordPublisherWithInitialFailureSubscription();
```
ashwing: Test case looks fine. But can we simulate the failure by mocking the executor service rather than the FailingPublisher, as that closely resembles the failure that we are actually trying to address. It should be a simple change though.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriber.java
yatins47: Modified the wording

amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriberTest.java
yatins47: As discussed offline, added a new test case specifically for Rejection execution exception from executor service.

PR #570, opened 2019-07-17
logging method entry/exit for shard syncer to help debug races faster

aggarwal:
Logging method entry/exit for shard syncer to help debug races faster.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncer.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncer.java
@@ -109,6 +109,7 @@ class ShardSyncer {
             boolean cleanupLeasesOfCompletedShards,
             boolean ignoreUnexpectedChildShards)
             throws DependencyException, InvalidStateException, ProvisionedThroughputException, KinesisClientLibIOException {
+        LOG.info("syncShardLeases: begin");
         List<Shard> shards = getShardList(kinesisProxy);
         LOG.debug("Num shards: " + shards.size());
 
@@ -148,6 +149,7 @@ class ShardSyncer {
                     trackedLeases,
                     leaseManager);
         }
+        LOG.info("syncShardLeases: done");
     }
     // CHECKSTYLE:ON CyclomaticComplexity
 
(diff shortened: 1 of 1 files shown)


sahilpalvia:
Have you tried enabling debug logs on the ShardSyncer class? The logs just seem to be marking the entry and exits of the methods.

aggarwal:
The goal of this change is to make it easier to reconstruct timelines of race conditions that lead to unexpected behavior. Customers don't generally enable debug logging by default, and we'd like to have sufficient data from the very first occurrence of an issue. Since the purpose is solely to reconstruct a timeline, we don't need to log any additional shard information.

sahilpalvia:
Could you elaborate on the race conditions? All the methods that are called from outside the class are synchronized and shouldn't be causing race conditions.

aggarwal:
Sure, the races we're concerned about are generally across hosts. So synchronization within a process does not eliminate the possibility of a race between two concurrently executing shard syncs.

An example:
T0 - Host A starts executing ShardSync
T1 - Host A experiences a GC pause shortly after completing calls to DescribeStream, but before creating any leases. The DescribeStream calls exposed Shard-A without an ending shard sequence number, meaning it is still open.
T2 - Host B starts executing ShardSync, creates new leases, and deletes SHARD_END leases for which child leases have been created. Let's say one such deleted lease is for Shard-A.
T3 - Host A now unpauses after GC, and now determines that it needs to create a lease for Shard-A.

The specific race condition is not important in the above example. I'm simply illustrating how races are possible.

sahilpalvia:
Thanks for that scenairo. But I don't think it should be an issue, as the lease can only be deleted by the ShardConsumer which holds the lease and has processed it to `SHARD_END`. Describing or Listing shards will not delete the lease. The lease will also be deleted only when the children shards have started processing and have checked pointed with a value other than `SHARD_END`.

aggarwal:
This statement is incorrect: "as the lease can only be deleted by the ShardConsumer which holds the lease". See this code: https://github.com/awslabs/amazon-kinesis-client/blob/v1.x/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncer.java#L144-L150

Any worker can delete leases regardless of lease ownership.

sahilpalvia:
Good catch, maybe that could be a better solution to force deletion done by the lease owner. That could help avoid this race condition.

aggarwal:
DDB Streams is going to inject it's own implementation of ShardSyncer, so this PR is no longer relevant.

PR #571, opened 2019-07-18
Adding the option to do periodic shard syncs on elected leader workers instead of shard sync on each shard end event across workers

abhishekkhanna01:
Presently KCL does shard syncs on each shard end event. Due to the shard growth issue in ddb streams, the shard end events are very frequent, resulting in describe streams calls for each such event across workers. This causes throttling for large customers. This throttling delays the shard discovery process, customer's processing and causes other inconsistency issues. This can be avoided if instead of shard syncs being done at shard end events on each worker, periodic shard sync are done and on a limited set of hosts, so that horizontal scaling is feasible for customers. This commit enables the option for periodic shard syncs on a selected set of leader hosts to solve the above problems.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/periodicshardsync/LeaderElectionStrategy.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/periodicshardsync/LeaderElectionStrategy.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates.
+ * Licensed under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.amazonaws.services.kinesis.clientlibrary.lib.periodicshardsync;
+
+import java.util.List;
+import java.util.Set;
+
+import com.amazonaws.services.kinesis.leases.impl.Lease;
+
+/**
+ * Interface for electing leaders and registering listeners interested in being
+ * notified on leader election
+ *
+ * @param <T>
+ */
+public interface LeaderElectionStrategy<T extends Lease> extends Runnable {
+	
+    /**
+     * @param listener Listener interested in being notified on leader election
+     */
+    void registerLeadersElectionListener(LeadersElectionListener listener);
+
+    /**
+     * @param leases Leases held by workers from which the leaders are chsoen
+     * @return Elected set of leaders based on the concrete leader election strategy implementation
+     */
+    Set<String> electLeaders(List<T> leases);
+}
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ActiveShardCountBasedShardSyncStrategyDecider.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ActiveShardCountBasedShardSyncStrategyDecider.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates.
+ * Licensed under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.amazonaws.services.kinesis.clientlibrary.lib.worker;
+
+import java.util.List;
+
+import com.amazonaws.services.kinesis.clientlibrary.types.ExtendedSequenceNumber;
+import com.amazonaws.services.kinesis.leases.exceptions.DependencyException;
+import com.amazonaws.services.kinesis.leases.exceptions.InvalidStateException;
+import com.amazonaws.services.kinesis.leases.exceptions.ProvisionedThroughputException;
+import com.amazonaws.services.kinesis.leases.impl.KinesisClientLease;
+import com.amazonaws.services.kinesis.leases.interfaces.ILeaseManager;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * ShardSyncStrategyDecider implementation which looks at the active shard count
+ * to decide the strategy for shard sync
+ *
+ */
+class ActiveShardCountBasedShardSyncStrategyDecider implements ShardSyncStrategyDecider {
+
+    private static final Object LOCK = new Object();
+    private static final Log LOG = LogFactory.getLog(ActiveShardCountBasedShardSyncStrategyDecider.class);
+    private ILeaseManager<KinesisClientLease> leaseManager;
+    private volatile List<KinesisClientLease> leases;
+
+    ActiveShardCountBasedShardSyncStrategyDecider(ILeaseManager<KinesisClientLease> leaseManager) {
+        this.leaseManager = leaseManager;
+    }
+
+    static final int MIN_ACTIVE_SHARDS_FOR_PERIODIC_SYNC = 3000;
+
+    /*
+     * Gets the active shards count and decides the shard sync strategy based on the threshold
+     *
+     * @see com.amazonaws.services.kinesis.clientlibrary.lib.worker.
+     * ShardSyncStrategyDecider#getShardSyncStrategy()
+     */
+    @Override
+    public ShardSyncStrategy getShardSyncStrategy()
+            throws DependencyException, InvalidStateException, ProvisionedThroughputException {
+        ShardSyncStrategy shardSyncStrategy = null;
+        int activeShardleasesCount = getActiveShardleasesCount();
+        if (activeShardleasesCount != 0) {
+            shardSyncStrategy = activeShardleasesCount >= MIN_ACTIVE_SHARDS_FOR_PERIODIC_SYNC ? ShardSyncStrategy.PERIODIC
+                                        : ShardSyncStrategy.SHARD_END;
+        }
+        LOG.debug(String.format("ActiveShardLeasesCount: %d, ShardSyncStrategy: %s", activeShardleasesCount, shardSyncStrategy.name()));
+        return shardSyncStrategy;
+    }
+
+    /**
+     * Gets the leases and filters out the SHARD_END leases to get the count of
+     * active leases
+     *
+     * @return Active leases count
+     * @throws DependencyException
+     * @throws InvalidStateException
+     * @throws ProvisionedThroughputException
+     */
+    private int getActiveShardleasesCount()
+            throws DependencyException, InvalidStateException, ProvisionedThroughputException {
+        // double checked locking to ensure only a single thread populates the
+        // leases as all threads will get roughly the same view of the leases at
+        // startup. This keeps a check on IOPS utilisation of Leases table. Also
+        // once populated, the strategy will never move back from Periodic to
+        // Shard sync as the number of shards never reduce. For the other case
+        // of increasing shards, its a fair assumption that there would be at
+        // least one deployment and the workers would initialize again before
+        // that happens
+        if(leases == null) {
+            synchronized (LOCK) {
+                if (leases == null) {
+                    getLeases();
+                }
+            }
+        }
+        return leases == null ? 0
+                       : (int) leases.stream().filter(lease -> lease.getCheckpoint() != null && lease.getCheckpoint() != ExtendedSequenceNumber.SHARD_END)
+                                     .count();
+    }
+
+    private void getLeases() throws DependencyException, InvalidStateException, ProvisionedThroughputException {
+        leases = leaseManager.listLeases();
+    }
+
+}
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleLeaderElection.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleLeaderElection.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates.
+ * Licensed under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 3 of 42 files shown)


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -83,6 +85,8 @@
     private static final Log LOG = LogFactory.getLog(Worker.class);
 
     private static final int MAX_INITIALIZATION_ATTEMPTS = 20;
+    private static final int LEADER_POLLER_THREAD_COUNT = 1;
```
aggarwal: Remove unused.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -83,6 +85,8 @@
...
+    private static final int PERIODIC_SHARD_SYNC_THREAD_COUNT = 2;
```
aggarwal: Remove unused.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -651,14 +675,21 @@ private void initialize() {
                     } else {
                         LOG.info("LeaseCoordinator is already running. No need to start it.");
                     }
-                    isDone = true;
+                    LOG.info("Checking the shardSyncStrategy");
+                    shardSyncStrategy = shardSyncStrategyDecider.getShardSyncStrategy();
+                    shardSyncStrategy.onWorkerInitialization();
```
aggarwal: You should move this inside the conditional that confirms that `shardSyncStrategy` is non-null.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncManager.java
```diff
@@ -0,0 +1,122 @@
...
+ */
+@Getter
+class PeriodicShardSyncManager {
+
+    private static final Log LOG = LogFactory.getLog(PeriodicShardSyncManager.class);
+
+    private final String workerId;
+
+    private final PeriodicShardSyncScheduler shardSyncScheduler;
+    private final ScheduledLeaseLeaderPoller leaderPoller;
```
aggarwal: Let's eliminate this reference to `leaderPoller` in this class. The start/stop of the leader poller can happen inside start/stop methods of `shardSyncScheduler`. Doing this will further improve encapsulation.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ScheduledLeaseLeaderPoller.java
```diff
@@ -0,0 +1,74 @@
...
+        this.config = config;
+        this.scheduledExecutor = scheduledThreadPoolExecutor;
+    }
+
+    public Boolean isLeader(String workerId) {
+        return leaderElectionStrategy.isLeader(workerId);
+    }
+
+    public void pollForLeaders() {
+        scheduledExecutor.scheduleWithFixedDelay(leaderElectionStrategy, POLLER_INITIAL_DELAY_MINS,
```
aggarwal: See my comment above about not having `LeaderElectionStrategy` extend `Runnable`. Just use a lambda here to invoke `isLeader(workerId)`. That's a more natural usage.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleLeaderElection.java
```diff
@@ -0,0 +1,144 @@
...
+ * {@code config.getPeriodicShardSyncMaxWorkers()}. Once the leaders are
+ * elected, all the registered listeners are informed about the new election
+ */
+class DeterministicShuffleLeaderElection
+        implements LeaderElectionStrategy<KinesisClientLease>, Comparator<KinesisClientLease> {
+
+    static final String LEAES_CACHE_KEY = "all_leases";
+    // Fixed seed so that the shuffle order is preserved across workers
+    static final int DETERMINISTIC_SHUFFLE_SEED = 1947;
+    private static final int PERIODIC_SHARD_SYNC_MAX_WORKERS = 5;
```
aggarwal: How will this be set differently for Kinesis and DDB? Kinesis wants a single worker.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleLeaderElection.java
```diff
@@ -0,0 +1,144 @@
...
+    @Override
+    public Set<String> electLeaders(List<KinesisClientLease> leases) {
+        if (CollectionUtils.isNullOrEmpty(leases)) {
+            return new HashSet<String>();
+        }
+        leases = leases.stream().filter(lease -> lease.getLeaseOwner() != null).sorted(this)
+                       .collect(Collectors.toList());
+        Collections.shuffle(leases, new Random(DETERMINISTIC_SHUFFLE_SEED));
+        Set<String> leaders = leases.stream()
+                                    .limit(Math.min(PERIODIC_SHARD_SYNC_MAX_WORKERS, getUniqueWorkerIds(leases).size()))
```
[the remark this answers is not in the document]
abhishekkhanna01: Agree. Will update

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleLeaderElection.java
```diff
@@ -0,0 +1,144 @@
...
+     *
+     * @see com.amazonaws.services.kinesis.clientlibrary.lib.periodicshardsync.
+     * interfaces.LeaderElectionStrategy#electLeaders(java.util.List)
+     */
+    @Override
+    public Set<String> electLeaders(List<KinesisClientLease> leases) {
+        if (CollectionUtils.isNullOrEmpty(leases)) {
+            return new HashSet<String>();
+        }
+        leases = leases.stream().filter(lease -> lease.getLeaseOwner() != null).sorted(this)
```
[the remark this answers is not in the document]
abhishekkhanna01: The result of scan would be different on different workers. For the random shuffle to be deterministic, both the seed and the input sequence of leases to the shuffle function should be the same, so sorting to ensure the same input sequence

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ActiveShardCountBasedShardSyncStrategyDecider.java
```diff
@@ -0,0 +1,226 @@
...
+                    getLeases();
+                }
+            }
+        }
+        return leases == null ? 0
+                       : (int) leases.stream().filter(lease -> lease.getCheckpoint() != null && lease.getCheckpoint() != ExtendedSequenceNumber.SHARD_END)
+                                     .count();
+    }
+
+    private void getLeases() throws DependencyException, InvalidStateException, ProvisionedThroughputException {
```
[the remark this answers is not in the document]
abhishekkhanna01: That would imply only one worker could be updating the leases at any given time. We just want that the leasesLastFetchTime should be accessed in a thread safe manner with multiple threads updating the leases concurrently

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ActiveShardCountBasedShardSyncStrategyDecider.java
```diff
@@ -0,0 +1,226 @@
...
+    /**
+     * Gets the leases and filters out the SHARD_END leases to get the count of
+     * active leases
+     *
+     * @return Active leases count
+     * @throws DependencyException
+     * @throws InvalidStateException
+     * @throws ProvisionedThroughputException
+     */
+    private int getActiveShardleasesCount()
```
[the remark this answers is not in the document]
abhishekkhanna01: will update

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleLeaderElection.java
```diff
@@ -0,0 +1,144 @@
...
+        Set<String> workers = new HashSet<String>();
+        for (KinesisClientLease lease : leases) {
+            if (StringUtils.isNotBlank(lease.getLeaseOwner())) {
+                workers.add(lease.getLeaseOwner());
+            }
+        }
+        return workers;
+    }
+
+    void updateLeases() {
```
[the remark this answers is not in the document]
abhishekkhanna01: Same as above

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleLeaderElection.java
```diff
@@ -0,0 +1,144 @@
...
+        Collections.shuffle(leases, new Random(DETERMINISTIC_SHUFFLE_SEED));
+        Set<String> leaders = leases.stream()
+                                    .limit(Math.min(PERIODIC_SHARD_SYNC_MAX_WORKERS, getUniqueWorkerIds(leases).size()))
+                                    .map(lease -> lease.getLeaseOwner()).collect(Collectors.toSet());
+        return leaders;
+    }
+
+    @Override
+    public Boolean isLeader(String workerId) {
+        Set<String> leaders = electLeaders(leases);
```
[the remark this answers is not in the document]
abhishekkhanna01: Agree. Will update

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/LeaderElectionStrategy.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.amazonaws.services.kinesis.clientlibrary.lib.worker;
+
+import java.util.List;
+
+public interface LeaderElectionStrategy {
```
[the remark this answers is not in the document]
abhishekkhanna01: This is unused. Missed deleting the same.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/periodicshardsync/LeaderElectionStrategy.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.amazonaws.services.kinesis.leases.impl.Lease;
+
+/**
+ * Interface for electing leaders and registering listeners interested in being
+ * notified on leader election
+ *
+ * @param <T>
+ */
+public interface LeaderElectionStrategy<T extends Lease> extends Runnable {
```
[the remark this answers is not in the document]
abhishekkhanna01: Agree. Will do

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
abhishekkhanna01: Will do

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
abhishekkhanna01: Will do

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
abhishekkhanna01: getShardSyncStrategy() could return null if the strategy is not able to be determined, in which case isDone is not set to true so that we have another initialization attempt

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncManager.java
abhishekkhanna01: To me these appear to be different logical entities. One responsible for leader election and the other for scheduling the shard syncs

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleLeaderElection.java
abhishekkhanna01: The default value of PERIODIC_SHARD_SYNC_MAX_WORKERS here would be 1 and the ddb streams adapter would inject another implementation of the LeaderElectionStrategy with number of leader workers as 5

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
aggarwal: If `shardSyncStrategy` is `null`, then line 680 will throw an NPE. Let's avoid that.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncManager.java
aggarwal: Yes, and you can maintain logical separation of code in different classes. Using `leaderPoller` directly from `shardSyncScheduler` does not change that. Using it directly there simplifies code.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ActiveShardCountBasedShardSyncStrategyDecider.java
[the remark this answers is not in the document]
aggarwal: You always want to update and read the two fields atomically. Otherwise, there is a risk of the data field corresponding to a different value of the timestamp field. If latency is a concern, perhaps the setting of the field (and not the scanning of leases table) could be made synchronized to minimize blocking.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncManager.java
aggarwal: Also, `shardSyncSchedule` already has a dependency on `leaderPoller`, so you wouldn't be introducing a new dependency either.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
```diff
@@ -166,6 +166,12 @@
      * during incremental deployments of an application).
      */
     public static final boolean DEFAULT_SKIP_SHARD_SYNC_AT_STARTUP_IF_LEASES_EXIST = false;
+    
+    /**
+     * Opt in for periodic shard sync by setting this flag to true. However, setting this to true doesn't guarantee periodic shard sync. The decision
+     * to trigger shard syncs is taken by the ShardSyncStrategyDecider
+     */
+    public static final boolean DEFAULT_ENABLE_PERIODIC_SHARD_SYNC = false;
```
yasemin-amzn: Let's not make this true/false config. Instead, let the customers choose the algorithm with constants. We can introduce another periodic shardsync alternative, that's based off of leader-election table for example. For that one we'd still be using the same config, with another value.

PR #572, opened 2019-07-18
Changes to delete a completed lease only when all its children are at SHARD_END

parijatsinha:
*Description of changes:*
This change ensures that lease for closed shard is deleted from the leases table only after all its children shards have completed processing. It is intended to prevent shard-lineage replay issues caused by race between 2 different workers during a shard sync operation.


parijatsinha:
Closing the pull request to revert an old commit and incorporate feedback.

PR #579, opened 2019-07-30
Changes to support periodic shard sync

parijatsinha:
*Issue #, if available:*

*Description of changes:*


aggarwal:
LGTM.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
```diff
@@ -0,0 +1,184 @@
...
+     * Create an instance overriding the default periodic shard sync worker count.
+     * This is public for use outside KCL.
+     *
+     * @param config                      KinesisClientLibConfiguration instance
+     * @param leaseManager                LeaseManager instance.
+     * @param numPeriodicShardSyncWorkers Max number of periodic shard sync workers.
+     */
+    public DeterministicShuffleShardSyncLeaderDecider(KinesisClientLibConfiguration config,
+            ILeaseManager<KinesisClientLease> leaseManager, int numPeriodicShardSyncWorkers) {
+        this(config, leaseManager, Executors.newScheduledThreadPool(LEADER_DECIDER_THREAD_COUNT),
```
ashwing: Did you look into my previous comment on using fixed threadpool?

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
```diff
@@ -0,0 +1,184 @@
...
+        this.numPeriodicShardSyncWorkers = numPeriodicShardSyncWorkers;
+    }
+
+    /*
+     * Shuffles the leases deterministically and elects numPeriodicShardSyncWorkers number of workers
+     * as leaders (workers that will perform shard sync).
+     */
+    private void electLeaders() {
+        try {
+            LOG.debug("Started leader election: " + System.currentTimeMillis());
```
ashwing: We generally use 'Instant' to capture time in KCL

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
```diff
@@ -0,0 +1,184 @@
...
+            List<String> uniqueHosts = leases.stream().map(KinesisClientLease::getLeaseOwner)
+                    .filter(owner -> owner != null).distinct().sorted().collect(Collectors.toList());
+
+            Collections.shuffle(uniqueHosts, new Random(DETERMINISTIC_SHUFFLE_SEED));
+            int numShardSyncWorkers = Math.min(uniqueHosts.size(), numPeriodicShardSyncWorkers);
+            // In case value is currently being read, we wait for reading to complete before updating the variable.
+            // This is to prevent any ConcurrentModificationException exceptions.
+            readWriteLock.writeLock().lock();
+            leaders = new HashSet<>(uniqueHosts.subList(0, numShardSyncWorkers));
+            LOG.debug("Elected leaders: " + String.join(", ", leaders));
```
ashwing: Please use the placeholder {} while logging as we have used elsewhere in KCL.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncManager.java
```diff
@@ -0,0 +1,142 @@
...
+        }
+        return new TaskResult(null);
+    }
+
+    public void stop() {
+        if (isRunning) {
+            LOG.info(String.format("Shutting down leader decider on worker %s", workerId));
+            leaderDecider.shutdown();
+            LOG.info(String.format("Shutting down periodic shard sync task scheduler on worker %s", workerId));
+            shardSyncThreadPool.shutdown();
```
ashwing: why shouldn't we use awaitTermination before shuttingdown? why would this not cause any intermittent inconsistency?

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncManager.java
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+
+    private void runShardSync() {
+        try {
+            if (leaderDecider.isLeader(workerId)) {
+                LOG.debug(String.format("WorkerId %s is a leader, running the shard sync task", workerId));
+                shardSyncTask.run();
+            } else {
+                LOG.debug(String.format("WorkerId %s is not a leader, not running the shard sync task", workerId));
```
ashwing: Same comment on preferred logging format applicable to all places

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncManager.java
```diff
@@ -0,0 +1,142 @@
...
+            }
+        } catch (Throwable t) {
+            LOG.error("Error during runShardSync.", t);
+        }
+    }
+
+    @Override public int hashCode() {
```
ashwing: Please use lombok to avoid the boiler plate code like hashcode and equals

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncStrategy.java
```diff
@@ -0,0 +1,38 @@
...
+public class PeriodicShardSyncStrategy implements ShardSyncStrategy {
+
+    public static final String NAME = "PeriodicShardSyncStrategy";
+    private PeriodicShardSyncManager periodicShardSyncManager;
+
+    public PeriodicShardSyncStrategy(PeriodicShardSyncManager periodicShardSyncManager) {
+        this.periodicShardSyncManager = periodicShardSyncManager;
+    }
+
+    @Override public String getName() {
```
ashwing: why do we have a instance method for a public static variable?

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardEndShardSyncStrategy.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public static final String NAME = "ShardEndShardSyncStrategy";
+    private static final Log LOG = LogFactory.getLog(Worker.class);
+    private ShardSyncTaskManager shardSyncTaskManager;
+
+    public ShardEndShardSyncStrategy(ShardSyncTaskManager shardSyncTaskManager) {
+        this.shardSyncTaskManager = shardSyncTaskManager;
+    }
+
+    @Override public String getName() {
```
ashwing: same as above comment

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
```diff
@@ -0,0 +1,184 @@
...
+    /**
+     * Create an instance using the default periodic shard sync worker count (1).
+     * This constructor is package-private to ensure external users use the public
+     * constructor and are thereby aware of the number of periodic shard sync workers
+     * they want in their application.
+     *
+     * @param config       KinesisClientLibConfiguration instance
+     * @param leaseManager LeaseManager instance.
+     */
+    DeterministicShuffleShardSyncLeaderDecider(KinesisClientLibConfiguration config,
```
sahilpalvia: Is there a reason for new class to already have chained constructors?

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
```diff
@@ -0,0 +1,184 @@
...
+            Collections.shuffle(uniqueHosts, new Random(DETERMINISTIC_SHUFFLE_SEED));
+            int numShardSyncWorkers = Math.min(uniqueHosts.size(), numPeriodicShardSyncWorkers);
+            // In case value is currently being read, we wait for reading to complete before updating the variable.
+            // This is to prevent any ConcurrentModificationException exceptions.
+            readWriteLock.writeLock().lock();
+            leaders = new HashSet<>(uniqueHosts.subList(0, numShardSyncWorkers));
+            LOG.debug("Elected leaders: " + String.join(", ", leaders));
+            LOG.debug("Completed leader election: " + System.currentTimeMillis());
+        } catch (DependencyException | InvalidStateException | ProvisionedThroughputException e) {
+            LOG.error("Exception occurred while trying to fetch all leases for leader election", e);
```
sahilpalvia: What is happening during these exceptions? Is it getting retried? These errors seem like known conditions, and safe thing is to retry, so it can be warning instead of error

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
```diff
@@ -0,0 +1,184 @@
...
+    private boolean leadersNullOrEmpty() {
+        try {
+            readWriteLock.readLock().lock();
+            return CollectionUtils.isNullOrEmpty(leaders);
+        } finally {
+            readWriteLock.readLock().unlock();
+        }
+    }
+
+    private boolean workerIdInLeaders(String workerId) {
```
sahilpalvia: Could you rename this method, it is not clear what is it doing.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncStrategy.java
```diff
@@ -0,0 +1,38 @@
...
+        return NAME;
+    }
+
+    @Override public TaskResult syncShards() {
```
sahilpalvia: nit: Could you have `Override` in its own line.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardEndShardSyncStrategy.java
```diff
@@ -0,0 +1,53 @@
+package com.amazonaws.services.kinesis.clientlibrary.lib.worker;
+
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+public class ShardEndShardSyncStrategy implements ShardSyncStrategy {
```
sahilpalvia: Does this class need to be public? Could you add javadoc to the changes.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncTask.java
```diff
@@ -27,7 +27,7 @@
  * It will clean up leases/activities for shards that have been completely processed (if
  * cleanupLeasesUponShardCompletion is true).
  */
-class ShardSyncTask implements ITask {
+class ShardSyncTask implements ITask, Runnable {
```
sahilpalvia: `ITask` is already a `Callable`, you don't need a `Runnable` to schedule it to a thread.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
parijatsinha: Changed.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncManager.java
parijatsinha: Switched to "@EqualsAndHashCode"

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncTask.java
parijatsinha: Yes it was redundant (leftover from previous commit on our fork). Removed.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncStrategy.java
parijatsinha: Changed.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardEndShardSyncStrategy.java
parijatsinha: Changed it to package-private.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
parijatsinha: Changed to leadersContainsWorkerId()

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
parijatsinha: Retries are within AWS SDK. If it's still thrown, it means retries couldn't get past the exception. Logging an error makes sense in that case.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncStrategy.java
parijatsinha: made it private.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardEndShardSyncStrategy.java
parijatsinha: made it private.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncManager.java
parijatsinha: Discussed offline.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
parijatsinha: Discussed offline.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
parijatsinha: Changed to singleThreadedScheduledExecutor.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
parijatsinha: Why not?

sahilpalvia: Are the chained constructors used for a reason? Just because we can add code doesn't mean we do. Its for maintenance purposes, if its a new class we can create the exact constructors that we need.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
sahilpalvia: In that case we could probably keep a count of how many times did the retry fail consecutively and if it more that X times then log as errors.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncManager.java
parijatsinha: Can't think of any issues due to intermittent inconsistencies at shutdown.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
parijatsinha: As discussed, we are going with the implemented design for now.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncStrategy.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Facade of methods that can be invoked at different points
+ * in KCL application execution to perform certain actions related to shard-sync.
+ */
+public interface ShardSyncStrategy {
+
+    enum StrategyType {
+        PERIODIC,
+        SHARD_END
```
yasemin-amzn: How is initial ShardSync (after deployment etc..) controlled with this change? It's running shardSync on all hosts independent of the StrategyType according to the ShardSyncAtStartup configuration?

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
```diff
@@ -167,6 +167,11 @@
      */
     public static final boolean DEFAULT_SKIP_SHARD_SYNC_AT_STARTUP_IF_LEASES_EXIST = false;
 
+    /**
+     * Opt in for periodic shard sync by setting this flag to true.
```
ashwing: Change this comment to fit the enum

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
```diff
@@ -0,0 +1,184 @@
...
+ * This ensures redundancy for shard-sync during host failures.
+ */
+public class DeterministicShuffleShardSyncLeaderDecider implements LeaderDecider {
+    // Fixed seed so that the shuffle order is preserved across workers
+    static final int DETERMINISTIC_SHUFFLE_SEED = 1947;
+    static final int PERIODIC_SHARD_SYNC_MAX_WORKERS_DEFAULT = 1; //Default for KCL.
+    private static final Log LOG = LogFactory.getLog(DeterministicShuffleShardSyncLeaderDecider.class);
+
+    private static final long ELECTION_INITIAL_DELAY_MILLIS = 60 * 1000;
+    private static final long ELECTION_SCHEDULING_INTERVAL = 5 * 60 * 1000;
```
ashwing: INTERVAL_MILLIS

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
```diff
@@ -0,0 +1,184 @@
...
+        this.numPeriodicShardSyncWorkers = numPeriodicShardSyncWorkers;
+    }
+
+    /*
+     * Shuffles the leases deterministically and elects numPeriodicShardSyncWorkers number of workers
+     * as leaders (workers that will perform shard sync).
+     */
+    private void electLeaders() {
+        try {
+            LOG.debug("Started leader election:" + Instant.now());
```
ashwing: "election at : "

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
```diff
@@ -0,0 +1,184 @@
...
+        } catch (Throwable t) {
+            LOG.error("Unknown exception during leader election.", t);
+        } finally {
+            readWriteLock.writeLock().unlock();
+        }
+    }
+
+    // Utility methods to ensure we acquire a readLock in case an election has completed
+    // and some thread is trying to update leaders variable.
+    private boolean leadersNullOrEmpty() {
```
ashwing: both leadersNullOrEmpty() and leadersContainWorkerId, can use a common read lock acquire mthod by using a BooleanSupplier

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/LeaderDecider.java
```diff
@@ -0,0 +1,39 @@
...
+ * worker is one of the leaders designated to execute shard-sync and then acts accordingly.
+ */
+public interface LeaderDecider {
+
+    /**
+     * Method invoked to check the given workerId corresponds to one of the workers
+     * designated to execute shard-syncs periodically.
+     *
+     * @param workerId ID of the worker
+     * @return True of the worker with ID workerId can execute shard-sync. False otherwise.
```
ashwing: nit True if

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
parijatsinha: We have 3 constructors - a default one with all default params (except config and lease manager), an overriding constructor that allows setting the number of shard sync workers (in case you want to go over 1, or allow customers override it, in future), and an all-args constructor for unit tests. I chained them to ensure less repetitive code. If you think any of these are un-necessary, or would like to structure it differently, I can make that change.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncStrategy.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * Can be used to provide a custom name for the implemented strategy.
+     *
+     * @return Name of the strategy.
+     */
+    String getName();
```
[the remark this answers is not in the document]
parijatsinha: Changed to ShardSyncStrategyType.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncStrategy.java
parijatsinha: We were discussing this yesterday - since all workers would ShardSync at initialization, it leads to throttling from DescribeStream for large table streams. This further delays initialization. One option that came up was to do it probabilistically so only a fraction of workers do it initially.
We can take it up as a future optimization. Let me know if you have any other ideas.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibConfiguration.java
parijatsinha: Changed.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
parijatsinha: Changed.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
parijatsinha: Changed.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/LeaderDecider.java
parijatsinha: Changed.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
parijatsinha: Changed to recommended pattern.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
parijatsinha: Changed it to have just one constructor.

parijatsinha:
Test run results.

Results :

Tests run: 361, Failures: 0, Errors: 0, Skipped: 1

[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 03:06 min
[INFO] Finished at: 2019-08-02T16:59:21-07:00
[INFO] Final Memory: 35M/449M
[INFO] ------------------------------------------------------------------------

Process finished with exit code 0

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncManager.java
```diff
@@ -0,0 +1,128 @@
...
+                shardSyncTask.call();
+            } else {
+                LOG.debug(String.format("WorkerId %s is not a leader, not running the shard sync task", workerId));
+            }
+        } catch (Throwable t) {
+            LOG.error("Error during runShardSync.", t);
+        }
+    }
+
+    static class PeriodicShardSyncManagerBuilder {
```
sahilpalvia: This builder logic looks a bit off, why not use the lombok `@Builder` annotation? You could also add `@NonNull` to the values that shouldn't be null, that way the constructor becomes simpler.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncStrategyType.java
```diff
@@ -0,0 +1,11 @@
+package com.amazonaws.services.kinesis.clientlibrary.lib.worker;
+
+/**
+ * Types of ShardSyncStrategy} implemented in KCL.
+ */
+public enum ShardSyncStrategyType {
+
+        PERIODIC, /* Shard sync are performed periodically */
```
sahilpalvia: nit: Could the Javadoc be on the line above the value.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -470,7 +466,8 @@ public Worker(
         this(applicationName, recordProcessorFactory, config, streamConfig, initialPositionInStream, parentShardPollIntervalMillis,
                 shardSyncIdleTimeMillis, cleanupLeasesUponShardCompletion, checkpoint, leaseCoordinator, execService,
                 metricsFactory, taskBackoffTimeMillis, failoverTimeMillis, skipShardSyncAtWorkerInitializationIfLeasesExist,
-                shardPrioritization, Optional.empty(), Optional.empty(), DEFAULT_WORKER_STATE_CHANGE_LISTENER, DEFAULT_LEASE_CLEANUP_VALIDATOR );
+                shardPrioritization, Optional.empty(), Optional.empty(), DEFAULT_WORKER_STATE_CHANGE_LISTENER,
+                DEFAULT_LEASE_CLEANUP_VALIDATOR, null/* leaderDecider */);
```
sahilpalvia: Could you remove the comment from inside the call.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -659,6 +701,7 @@ private void initialize() {
                 LOG.error("Caught exception when initializing LeaseCoordinator", e);
                 lastException = e;
             } catch (Exception e) {
+                LOG.error("Caught exception when initializing worker", e);
```
sahilpalvia: Let's not change the logging behavior here. This will keep on spitting errors for all retry attempts. Could you remove this change.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -1286,14 +1363,32 @@ public Worker build() {
                 workerStateChangeListener = DEFAULT_WORKER_STATE_CHANGE_LISTENER;
             }
 
-            if(leaseCleanupValidator == null) {
+            if (shardSyncer == null) {
```
sahilpalvia: This will override the behavior of the lease clean up validator if it is already set. Would be better to do this in 2 checks

```java
if (leaseCleanupValidator == null) {
    leaseCleanupValidator = DEFAULT_LEASE_CLEANUP_VALIDATOR;
}

if (shardSyncer == null) {
    shardSyncer = new KinesisShardSyncer(leaseCleanupValidator);
}
```

src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseCoordinator.java
```diff
@@ -174,7 +173,7 @@ public LeaseCoordinator(ILeaseManager<T> leaseManager,
      * @param maxLeasesToStealAtOneTime Steal up to these many leases at a time (for load balancing)
      * @param metricsFactory Used to publish metrics about lease operations
      */
-    public LeaseCoordinator(ILeaseManager<T> leaseManager,
+    LeaseCoordinator(ILeaseManager<T> leaseManager,
```
sahilpalvia: The `public` can't be removed. This class is a public class, this would be a breaking change.

src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseCoordinator.java
```diff
@@ -437,10 +438,10 @@ public boolean updateLease(T lease, UUID concurrencyToken)
      * @param maximumPoolSize Maximum allowed thread pool size
      * @return Executor service that should be used for lease renewal.
      */
-    private static ExecutorService getLeaseRenewalExecutorService(int maximumPoolSize) {
+    public static ExecutorService getDefaultLeaseRenewalExecutorService(int maximumPoolSize) {
```
sahilpalvia: Does this need to be public? Seems like this is only being used in the Worker, so making it package private would be better.

src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseCoordinator.java
```diff
@@ -186,44 +185,46 @@ public LeaseCoordinator(ILeaseManager<T> leaseManager,
...
-     * @param leaseManager LeaseManager instance to use
-     * @param leaseSelector LeaseSelector instance to use
-     * @param workerIdentifier Identifies the worker (e.g. useful to track lease ownership)
-     * @param leaseDurationMillis Duration of a lease
-     * @param epsilonMillis Allow for some variance when calculating lease expirations
-     * @param maxLeasesForWorker Max leases this Worker can handle at a time
-     * @param maxLeasesToStealAtOneTime Steal up to these many leases at a time (for load balancing)
-     * @param metricsFactory Used to publish metrics about lease operations
-     */
     public LeaseCoordinator(ILeaseManager<T> leaseManager,
```
sahilpalvia: No need for this to be `public`. Could you also add javadocs.

src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseCoordinator.java
```diff
@@ -186,44 +185,46 @@ public LeaseCoordinator(ILeaseManager<T> leaseManager,
...
+                        .withMaxLeasesToStealAtOneTime(maxLeasesToStealAtOneTime),
+                new LeaseRenewer<>(leaseManager, workerIdentifier, leaseDurationMillis,  getDefaultLeaseRenewalExecutorService(maxLeaseRenewerThreadCount)),
+                leaseDurationMillis,
+                epsilonMillis,
+                maxLeasesForWorker,
+                maxLeasesToStealAtOneTime,
+                metricsFactory);
+    }
+
+    public LeaseCoordinator(final ILeaseTaker<T> leaseTaker,
```
sahilpalvia: Javadocs missing.

src/main/java/com/amazonaws/services/kinesis/leases/interfaces/ILeaseRenewer.java
```diff
@@ -97,4 +97,9 @@
     boolean updateLease(T lease, UUID concurrencyToken)
         throws DependencyException, InvalidStateException, ProvisionedThroughputException;
 
+    /**
+     * Shutdown any clients and thread-pools.
+     */
+    void shutdown();
```
sahilpalvia: Could you add a default behavior or else this would become a breaking change.

```java
default void shutdown() {
    // Does nothing
}
```

src/test/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncTestBase.java
```diff
@@ -0,0 +1,41 @@
...
+package com.amazonaws.services.kinesis.clientlibrary.lib.worker;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+
+import com.amazonaws.services.kinesis.clientlibrary.types.ExtendedSequenceNumber;
+import com.amazonaws.services.kinesis.leases.impl.KinesisClientLease;
+
+public class PeriodicShardSyncTestBase {
```
sahilpalvia: No need to create this base class it is only used in one test. Just add it to the test class. It is easier to understand test cases if all code is contained.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
```diff
@@ -0,0 +1,151 @@
...
+            readWriteLock.writeLock().unlock();
+        }
+    }
+
+    private boolean isWorkerLeaderForShardSync(String workerId) {
+        return CollectionUtils.isNullOrEmpty(leaders) ||
+            (!CollectionUtils.isNullOrEmpty(leaders) && leaders.contains(workerId));
+    }
+
+    @Override public synchronized Boolean isLeader(String workerId) {
```
sahilpalvia: Nit: could you make sure `@Override` is in the line above the method. And also could you make sure all annotations in the PR are in the line above.

src/test/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/WorkerTest.java
```diff
@@ -161,7 +164,8 @@
 
     private RecordsFetcherFactory recordsFetcherFactory;
     private KinesisClientLibConfiguration config;
-    private ShardSyncer shardSyncer = new ShardSyncer(new KinesisLeaseCleanupValidator());
+    private KinesisShardSyncer shardSyncer = new KinesisShardSyncer(new KinesisLeaseCleanupValidator());
```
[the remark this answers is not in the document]
parijatsinha: Don't quite understand what you mean here. KinesisShardSyncer uses kinesisproxy and leasemanager for all data operations, and only contains business logic. As long kinesisproxy and leasemanager are mocked/stubbed correctly, we should be able to use KinesisShardSyncer as-is in tests.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncManager.java
parijatsinha: I thought about using it. We would still need a Builder to set the default value for the executor. Explicitly setting the values makes it easy to understand and the reader doesn't have to get familiar with syntactic-sugar libraries.

src/main/java/com/amazonaws/services/kinesis/leases/interfaces/ILeaseRenewer.java
parijatsinha: Will add.  Not sure why this interface was public, since it wasn't injectable. Would this be a breaking change if it wasn't injectable?

src/test/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShutdownTaskTest.java
```diff
@@ -57,10 +59,13 @@
             defaultParentShardIds,
             ExtendedSequenceNumber.LATEST);
     IRecordProcessor defaultRecordProcessor = new TestStreamlet();
-    ShardSyncer shardSyncer = new ShardSyncer(new KinesisLeaseCleanupValidator());
+    ShardSyncer shardSyncer = new KinesisShardSyncer(new KinesisLeaseCleanupValidator());
```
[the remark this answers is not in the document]
parijatsinha: Discussed offline.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncManager.java
parijatsinha: As discussed, removed Builder and added 2 chained constructors.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/ShardSyncStrategyType.java
parijatsinha: Changed.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
parijatsinha: Removed.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
parijatsinha: Changing it since it's already been exposed to customers in an earlier release.

I had done this initially but changed it to just basing both object's creation on shardSyncer. leaseCleanupValidator is used only from within ShardSyncer and hence it doesn't make sense to set it separately if we are anyway injecting ShardSyncer.

src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseCoordinator.java
parijatsinha: Made public.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
parijatsinha: Removed.

src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseCoordinator.java
parijatsinha: Worker is in a different package (com.amazonaws.services.kinesis.clientlibrary.lib.worker) than LeaseCoordinator (com.amazonaws.services.kinesis.leases.impl). We would need to keep it public.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncManager.java
parijatsinha: Removed builder as discussed offline.

src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseCoordinator.java
parijatsinha: Removed public. Added Javadocs.

src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseCoordinator.java
parijatsinha: Added Javadocs.

src/main/java/com/amazonaws/services/kinesis/leases/interfaces/ILeaseRenewer.java
parijatsinha: Changed.

src/test/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/PeriodicShardSyncTestBase.java
parijatsinha: Removed.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/DeterministicShuffleShardSyncLeaderDecider.java
parijatsinha: Moved.

parijatsinha: Changed.

src/test/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/WorkerTest.java
```diff
@@ -516,6 +522,8 @@ public final void testOneSplitShard2Threads() throws Exception {
      * Runs worker with threadPoolSize < numShards
      * Test method for {@link Worker#run()}.
      */
+    //TODO: Fix this failing test
+    @Ignore
```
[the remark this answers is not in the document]
parijatsinha: Fixed the test.

PR #585, opened 2019-08-08
Add logging around shard end codepaths

micah-jaffe:
KinesisDataFetcher can internally decide it's reached the end of a shard through 4 different codepaths. The logs currently do not provide insight into which codepath a shard end was reached from, making it difficult to debug. This PR adds unique logging items for each codepath where a KinesisDataFetcher can decide it has reached the end of a shard.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisDataFetcher.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisDataFetcher.java
@@ -71,12 +71,15 @@ class KinesisDataFetcher {
         
         if (nextIterator != null) {
             try {
-                return new AdvancingResult(kinesisProxy.get(nextIterator, maxRecords));
+                GetRecordsResult getRecordsResult = kinesisProxy.get(nextIterator, maxRecords);
+                LOG.info("Successfully fetched records from Kinesis for shard " + shardId + ": " + getRecordsResult);
+                return new AdvancingResult(getRecordsResult);
             } catch (ResourceNotFoundException e) {
                 LOG.info("Caught ResourceNotFoundException when fetching records for shard " + shardId);
                 return TERMINAL_RESULT;
             }
         } else {
+            LOG.info("Cannot fetch records from Kinesis for shard " + shardId + ": nextIterator is null.");
             return TERMINAL_RESULT;
         }
     }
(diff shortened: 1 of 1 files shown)


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisDataFetcher.java
```diff
@@ -71,12 +71,15 @@ public DataFetcherResult getRecords(int maxRecords) {
         
         if (nextIterator != null) {
             try {
-                return new AdvancingResult(kinesisProxy.get(nextIterator, maxRecords));
+                GetRecordsResult getRecordsResult = kinesisProxy.get(nextIterator, maxRecords);
+                LOG.info("Successfully fetched records from Kinesis for shard " + shardId + ": " + getRecordsResult);
```
yasemin-amzn: Logging every response at INFO level can be too much for existing applications. Let's log at INFO only when nextIterator is null, which is the interesting case.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisDataFetcher.java
```diff
@@ -71,12 +71,15 @@ public DataFetcherResult getRecords(int maxRecords) {
...
+                return new AdvancingResult(getRecordsResult);
             } catch (ResourceNotFoundException e) {
                 LOG.info("Caught ResourceNotFoundException when fetching records for shard " + shardId);
                 return TERMINAL_RESULT;
             }
         } else {
+            LOG.info("Cannot fetch records from Kinesis for shard " + shardId + ": nextIterator is null.");
```
yasemin-amzn: This is more of a skipping to fetch case, instead of can not. Re-phrasing would be nice.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisDataFetcher.java
micah-jaffe: Made the change

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisDataFetcher.java
micah-jaffe: Good callout - rephrased the log

PR #596, opened 2019-08-17
Preventing duplicate delivery due to unacknowledged event while completing the subscription

ashwing:
Preventing duplicate delivery due to unacknowledged event while completing the subscription

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
@@ -242,7 +242,7 @@ public class FanOutRecordsPublisher implements RecordsPublisher {
     private void subscribeToShard(String sequenceNumber) {
         synchronized (lockObject) {
             // Clear the delivery queue so that any stale entries from previous subscription are discarded.
-            resetRecordsDeliveryStateOnSubscriptionInit();
+            resetRecordsDeliveryStateOnSubscriptionOnInit();
             SubscribeToShardRequest.Builder builder = KinesisRequestsBuilder.subscribeToShardRequestBuilder()
                     .shardId(shardId).consumerARN(consumerArn);
             SubscribeToShardRequest request;
@@ -332,13 +332,19 @@ public class FanOutRecordsPublisher implements RecordsPublisher {
     }
 
     // This method is not thread safe. This needs to be executed after acquiring lock on this.lockObject
-    private void resetRecordsDeliveryStateOnSubscriptionInit() {
+    private void resetRecordsDeliveryStateOnSubscriptionOnInit() {
         // Clear any lingering records in the queue.
         if (!recordsDeliveryQueue.isEmpty()) {
+            log.warn("{}: Found non-empty queue while starting subscription. This indicates unsuccessful clean up of"
+                    + "previous subscription - {} ", shardId, subscribeToShardId);
             recordsDeliveryQueue.clear();
         }
-        // Set pendingActiveSubscriptionShutdown to default value.
-        pendingActiveSubscriptionShutdown = false;
+        if(pendingActiveSubscriptionShutdown) {
+            log.warn("{}: Found current subscription to be in pendingShutdown state while initializing. This indicates unsuccessful clean up of"
+                    + "previous subscription - {} ", shardId, subscribeToShardId);
+            // Set pendingActiveSubscriptionShutdown to default value.
+            pendingActiveSubscriptionShutdown = false;
+        }
     }
 
     // This method is not thread safe. This needs to be executed after acquiring lock on this.lockObject
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
@@ -64,6 +64,7 @@ public class FanOutRecordsPublisher implements RecordsPublisher {
             ThrowableType.ACQUIRE_TIMEOUT);
     private static final ThrowableCategory READ_TIMEOUT_CATEGORY = new ThrowableCategory(ThrowableType.READ_TIMEOUT);
     private static final int MAX_EVENT_BURST_FROM_SERVICE = 10;
+    private static final long TIME_TO_WAIT_FOR_FINAL_ACK_MILLIS = 1000;
 
     private final KinesisAsyncClient kinesis;
     private final String shardId;
@@ -82,6 +83,7 @@ public class FanOutRecordsPublisher implements RecordsPublisher {
     private long availableQueueSpace = 0;
 
     private BlockingQueue<RecordsRetrievedContext> recordsDeliveryQueue = new LinkedBlockingQueue<>(MAX_EVENT_BURST_FROM_SERVICE);
+    private boolean isAwaitingFinalAckForCurrentSubscription = false;
 
     @Override
     public void start(ExtendedSequenceNumber extendedSequenceNumber,
@@ -132,6 +134,10 @@ public class FanOutRecordsPublisher implements RecordsPublisher {
                 triggeringFlow = evictAckedEventAndScheduleNextEvent(recordsDeliveryAck);
             } catch (Throwable t) {
                 errorOccurred(triggeringFlow, t);
+            } finally {
+                if(isAwaitingFinalAckForCurrentSubscription) {
+                    lockObject.notifyAll();
+                }
             }
             if (triggeringFlow != null) {
                 updateAvailableQueueSpaceAndRequestUpstream(triggeringFlow);
@@ -160,7 +166,7 @@ public class FanOutRecordsPublisher implements RecordsPublisher {
             // Update the triggering flow for post scheduling upstream request.
             flowToBeReturned = recordsRetrievedContext.getRecordFlow();
             // Try scheduling the next event in the queue, if available.
-            if (recordsDeliveryQueue.peek() != null) {
+            if (!isAwaitingFinalAckForCurrentSubscription && recordsDeliveryQueue.peek() != null) {
                 subscriber.onNext(recordsDeliveryQueue.peek().getRecordsRetrieved());
             }
         } else {
@@ -193,7 +199,7 @@ public class FanOutRecordsPublisher implements RecordsPublisher {
             // Note: This does not block wait to enqueue.
             recordsDeliveryQueue.add(recordsRetrievedContext);
             // If the current batch is the only element in the queue, then try scheduling the event delivery.
-            if (recordsDeliveryQueue.size() == 1) {
+            if (!isAwaitingFinalAckForCurrentSubscription && recordsDeliveryQueue.size() == 1) {
                 subscriber.onNext(recordsRetrieved);
             }
         } catch (IllegalStateException e) {
@@ -224,7 +230,7 @@ public class FanOutRecordsPublisher implements RecordsPublisher {
     private void subscribeToShard(String sequenceNumber) {
         synchronized (lockObject) {
             // Clear the queue so that any stale entries from previous subscription are discarded.
-            clearRecordsDeliveryQueue();
+            clearRecordsDeliveryQueue(false);
             SubscribeToShardRequest.Builder builder = KinesisRequestsBuilder.subscribeToShardRequestBuilder()
                     .shardId(shardId).consumerARN(consumerArn);
             SubscribeToShardRequest request;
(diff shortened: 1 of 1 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
```diff
@@ -313,8 +319,24 @@ private void errorOccurred(RecordFlow triggeringFlow, Throwable t) {
         }
     }
 
-    private void clearRecordsDeliveryQueue() {
+    // This method is not thread safe. This needs to be executed after acquiring lock on this.lockObject
+    private void clearRecordsDeliveryQueue(boolean isSubscriptionCompleting) {
```
yatins47: This method seems to be no-op if the delivery q is empty. Just check that in the beginning and return and continue if not.

    if(recordsDeliveryQueue.isEmpty()) {
      return;
    }
    // This will prevent further events from getting scheduled
    isAwaitingFinalAckForCurrentSubscription = true;
    recordsDeliveryQueue.clear();
    if(isSubscriptionCompleting) {
            try {
                // Wait for the configured time to get a notification for already delivered event, if any.
                lockObject.wait(TIME_TO_WAIT_FOR_FINAL_ACK_MILLIS);
            
                Thread.currentThread().interrupt();
            }
        }
Set isAwaitingFinalAckForCurrentSubscription to default value.
isAwaitingFinalAckForCurrentSubscription = false;

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
```diff
@@ -313,8 +319,24 @@ private void errorOccurred(RecordFlow triggeringFlow, Throwable t) {
...
+        if(isSubscriptionCompleting) {
+            // This will prevent further events from getting scheduled
+            isAwaitingFinalAckForCurrentSubscription = true;
+            try {
+                if (!recordsDeliveryQueue.isEmpty()) {
+                    // Wait for the configured time to get a notification for already delivered event, if any.
+                    lockObject.wait(TIME_TO_WAIT_FOR_FINAL_ACK_MILLIS);
```
yatins47: Why are we not waiting if the delivery queue is empty. There might still be a event in which was already delivered?

ashwing: when the delivery queue is empty then it is guaranteed that there will be no valid in-flight event.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
ashwing: Ack. Will make changes

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
```diff
@@ -132,6 +134,10 @@ public void notify(RecordsDeliveryAck recordsDeliveryAck) {
                 triggeringFlow = evictAckedEventAndScheduleNextEvent(recordsDeliveryAck);
             } catch (Throwable t) {
                 errorOccurred(triggeringFlow, t);
+            } finally {
+                if(isAwaitingFinalAckForCurrentSubscription) {
+                    lockObject.notifyAll();
```
yasemin-amzn: This is critical line, worth adding comments explaining the design.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
```diff
@@ -160,7 +166,7 @@ RecordFlow evictAckedEventAndScheduleNextEvent(RecordsDeliveryAck recordsDeliver
             // Update the triggering flow for post scheduling upstream request.
             flowToBeReturned = recordsRetrievedContext.getRecordFlow();
             // Try scheduling the next event in the queue, if available.
-            if (recordsDeliveryQueue.peek() != null) {
+            if (!isAwaitingFinalAckForCurrentSubscription && recordsDeliveryQueue.peek() != null) {
                 subscriber.onNext(recordsDeliveryQueue.peek().getRecordsRetrieved());
```
yasemin-amzn: given we schedule the next event under two different conditions (here and in bufferCurrentEventAndScheduleIfRequired method) , and both required isShuttingDown check, let's move out the schedule call it its own method, with the check. And add comment on why the check is necessary. 

i.e. scheduleNextEvent(event) {if !pendingShutdown; then subscriber.onNext(event)}

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
yasemin-amzn: Couple comment on this method: 

Why it's necessary to introduce split-brain logic here? Do we expect queue to be full, by the time subscribeToShard call is made? That invocation seems to be backup plan. We can either pull that out to its own method, or just let the start of a new subscription wait as well. 

isAwaitingFinalAck variable is tightly couple with the queue status, why not update that within the if? i.e. if !queue.isEmpty -> then set the variable, and lock.wait()

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
```diff
@@ -82,6 +83,7 @@
     private long availableQueueSpace = 0;
 
     private BlockingQueue<RecordsRetrievedContext> recordsDeliveryQueue = new LinkedBlockingQueue<>(MAX_EVENT_BURST_FROM_SERVICE);
+    private boolean isAwaitingFinalAckForCurrentSubscription = false;
```
yasemin-amzn: What do you think about introducing this variable as a more generic status indication boolean, i.e. pendingShutdown. 

Waiting to get the ACK for the last submitted batch is current design, that flips the pendingShutdown status. We can later wait to drain all events in the queue etc..

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
ashwing: Sure

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
ashwing: Ya. makes sense

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
ashwing: pendingShutdown might be misleading, as we are not letting know what is being shutdown. Do you think pendingCurrentSubscriptionShutdown or pendingActiveSubscriptionShutdown would be better?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
ashwing: `Do we expect queue to be full, by the time subscribeToShard call is made? ` - Ideally no, just for precaution. 
`We can either pull that out to its own method` - would be preferred instead of waiting
`isAwaitingFinalAck variable is tightly couple with the queue status, why not update that within the if?` - Updates as part of yatins comment

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
```diff
@@ -166,8 +170,8 @@ RecordFlow evictAckedEventAndScheduleNextEvent(RecordsDeliveryAck recordsDeliver
             // Update the triggering flow for post scheduling upstream request.
             flowToBeReturned = recordsRetrievedContext.getRecordFlow();
             // Try scheduling the next event in the queue, if available.
-            if (!isAwaitingFinalAckForCurrentSubscription && recordsDeliveryQueue.peek() != null) {
-                subscriber.onNext(recordsDeliveryQueue.peek().getRecordsRetrieved());
+            if (recordsDeliveryQueue.peek() != null) {
```
yatins47: check with isEmpty() ?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
```diff
@@ -212,6 +216,14 @@ void bufferCurrentEventAndScheduleIfRequired(RecordsRetrieved recordsRetrieved,
         }
     }
 
+    // This method is not thread-safe. You need to acquire a lock in the caller in order to execute this.
+    // Schedule the next event only when the active subscription is not pending shutdown.
+    private void scheduleNextEvent(RecordsRetrieved recordsRetrieved) {
```
yatins47: nit : what about checkSubscriptionStatusAndScheduleNextEvent ?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
```diff
@@ -320,27 +332,32 @@ private void errorOccurred(RecordFlow triggeringFlow, Throwable t) {
...
-            isAwaitingFinalAckForCurrentSubscription = true;
+            pendingActiveSubscriptionShutdown = true;
             try {
                 // Wait for the configured time to get a notification for already delivered event, if any.
                 lockObject.wait(TIME_TO_WAIT_FOR_FINAL_ACK_MILLIS);
             } catch (InterruptedException e) {
                 Thread.currentThread().interrupt();
             }
+            // Clear the queue to remove any remaining entries from the queue.
+            recordsDeliveryQueue.clear();
```
yatins47: Why cant we clear the queue before waiting ?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
ashwing: Hmmm we might add more checks in future and we can't keep extending the name. Hence just mentioning the core functionality.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
ashwing: If we clear before waiting, the ack received will be marked as stale and records will be duplicated.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
yasemin-amzn: sure, either one is good.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
```diff
@@ -313,8 +331,33 @@ private void errorOccurred(RecordFlow triggeringFlow, Throwable t) {
...
-    private void clearRecordsDeliveryQueue() {
-        recordsDeliveryQueue.clear();
+    // This method is not thread safe. This needs to be executed after acquiring lock on this.lockObject
+    private void resetRecordsDeliveryStateOnSubscriptionInit() {
+        // Clear any lingering records in the queue.
+        if (!recordsDeliveryQueue.isEmpty()) {
+            recordsDeliveryQueue.clear();
+        }
+        // Set pendingActiveSubscriptionShutdown to default value.
+        pendingActiveSubscriptionShutdown = false;
```
yasemin-amzn: Should we log warning if pendingActiveSubscriptionShutdown is found true, or there are records in the queue? 

Either of these cases points to something not working as expected, and logging status under such conditions can help us for RC'ing issues.

ashwing: yep, they will be informational.

PR #603, opened 2019-08-28
Fix to prevent data loss and stuck shards in the event of failed records delivery in Polling readers

ashwing:
Fix to prevent data loss and stuck shards in the event of failed records delivery in **`Polling`** readers

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
@@ -68,6 +68,12 @@ import static software.amazon.kinesis.common.DiagnosticUtils.takeDelayedDelivery
  * i.e. the byte size of the records stored in the cache and maxRecordsCount i.e. the max number of records that should
  * be present in the cache across multiple GetRecordsResult object. If no data is available in the cache, the call from
  * the record processor is blocked till records are retrieved from Kinesis.
+ *
+ * There are three threads namely publisher, demand-notifier and ack-notifier which will contend to drain the events
+ * to the Subscriber (ShardConsumer in KCL). The publisher/demand-notifier thread gains the control to drain only when
+ * there is no pending event in the prefetch queue waiting for the ack. Otherwise, it will be the ack-notifier thread
+ * which will drain an event on the receipt of an ack.
+ *
  */
 @Slf4j
 @KinesisClientInternalApi
@@ -104,7 +110,7 @@ public class PrefetchRecordsPublisher implements RecordsPublisher {
     // This flag controls who should drain the next request in the prefetch queue.
     // When set to false, the publisher and demand-notifier thread would have the control.
     // When set to true, the event-notifier thread would have the control.
-    private AtomicBoolean shouldDrainEventOnAck = new AtomicBoolean(false);
+    private AtomicBoolean shouldDrainEventOnlyOnAck = new AtomicBoolean(false);
 
     /**
      * Constructor for the PrefetchRecordsPublisher. This cache prefetches records from Kinesis and stores them in a
@@ -179,15 +185,16 @@ public class PrefetchRecordsPublisher implements RecordsPublisher {
         return result == null ? result : result.prepareForPublish();
     }
 
-    RecordsRetrieved evictNextResult() {
+    RecordsRetrieved pollNextResultAndUpdatePrefetchCounters() {
         throwOnIllegalState();
         final PrefetchRecordsRetrieved result = getRecordsResultQueue.poll();
         if (result != null) {
             prefetchCounters.removed(result.processRecordsInput);
             requestedResponses.decrementAndGet();
         } else {
-            log.info("{}: No record batch found while evicting from the prefetch queue. This indicates the prefetch buffer"
-                    + "was reset.", shardId);
+            log.info(
+                    "{}: No record batch found while evicting from the prefetch queue. This indicates the prefetch buffer"
+                            + "was reset.", shardId);
         }
         return result;
     }
@@ -213,7 +220,7 @@ public class PrefetchRecordsPublisher implements RecordsPublisher {
             // Give the drain control to publisher/demand-notifier thread.
             log.debug("{} : Publisher thread takes over the draining control. Queue Size : {}, Demand : {}", shardId,
                     getRecordsResultQueue.size(), requestedResponses.get());
-            shouldDrainEventOnAck.set(false);
+            shouldDrainEventOnlyOnAck.set(false);
 
             prefetchCounters.reset();
 
@@ -233,7 +240,7 @@ public class PrefetchRecordsPublisher implements RecordsPublisher {
             @Override
             public void request(long n) {
                 requestedResponses.addAndGet(n);
-                initiateDrainQueueForRequests();
+                drainQueueForRequestsIfAllowed();
             }
 
             @Override
@@ -249,12 +256,12 @@ public class PrefetchRecordsPublisher implements RecordsPublisher {
         // Verify if the ack matches the head of the queue and evict it.
         if (recordsToCheck != null && recordsToCheck.batchUniqueIdentifier()
                 .equals(recordsDeliveryAck.batchUniqueIdentifier())) {
-            evictNextResult();
+            pollNextResultAndUpdatePrefetchCounters();
             // Upon evicting, check if queue is empty. if yes, then give the drain control back to publisher thread.
-            if(getRecordsResultQueue.isEmpty()) {
-                log.debug("{} : Publisher thread takes over the draining control. Queue Size : {}, Demand : {}", shardId,
-                        getRecordsResultQueue.size(), requestedResponses.get());
-                shouldDrainEventOnAck.set(false);
+            if (getRecordsResultQueue.isEmpty()) {
+                log.debug("{} : Publisher thread takes over the draining control. Queue Size : {}, Demand : {}",
+                        shardId, getRecordsResultQueue.size(), requestedResponses.get());
+                shouldDrainEventOnlyOnAck.set(false);
             } else {
                 // Else attempt to drain the queue.
                 drainQueueForRequests();
@@ -263,9 +270,10 @@ public class PrefetchRecordsPublisher implements RecordsPublisher {
             // Log and ignore any other ack received. As long as an ack is received for head of the queue
             // we are good. Any stale or future ack received can be ignored, though the latter is not feasible
             // to happen.
-            final BatchUniqueIdentifier peekedBatchUniqueIdentifier = recordsToCheck == null ? null : recordsToCheck.batchUniqueIdentifier();
-            log.info("{} :  Received a stale notification with id {} instead of expected id {} at {}. Will ignore.", shardId,
-                    recordsDeliveryAck.batchUniqueIdentifier(), peekedBatchUniqueIdentifier, Instant.now());
+            final BatchUniqueIdentifier peekedBatchUniqueIdentifier =
+                    recordsToCheck == null ? null : recordsToCheck.batchUniqueIdentifier();
+            log.info("{} :  Received a stale notification with id {} instead of expected id {} at {}. Will ignore.",
+                    shardId, recordsDeliveryAck.batchUniqueIdentifier(), peekedBatchUniqueIdentifier, Instant.now());
         }
         // Take action based on the time spent by the event in queue.
         takeDelayedDeliveryActionIfRequired(shardId, lastEventDeliveryTime, log);
@@ -294,8 +302,8 @@ public class PrefetchRecordsPublisher implements RecordsPublisher {
      * Method that will be called by the 'publisher thread' and the 'demand notifying thread',
      * to drain the events if the 'event notifying thread' do not have the control.
      */
-    private synchronized void initiateDrainQueueForRequests() {
-        if(!shouldDrainEventOnAck.get()) {
+    private synchronized void drainQueueForRequestsIfAllowed() {
+        if (!shouldDrainEventOnlyOnAck.get()) {
             drainQueueForRequests();
         }
     }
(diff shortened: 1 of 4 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
```diff
@@ -248,11 +290,39 @@ private void addArrivedRecordsInput(PrefetchRecordsRetrieved recordsRetrieved) t
         prefetchCounters.added(recordsRetrieved.processRecordsInput);
     }
 
+    /**
+     * Method that will be called by the 'publisher thread' and the 'demand notifying thread',
+     * to drain the events if the 'event notifying thread' do not have the control.
+     */
+    private synchronized void initiateDrainQueueForRequests() {
```
isurues: initiate may not be the best word here. maybe drainQueueForRequestsIfAllowed?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
```diff
@@ -248,11 +290,39 @@ private void addArrivedRecordsInput(PrefetchRecordsRetrieved recordsRetrieved) t
...
-            eventDeliveryLock.acquireUninterruptibly();
-            eventDeliveryLockAcquireTime = Instant.now();
-            subscriber.onNext(getNextResult());
+        final RecordsRetrieved recordsToDeliver = peekNextResult();
+        // If there is an event available to drain and if there is at least one demand,
+        // then schedule it for delivery
+        if (requestedResponses.get() > 0 && recordsToDeliver != null) {
+            lastEventDeliveryTime = Instant.now();
+            subscriber.onNext(recordsToDeliver);
+            if(!shouldDrainEventOnAck.get()) {
```
isurues: nit: missing space between if and (

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
```diff
@@ -248,11 +290,39 @@ private void addArrivedRecordsInput(PrefetchRecordsRetrieved recordsRetrieved) t
...
+                log.debug("{} : Notifier thread takes over the draining control. Queue Size : {}, Demand : {}", shardId,
+                        getRecordsResultQueue.size(), requestedResponses.get());
+                shouldDrainEventOnAck.set(true);
+            }
+        } else {
+            // Since we haven't scheduled the event delivery, give the drain control back to publisher/demand-notifier
+            // thread.
+            if(shouldDrainEventOnAck.get()){
```
isurues: nit: missing space between if and (. check all such places.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
```diff
@@ -195,6 +209,12 @@ public void restartFrom(RecordsRetrieved recordsRetrieved) {
         resetLock.writeLock().lock();
         try {
             getRecordsResultQueue.clear();
+
+            // Give the drain control to publisher/demand-notifier thread.
+            log.debug("{} : Publisher thread takes over the draining control. Queue Size : {}, Demand : {}", shardId,
+                    getRecordsResultQueue.size(), requestedResponses.get());
+            shouldDrainEventOnAck.set(false);
```
isurues: Before this change, we had the eventDeliveryLock semaphore which was only released upon receiving the ack. So if the ack is lost somewhere, the shard gets stuck for ever. Was there any reason why we didn't release it within restartFrom here?

Looks like we are fixing that issue in this change. Shouldn't adding this step into the Semaphore solution fix the issue? What do we gain on top of that by this new solution?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
```diff
@@ -224,12 +244,34 @@ public void cancel() {
...
+                .equals(recordsDeliveryAck.batchUniqueIdentifier())) {
+            evictNextResult();
+            // Upon evicting, check if queue is empty. if yes, then give the drain control back to publisher thread.
+            if(getRecordsResultQueue.isEmpty()) {
+                log.debug("{} : Publisher thread takes over the draining control. Queue Size : {}, Demand : {}", shardId,
+                        getRecordsResultQueue.size(), requestedResponses.get());
+                shouldDrainEventOnAck.set(false);
+            } else {
+                // Else attempt to drain the queue.
+                drainQueueForRequests();
```
isurues: In the happy case where we continue to receive acks for each event delivered, always the ack-notifier thread will be draining the queue. Publisher thread will continue to do only the fetch part. I think this conflicts a bit with the design of this class and the purpose of the publisher thread.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
```diff
@@ -224,12 +244,34 @@ public void cancel() {
...
+            }
+        } else {
+            // Log and ignore any other ack received. As long as an ack is received for head of the queue
+            // we are good. Any stale or future ack received can be ignored, though the latter is not feasible
+            // to happen.
```
isurues: Another reason to come here is when the queue is reset before the ack comes back. Add this too in the comment?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
```diff
@@ -97,8 +100,11 @@
     private final ReentrantReadWriteLock resetLock = new ReentrantReadWriteLock();
     private boolean wasReset = false;
 
-    private final Semaphore eventDeliveryLock = new Semaphore(1);
-    private Instant eventDeliveryLockAcquireTime = Instant.EPOCH;
+    private Instant lastEventDeliveryTime = Instant.EPOCH;
+    // This flag controls who should drain the next request in the prefetch queue.
+    // When set to false, the publisher and demand-notifier thread would have the control.
+    // When set to true, the event-notifier thread would have the control.
+    private AtomicBoolean shouldDrainEventOnAck = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
isurues: It looks like the reason for shards getting stuck in the previous semaphore solution was that we were not releasing the semaphore when the health checker calls restartFrom. If we do that, shouldn't it solve the entire problem of shards getting stuck forever?

This solution fixes the above problem by handing the control over to publisher thread when the restartFrom is called. But what do we gain by letting the ack-notifier thread drain the queue?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
```diff
@@ -151,30 +157,38 @@ public void start(ExtendedSequenceNumber extendedSequenceNumber, InitialPosition
...
-            result = getRecordsResultQueue.take().prepareForPublish();
+    }
+
+    RecordsRetrieved peekNextResult() {
+        throwOnIllegalState();
+        final PrefetchRecordsRetrieved result = getRecordsResultQueue.peek();
+        return result == null ? result : result.prepareForPublish();
+    }
+
+    RecordsRetrieved evictNextResult() {
```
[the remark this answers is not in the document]
ashwing: Hmmm no strong preference. Will change

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
```diff
@@ -97,8 +100,11 @@
     private final ReentrantReadWriteLock resetLock = new ReentrantReadWriteLock();
     private boolean wasReset = false;
 
-    private final Semaphore eventDeliveryLock = new Semaphore(1);
-    private Instant eventDeliveryLockAcquireTime = Instant.EPOCH;
+    private Instant lastEventDeliveryTime = Instant.EPOCH;
+    // This flag controls who should drain the next request in the prefetch queue.
+    // When set to false, the publisher and demand-notifier thread would have the control.
+    // When set to true, the event-notifier thread would have the control.
```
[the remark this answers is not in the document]
ashwing: Okay

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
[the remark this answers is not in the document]
ashwing: 1. The blocking publisher thread on the semaphore would have acquired the readlock. In order to release the semaphore, the reset thread should acquire readlock as well. Assume the reset thread acquires the readlock and releases the semaphore, we should avoid the publisher thread from publisher thread from further publishing, until the reset thread acquires the write lock.
2. Also if we have a stale notification after clearing the queue in reset say after 60 seconds, then this would go ahead and release a semaphore, which means the publisher can publish two events without receiving ack for first event, which would cause durability issues.

Considering these factors among others, we decided it would be better if we can validate the ack before publishing the next event, it would prevent the durability issue. Also to ensure that an event is scheduled as soon as possible when there are prefetched events and to avoid having any blocking calls, we introduced this concept of scheduling from different threads and exchanging controls based on the queue state and demand.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
ashwing: Yep

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
ashwing: will address

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
ashwing: answered above

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
ashwing: Publisher thread will be a kick starter in the event of paused deliveries. In happy case, yes ack-notifier will be scheduling for minimum delay between each delivery and for simplicity.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
ashwing: That would be considered as a stale ack.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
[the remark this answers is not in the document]
isurues: Aha.. yes. Makes sense. Thanks for the details.

But I still think we can achieve this without letting the ack thread drain the queue. We can discuss this later. :)

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
isurues: Yep, sorry about the duplicate. :)

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
isurues: I agree that it'll help a bit with the delay. But not with the simplicity. :) We have 3 different threads draining the queue. I think it would be better and simpler if we had only one thread publishing and other two just updating the demand and registering the ack. Anyway that's a bigger change and we can consider that in the future. :)

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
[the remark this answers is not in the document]
ashwing: The publisher thread has this logic of spinning in a while loop to offer the prefetched element to the queue. That's the reason why the ack-thread (now) and demand-thread (earlier) are calling the drainQueue directly. Let's discuss this offline.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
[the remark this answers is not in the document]
ashwing: Updated to pollNextResultAndUpdatePrefetchCounters

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
ashwing: Addressed this in https://github.com/awslabs/amazon-kinesis-client/pull/606/files

PR #621, opened 2019-10-07
handling completed and blocked tasks better during graceful shutdown

aggarwal:
*Description of changes:*
Graceful shutdown gets blocked by shard consumers that are either already in shutdown state or are blocked on parent tasks. This change avoids over-counting consumers in the graceful shutdown latch, and allows blocked tasks to transition to shutdown state when shutdown is requested.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
@@ -905,6 +905,13 @@ public class Worker implements Runnable {
                 ShardConsumer consumer = shardInfoShardConsumerMap.get(shardInfo);
 
                 if (consumer == null || ConsumerStates.ShardConsumerState.SHUTDOWN_COMPLETE.equals(consumer.getCurrentState())) {
+                    //
+                    // CASE1: There is a race condition between retrieving the current assignments, and creating the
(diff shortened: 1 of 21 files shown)


aggarwal:
Added some unit tests.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -903,16 +903,12 @@ public boolean hasGracefulShutdownStarted() {
...
-                if (consumer != null) {
-                    consumer.notifyShutdownRequested(shutdownNotification);
-                } else {
-                    //
-                    // There is a race condition between retrieving the current assignments, and creating the
-                    // notification. If the a lease is lost in between these two points, we explicitly decrement the
-                    // notification latches to clear the shutdown.
-                    //
+
+                if (consumer == null || ConsumerStates.ShardConsumerState.SHUTDOWN_COMPLETE.equals(consumer.getCurrentState())) {
```
ashwing: Can we preserve the existing comment and a new comment for the shutdown_complete case?

aggarwal: Sure, added the comment back. Also added another comment explaining the shutdown case.

aggarwal:
Deleting since upstream source has changed in a way that conflicts with this PR. Will submit a new PR with the conflict resolved.

PR #624, opened 2019-10-14
Shard end v2

ychunxue:
*Description of changes:*

Change is made in ShutdownTask. When reaching a ShutdownTask with Shard_End shutdown reason, do a re-validation about if the current shard reached shard_end. If it didn't reach the shard_end, force the current ShardConsumer to lose the lease and shutdown with Lease_Lost state. This would enable other workers to pick up the lease for the unclosed shard.

Test did for the changes:

    Unit tests covered ShardConsumer shutdown in three different scenarios: Lease lost, ShardEnd and incorrect ShardEnd.

    Functional test done in real ShardEnd and incorrect ShardEnd scenarios.

    Long running regression test is done with Split and Merge accounts. And also having the incorrect ShardEnd triggered randomly during the test to catch regressions.

### DIFF
--- a/amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/HierarchicalShardSyncerTest.java
+++ b/amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/HierarchicalShardSyncerTest.java
@@ -171,6 +171,9 @@ public class HierarchicalShardSyncerTest {
         testCheckAndCreateLeasesForShardsIfMissing(INITIAL_POSITION_LATEST);
     }
 
+    /**
+     * Test checkAndCreateLeaseForNewShards while not providing a pre-fetched list of shards
+     */
     @Test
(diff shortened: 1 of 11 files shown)


amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/HierarchicalShardSyncerTest.java
```diff
@@ -205,6 +205,66 @@ public void testCheckAndCreateLeasesForShardsIfMissingAtLatest() throws Exceptio
 
     }
 
+    @Test
```
Cory-Bradshaw: Comment for file in general: Can we add comments explaining what each test is doing, and explicitly calling out any changes with code comments to common setup between very similar tests.

ychunxue: Yep, addressing

PR #638, opened 2019-10-25
Updating sdk version and fixing unit tests

ychunxue:
*Description of changes:*
Upgrade SDK version to 2.9.25. Fixed several unit tests.

### DIFF
--- a/amazon-kinesis-client/src/test/java/software/amazon/kinesis/retrieval/fanout/FanOutConsumerRegistrationTest.java
+++ b/amazon-kinesis-client/src/test/java/software/amazon/kinesis/retrieval/fanout/FanOutConsumerRegistrationTest.java
@@ -90,8 +90,8 @@ public class FanOutConsumerRegistrationTest {
 
         assertThat(consumerArn, equalTo(CONSUMER_ARN));
 
-        verify(client).describeStreamConsumer(eq(createDescribeStreamConsumerRequest(null)));
-        verify(client).describeStreamSummary(eq(createDescribeStreamSummaryRequest()));
+        verify(client, times(2)).describeStreamConsumer(any(DescribeStreamConsumerRequest.class));
+        verify(client).describeStreamSummary(any(DescribeStreamSummaryRequest.class));
         verify(client, never()).registerStreamConsumer(any(RegisterStreamConsumerRequest.class));
     }
 
@@ -112,8 +112,8 @@ public class FanOutConsumerRegistrationTest {
         assertThat(firstCall, equalTo(CONSUMER_ARN));
         assertThat(secondCall, equalTo(CONSUMER_ARN));
 
-        verify(client).describeStreamConsumer(eq(createDescribeStreamConsumerRequest(null)));
-        verify(client).describeStreamSummary(eq(createDescribeStreamSummaryRequest()));
+        verify(client, times(2)).describeStreamConsumer(any(DescribeStreamConsumerRequest.class));
+        verify(client).describeStreamSummary(any(DescribeStreamSummaryRequest.class));
         verify(client, never()).registerStreamConsumer(any(RegisterStreamConsumerRequest.class));
     }
 
@@ -131,9 +131,9 @@ public class FanOutConsumerRegistrationTest {
         try {
             consumerRegistration.getOrCreateStreamConsumerArn();
         } finally {
-            verify(client).describeStreamSummary(eq(createDescribeStreamSummaryRequest()));
+            verify(client).describeStreamSummary(any(DescribeStreamSummaryRequest.class));
             verify(client, times(MAX_DSC_RETRIES))
-                    .describeStreamConsumer(eq(createDescribeStreamConsumerRequest(null)));
+                    .describeStreamConsumer(any(DescribeStreamConsumerRequest.class));
         }
     }
 
@@ -156,9 +156,9 @@ public class FanOutConsumerRegistrationTest {
             consumerRegistration.getOrCreateStreamConsumerArn();
         } finally {
             verify(client, times(RSC_RETRIES))
-                    .registerStreamConsumer(eq(createRegisterStreamConsumerRequest()));
+                    .registerStreamConsumer(any(RegisterStreamConsumerRequest.class));
             // Verify that DescribeStreamConsumer was called for at least RegisterStreamConsumer retries + 1 at start.
-            verify(client).describeStreamConsumer(eq(createDescribeStreamConsumerRequest(null)));
+            verify(client).describeStreamConsumer(any(DescribeStreamConsumerRequest.class));
         }
     }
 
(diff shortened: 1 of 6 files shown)


amazon-kinesis-client/src/test/java/software/amazon/kinesis/retrieval/fanout/FanOutConsumerRegistrationTest.java
```diff
@@ -90,8 +90,8 @@ public void testConsumerAlreadyExists() throws Exception {
 
         assertThat(consumerArn, equalTo(CONSUMER_ARN));
 
-        verify(client).describeStreamConsumer(eq(createDescribeStreamConsumerRequest(null)));
-        verify(client).describeStreamSummary(eq(createDescribeStreamSummaryRequest()));
+        verify(client, times(2)).describeStreamConsumer(any(DescribeStreamConsumerRequest.class));
+        verify(client).describeStreamSummary(any(DescribeStreamSummaryRequest.class));
```
Cory-Bradshaw: This falls under "Test behavior, not implementation" 

Does it really matter that we are calling describeStreamConsumer, or is verifying that we never register the already existing consumer again really the behavior we are looking to confirm?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/retrieval/fanout/FanOutConsumerRegistrationTest.java
```diff
@@ -112,8 +112,8 @@ public void testConsumerAlreadyExistsMultipleCalls() throws Exception {
         assertThat(firstCall, equalTo(CONSUMER_ARN));
         assertThat(secondCall, equalTo(CONSUMER_ARN));
 
-        verify(client).describeStreamConsumer(eq(createDescribeStreamConsumerRequest(null)));
-        verify(client).describeStreamSummary(eq(createDescribeStreamSummaryRequest()));
+        verify(client, times(2)).describeStreamConsumer(any(DescribeStreamConsumerRequest.class));
+        verify(client).describeStreamSummary(any(DescribeStreamSummaryRequest.class));
```
Cory-Bradshaw: Same as above, This test is trying harder to recreate an already existing consumer by calling multiple times. Do we really care that we are verifying the expected increase in the internal API call (.describeStreamConsumer), or should we just care that we do not register that already existing consumer?

Note: If this API has strict TPS limits (Like DescribeStream), then perhaps we really do care anytime this API is invoked.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/retrieval/fanout/FanOutConsumerRegistrationTest.java
```diff
@@ -156,9 +156,9 @@ public void testRegisterStreamConsumerThrottled() throws Exception {
             consumerRegistration.getOrCreateStreamConsumerArn();
         } finally {
             verify(client, times(RSC_RETRIES))
-                    .registerStreamConsumer(eq(createRegisterStreamConsumerRequest()));
+                    .registerStreamConsumer(any(RegisterStreamConsumerRequest.class));
             // Verify that DescribeStreamConsumer was called for at least RegisterStreamConsumer retries + 1 at start.
-            verify(client).describeStreamConsumer(eq(createDescribeStreamConsumerRequest(null)));
+            verify(client).describeStreamConsumer(any(DescribeStreamConsumerRequest.class));
```
Cory-Bradshaw: "Verify that DescribeStreamConsumer was called for at least RegisterStreamConsumer retries + 1 at start."

This is not what we are verifying. This verifies that we are calling it exactly once, I'm not saying this is guaranteed to be an implementation detail that we are testing, but take a cursory glance to see if this is a worthwhile verification.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/retrieval/fanout/FanOutConsumerRegistrationTest.java
```diff
@@ -188,11 +188,10 @@ public void testNewRegisterStreamConsumer() throws Exception {
...
 
-        verify(client).registerStreamConsumer(eq(createRegisterStreamConsumerRequest()));
-        verify(client).describeStreamSummary(eq(createDescribeStreamSummaryRequest()));
-        verify(client).describeStreamConsumer(eq(createDescribeStreamConsumerRequest(null)));
-        verify(client, times(2))
-                .describeStreamConsumer(eq(createDescribeStreamConsumerRequest(CONSUMER_ARN)));
+        verify(client).registerStreamConsumer(any(RegisterStreamConsumerRequest.class));
+        verify(client).describeStreamSummary(any(DescribeStreamSummaryRequest.class));
+        verify(client, times(3))
+                .describeStreamConsumer(any(DescribeStreamConsumerRequest.class));
```
Cory-Bradshaw: Same note on this test as above tests, but even more-so now that we have to broaden the search which is picking up an additional request that we didn't care about before. Do we really care if we are calling this API here, especially when considering that we previously ignored one of these invocations?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisherTest.java
```diff
@@ -1086,7 +1086,7 @@ public void testContinuesAfterSequence() {
                         .type(ShardIteratorType.AT_SEQUENCE_NUMBER).build())
                 .build();
 
-        verify(kinesisClient).subscribeToShard(eq(expected), flowCaptor.capture());
+        verify(kinesisClient).subscribeToShard(any(SubscribeToShardRequest.class), flowCaptor.capture());
```
Cory-Bradshaw: This makes the verification/test pretty useless since we cannot tell where we are trying to start at. If we cannot verify the position here, can we verify elsewhere which sequence number we are attempting to start at?

See final comment for suggestion for new argument matcher to confirm only the fields we care about equal, we don't really care about object equality in any of these tests. (this can be applied to most edits in this commit if we decide we continue to care about these verifications)

amazon-kinesis-client/src/test/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisherTest.java
```diff
@@ -1112,7 +1112,7 @@ public void testContinuesAfterSequence() {
                         .type(ShardIteratorType.AFTER_SEQUENCE_NUMBER).build())
                 .build();
 
-        verify(kinesisClient).subscribeToShard(eq(nextExpected), nextFlowCaptor.capture());
+        verify(kinesisClient, times(2)).subscribeToShard(any(SubscribeToShardRequest.class), nextFlowCaptor.capture());
```
Cory-Bradshaw: Same as above, we at least need to verify the sequence number we are restarting at, can we use any other data point or object in our test to tell us this information?

Take a look here for how/when to use these in general: https://www.baeldung.com/mockito-argument-matchers 
The custom Argument matcher might be useful for checking on the parameters of the object that we care about.


````
public class SubscribeToShardRequestMatcher implements ArgumentMatcher<SubscribeToShardRequest> {
 
    private SubscribeToShardRequest left;
 
    // constructors
 
    @Override
    public boolean matches(SubscribeToShardRequest right) {
        return left.getShardId().equals(right.getShardId()) &&
          left. getStartingPosition().equals(right.getStartingPosition()) &&
          left. getConsumerARN().equals(right.getConsumerARN());
    }
}
````

And then the verification becomes:

````
verify(kinesisClient).subscribeToShard(argThat(new SubscribeToShardRequestMatcher(nextExpected)), nextFlowCaptor.capture());
````

amazon-kinesis-client/src/test/java/software/amazon/kinesis/retrieval/fanout/FanOutConsumerRegistrationTest.java
ychunxue: Yep, let's remove these describeStreamConsumer and describeStreamSummary verification

amazon-kinesis-client/src/test/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisherTest.java
ychunxue: Addressed.

Cory-Bradshaw:
Looks good to me. Thanks for taking the time to update the unit tests!

PR #642, opened 2019-11-01
Remove the shutdown event from the queue before executing the shudown…

ychunxue:
*Description of changes:*

Remove the shutdown event from the recordsDeliveryQueue to avoid non-empty queue in the new subscription.

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
@@ -168,7 +168,7 @@ public class FanOutRecordsPublisher implements RecordsPublisher {
             flowToBeReturned = recordsRetrievedContext.getRecordFlow();
             // Try scheduling the next event in the queue or execute the subscription shutdown action.
             if (!recordsDeliveryQueue.isEmpty()) {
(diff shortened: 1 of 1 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
```diff
@@ -225,9 +225,15 @@ RecordsRetrieved getRecordsRetrieved() {
...
-        void executeEventAction(Subscriber<? super RecordsRetrieved> subscriber) {
+        void executeEventAction(Subscriber<? super RecordsRetrieved> subscriber, BlockingQueue<RecordsRetrievedContext> recordsDeliveryQueue) {
             recordsOrShutdownEvent.apply(recordsEvent -> subscriber.onNext(recordsEvent),
-                    shutdownEvent -> shutdownEvent.getSubscriptionShutdownAction().run());
+                    shutdownEvent -> handleShutdownEvent(shutdownEvent, recordsDeliveryQueue));
+        }
+
+        private void handleShutdownEvent(SubscriptionShutdownEvent shutdownEvent, BlockingQueue<RecordsRetrievedContext> recordsDeliveryQueue) {
+            //Removing the Shutdown Event from the queue before executing the Shutdown action.
+            recordsDeliveryQueue.poll();
```
ashwing: Let's add this to the ShutdownEvent runnable, rather than having a new method here.

Issue #475, opened 2018-12-12
Unnecessary wait during scheduler initialization due to parentShardPollIntervalMillis sleep


bbeaudreault:
According to the docs, `parentShardPollIntervalMillis` has the following effect:

```
/**
     * Interval in milliseconds between polling to check for parent shard completion.
     * Polling frequently will take up more DynamoDB IOPS (when there are leases for shards waiting on
     * completion of parent shards).
     *
     * <p>Default value: 10000L</p>
     */
```

But currently this property is being used in two ways:

1) [At Scheduler initialization time, to back off on initialization attempts](https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java#L271)

2) [During BlockedOnParent state](https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ConsumerStates.java#L139)

The use-case of number 2 makes sense -- after a shard has split, it may take time for the parent to be drained. 

However use-case number 1 seems a little broken. It seems to be being used as a backoff timeout incase of retries. But the cost is being paid even if the first attempt was successful. We should either break out of the loop early, or use a different interval config if there is some other reason to wait here.

I was able to speed up my consumer start time by almost 10s by reducing parentShardPollIntervalMillis to 500ms. This means that in a sharding scenario or failure case I will be hammering dynamodb a bit. I've scaled my dynamodb to account for this, but I shouldn't have to. Instead the timeout should either be skipped after a successful attempt, or the timeout should get its own config variable.

rehevkor5:
This also occurs during initialization even if the problem is due to an IAM permissions issue. Ideally, if it is a non-transient issue like an IAM issue, it would fail eagerly instead of retrying for multiple 10s periods.

etspaceman:
This can be closed thanks to https://github.com/awslabs/amazon-kinesis-client/commit/3e32ff190687e7a34274923d36f640d337298c35

bbeaudreault:
Thanks!

PR #665, opened 2019-12-13
Fix for #475

etspaceman:
*Issue #, if available:*

#475 

*Description of changes:*

- Add a new configuration, `schedulerInitializationRetryIntervalMillis`, which is used in place of `parentShardPollIntervalMillis` when retrying failures during initialization. The default for this is 1 second.
- Updated the Scheduler initialization so that the sleep only occurs when initialization has not completed.

### DIFF
--- a/amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/coordinator/KinesisClientLibConfiguration.java
+++ b/amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/coordinator/KinesisClientLibConfiguration.java
@@ -202,6 +202,11 @@ public class KinesisClientLibConfiguration {
      */
     public static final int DEFAULT_MAX_LIST_SHARDS_RETRY_ATTEMPTS = 50;
 
+    /**
+     * The sleep time between initialization attempts for the scheduler.
+     */
+    public static final long DEFAULT_SCHEDULER_INITIALIZATION_BACKOFF_TIME_MILLIS = 1000;
+
     private String applicationName;
     private String tableName;
     private String streamName;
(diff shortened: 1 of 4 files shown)


etspaceman:
The build errors on this seem unrelated and are experienced on master as well.

amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/coordinator/KinesisClientLibConfiguration.java
```diff
@@ -202,6 +202,11 @@
      */
     public static final int DEFAULT_MAX_LIST_SHARDS_RETRY_ATTEMPTS = 50;
 
+    /**
+     * The number of times the Proxy will retry listShards call when throttled.
+     */
+    public static final long DEFAULT_SCHEDULER_INITIALIZATION_RETRY_INTERVAL_MILLIS = 1000;
```
etspaceman: I'm honestly not sure if this file needs changes. It doesn't seem to actually be used by anything but I made the changes nontheless.

etspaceman:
@yasemin-amzn comments addressed

ashwing:
Do we know why the code builds are failing?

etspaceman:
I am not sure @ashwing but these same errors were present on Master when I was testing.

etspaceman:
@yasemin-amzn are there any other changes that would be required for this?

Cory-Bradshaw:
Retrying builds succeeding for OpenJDK: openjdk8-build-kcl:cdef2578-66d2-4ab4-982f-a91ec4997b8e

Cory-Bradshaw:
Retried the Travis CI, both versions of java is building as expected.

PR #691, opened 2020-02-20
Introducing multistream support in KCL 2.X

ashwing:
*Issue #, if available:*

*Description of changes:*
Introducing multistream support in KCL 2.X

*Known Pending Tasks:*
Periodic Lease refresh logic for new streams identified through Multistreamtracker
Metric changes to add stream as dimension
Log changes to include streamName
ConsumerArn to be added in StreamConfig


amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/ConfigsBuilder.java
```diff
@@ -108,6 +109,8 @@ public String namespace() {
         return namespace;
     }
 
+    private MultiStreamTracker multiStreamTracker;
```
satselva: Can we make this final and immutable ?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamConfig.java
```diff
@@ -0,0 +1,14 @@
+package software.amazon.kinesis.common;
+
+import lombok.AccessLevel;
+import lombok.Data;
+import lombok.experimental.Accessors;
+import lombok.experimental.FieldDefaults;
+
+@Data
```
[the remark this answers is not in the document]
ashwing: Ya, will update

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
```diff
@@ -240,39 +262,100 @@ public LeaseManagementConfig metricsFactory(final MetricsFactory metricsFactory)
...
         }
         return leaseManagementFactory;
     }
 
+    public LeaseManagementConfig leaseManagementFactory(final LeaseManagementFactory leaseManagementFactory) {
+        this.leaseManagementFactory = leaseManagementFactory;
+        return this;
+    }
+
+//    private InitialPositionInStreamExtended getInitialPositionExtendedForStream(String streamName) {
```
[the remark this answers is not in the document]
ashwing: Yes

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/MultiStreamLease.java
```diff
@@ -0,0 +1,91 @@
...
+            return false;
+        }
+        MultiStreamLease other = (MultiStreamLease) obj;
+        if (streamName == null) {
+            if (other.streamName != null) {
+                return false;
+            }
+        } else if (!streamName.equals(other.streamName)) {
+            return false;
+        }
```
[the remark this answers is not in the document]
ashwing: This was from legacy code, will change

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/ShardInfo.java
```diff
@@ -118,8 +129,20 @@ public boolean equals(Object obj) {
...
+                .append(streamName.orElse(""), other.streamName.orElse("")).isEquals();
 
     }
 
+    /**
+     *
+     * @param shardInfo
+     * @return
+     */
+    public static String getLeaseKey(ShardInfo shardInfo) {
```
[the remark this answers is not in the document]
ashwing: This is a static method which acts as helper to derive key from ShardInfo. Let me know if there is a strong reason to move to other util.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseManagementFactory.java
```diff
@@ -321,13 +329,83 @@ public DynamoDBLeaseManagementFactory(final KinesisAsyncClient kinesisClient, fi
...
+     */
+    public DynamoDBLeaseManagementFactory(final KinesisAsyncClient kinesisClient, final StreamConfig streamConfig,
+            final DynamoDbAsyncClient dynamoDBClient, final String tableName, final String workerIdentifier,
+            final ExecutorService executorService, final long failoverTimeMillis, final long epsilonMillis,
+            final int maxLeasesForWorker, final int maxLeasesToStealAtOneTime, final int maxLeaseRenewalThreads,
+            final boolean cleanupLeasesUponShardCompletion, final boolean ignoreUnexpectedChildShards,
+            final long shardSyncIntervalMillis, final boolean consistentReads, final long listShardsBackoffTimeMillis,
+            final int maxListShardsRetryAttempts, final int maxCacheMissesBeforeReload,
+            final long listShardsCacheAllowedAgeInSeconds, final int cacheMissWarningModulus,
+            final long initialLeaseTableReadCapacity, final long initialLeaseTableWriteCapacity,
```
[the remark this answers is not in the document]
ashwing: Will check

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/MultiStreamLease.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Override
+    public void update(Lease other) {
+        MultiStreamLease casted = validateAndCast(other);
+        super.update(casted);
+        streamName(casted.streamName);
+        shardId(casted.shardId);
+    }
+
+    public static String getLeaseKey(String streamName, String shardId) {
```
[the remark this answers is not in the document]
ashwing: This is from legacy code. Will update this, if we always deduce the key from the lease instance

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/MultiStreamLease.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.util.Objects;
+
+import static com.google.common.base.Verify.verifyNotNull;
+
+@Setter
+@NoArgsConstructor
+@Getter
+@Accessors(fluent = true)
+public class MultiStreamLease extends Lease {
```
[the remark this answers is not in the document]
ashwing: I believe any lease class we introduce will always be a subset of the subsequent lease. For instance Lease will be a subset of MultiStreamLease, MultiStreamLease will be a subset of MultiStreamMultiAccountLease, MultiStreamMultiAccountLease will be a subset of MultiStreamMultiAccountTwoPhaseCommitLease. The problem will be when there is a need for MultiStream + twoPhase commit only and without multiaccount support, or when there is a need for say basic lease + two phase commit. We can use strategy pattern, if we plan to introduce 2-phase commit as an independent feature of multistream/multiaccount lease or if we decide that leases can be combinational rather than incremental.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/ConfigsBuilder.java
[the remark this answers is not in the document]
yasemin-amzn: Why being optional is a blocker for being a final/immutable? 
Let's model as Optional<MultistreamTracker>, and set to absent for cases where it's not used.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -170,9 +181,26 @@ protected Scheduler(@NonNull final CheckpointConfig checkpointConfig,
         this.retrievalConfig = retrievalConfig;
 
         this.applicationName = this.coordinatorConfig.applicationName();
+        final MultiStreamTracker multiStreamTracker = this.retrievalConfig.multiStreamTracker();
+        if(multiStreamTracker == null) {
+            final StreamConfig streamConfig = new StreamConfig(this.retrievalConfig.streamName(),
+                    this.retrievalConfig.initialPositionInStreamExtended());
+            this.appStreamTracker = Either.right(streamConfig);
```
yasemin-amzn: There is an expectation that configuration is setup correctly, for multi-stream vs single-stream. let'e validate that with null check, both in here, and also in else case.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/MultiStreamLease.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @NonNull private String streamName;
+    @NonNull private String shardId;
+
+    public MultiStreamLease(Lease other) {
```
yasemin-amzn: What's the goal of this object? Seems like its same as lease object. Allowing both Lease and MultiStreamLease to be used within the same app will be confusing. If the app initialized to use multi-stream-leases, then it should be initializing multi-stream leases, not initializing leases and casting to multi-stream lease. 

If the only difference between the two is serialization, we should see if the multi-stream-lease abstraction should be added at the lease level, or at the higher serialization level according to the application configuration.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/processor/MultiStreamTracker.java
```diff
@@ -0,0 +1,19 @@
...
+ * to consume data from multiple streams.
+ */
+public interface MultiStreamTracker {
+
+    /**
+     * Returns the map of streams and its associated stream specific config.
+     *
+     * @return List of stream names
+     */
+    Map<String, StreamConfig> streamConfigMap();
```
yasemin-amzn: This iface even be used for singleStream setup. anything blocking us using this iface as the default for all applications? we would be just calling this just StreamTracker in that case.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/processor/ShardRecordProcessorFactory.java
```diff
@@ -25,4 +25,13 @@
      * @return
      */
     ShardRecordProcessor shardRecordProcessor();
+
+    /**
+     * Returns a new instance of the ShardRecordProcessor for a stream
+     * @param streamName
+     * @return ShardRecordProcessor
+     */
+    default ShardRecordProcessor shardRecordProcessor(String streamName) {
```
yasemin-amzn: stream name is ignored?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutConfig.java
```diff
@@ -82,26 +82,26 @@
 
     @Override
     public RetrievalFactory retrievalFactory() {
-        return new FanOutRetrievalFactory(kinesisClient, getOrCreateConsumerArn());
+        return new FanOutRetrievalFactory(kinesisClient, streamName, this::getOrCreateConsumerArn);
     }
 
-    private String getOrCreateConsumerArn() {
+    private String getOrCreateConsumerArn(String streamName) {
```
yasemin-amzn: is stream name sufficient here? stream creation timestamp sounds like the other critical piece of information for consumer creation.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/SynchronousBlockingRetrievalFactory.java
```diff
@@ -63,7 +63,7 @@ public SynchronousBlockingRetrievalFactory(String streamName, KinesisAsyncClient
     public GetRecordsRetrievalStrategy createGetRecordsRetrievalStrategy(@NonNull final ShardInfo shardInfo,
             @NonNull final MetricsFactory metricsFactory) {
         return new SynchronousGetRecordsRetrievalStrategy(
-                new KinesisDataFetcher(kinesisClient, streamName, shardInfo.shardId(), maxRecords, metricsFactory, kinesisRequestTimeout));
+                new KinesisDataFetcher(kinesisClient, shardInfo.streamName().orElse(streamName), shardInfo.shardId(), maxRecords, metricsFactory, kinesisRequestTimeout));
```
yasemin-amzn: doesn't shardInfo always have streamName?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/ConfigsBuilder.java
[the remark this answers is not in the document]
ashwing: We use lombok builder for building this config with mandatory fields like streamName, client passed in the constructor and optional fields can be set through the builder. If we make tracker as immutable, then we have to initialize the tracker in constructor with Optional.empty(), which will prevent the overriding through the builder. I preferred to introduce any new config as optional, to avoid creating explicit constructors. Alternatively, we can explicitly add new constructors that either allows to set streamName or MultiStreamTracker.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
ashwing: The current idea is customer can choose to override their current stream & initialposition with streamtracker. We can make this explicit by allowing them to set either stream+initialposition or streamtracker and then add validations.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/MultiStreamLease.java
ashwing: We either use Lease or Multistream lease in an application. This is determined through the serializer we set while initializing the scheduler. MultiStreamLease supports serializing and deserializing new fields like streamName and shardId, which we dont support with current Lease. We can rename this to be LeaseV2, as we might add pendingCheckpointState as well to for two phase commit.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/processor/MultiStreamTracker.java
ashwing: The idea is customer who implement this interface will see change in their ddb schema as their lease object will be different now. 

In scheduler, we use this to differentiate between Multistream and single stream in order to initialize lease serializer, lease manager, shard syncer etc.

`private final Either<MultiStreamTracker, StreamConfig> appStreamTracker;`

Also in scheduler, we use the following to track both multistream and single stream eventually.

`private final Map<String, StreamConfig> currentStreamConfigMap;`

amazon-kinesis-client/src/main/java/software/amazon/kinesis/processor/ShardRecordProcessorFactory.java
ashwing: Our code now internally uses shardRecordProcessor(String streamName) by default and falls back to shardRecordProcessor() if customer does not set the multistreamtracker. If they use mutistreamtracker, they need to override this method in their concrete impl to provide a stream specific record processor.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/SynchronousBlockingRetrievalFactory.java
ashwing: The current version of ShardInfo doesn't have streamName, as we do not have this this in DDB lease record, from which this is constructed.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutConfig.java
ashwing: Hmmm, let's sync up on this offline.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamIdentifier.java
```diff
@@ -0,0 +1,69 @@
...
+@EqualsAndHashCode @Getter @Accessors(fluent = true)
+public class StreamIdentifier {
+    private final Optional<String> accountIdOptional;
+    private final String streamName;
+    private final Optional<Long> streamCreationEpochOptional;
+
+    private static final String DELIMITER = ":";
+    private static final Pattern PATTERN = Pattern.compile(".*" + ":" + ".*" + ":" + "[0-9]*");
+
+    private StreamIdentifier(Optional<String> accountIdOptional, String streamName,
```
satselva: It looks better if we have two constructor with all field as mandatory values

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/MultiStreamLease.java
```diff
@@ -0,0 +1,83 @@
...
+    public static String getLeaseKey(String streamIdentifier, String shardId) {
+        verifyNotNull(streamIdentifier, "streamIdentifier should not be null");
+        verifyNotNull(shardId, "shardId should not be null");
+        return streamIdentifier + ":" + shardId;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(super.hashCode(), streamIdentifier);
+    }
```
satselva: Dont we need shard id in the Equals and Hashcode comparison ?

ashwing: Actually the shardId info is present in the base class leaseKey. Will refactor as required

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamIdentifier.java
ashwing: Agree. Will change

ashwing:
Closing this as this was merged to https://github.com/ashwing/amazon-kinesis-client/tree/ltr_1

PR #694, opened 2020-02-26
Stop passing latest shards to the next ShardSyncTask in ShardSyncTask…

ychunxue:
*Issue #, if available:*

*Description of changes:*

Currently KCL 1.x is having issue picking up new shards after resharding. This is due to a bug in ShardSyncTaskManager that we are passing the latestShards to the next pending ShardSyncTask. Fixing the issue by always passing null to the next ShardSyncTask.


Cory-Bradshaw:
Looks good to me.

PR #712, opened 2020-04-23
Also consider TableStatus.UPDATING status

Cai41:
*Description of changes:*
Based on: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/WorkingWithTables.Basics.html#WorkingWithTables.Basics.UpdateTable

```
When you issue an UpdateTable request, the status of the table changes from AVAILABLE to UPDATING. The table remains fully available for use while it is UPDATING. When this process is completed, the table status changes from UPDATING to AVAILABLE.
```

If we are updating a table's IOPS or billing mode, the UPDATING status can remain few minutes to one hour. This prevents KCL thread from starting. However based on DDB documentation, it is still functioning during UPDATING process, so KCL thread should be able to continue in this situation.

`mvn clean install -DskipITs` passed.


Cory-Bradshaw:
looks good to me.

PR #763, opened 2020-12-08
Fixing bug in PrefetchRecordsPublisher which was causing retry storms…

yatins47:
… if initial requests fail due to no wait time between get calls.

*Issue #, if available:*

*Description of changes:* Fixing bug in PrefetchRecordsPublisher which was causing retry storms if initial requests fail due to no wait time between get calls.

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
@@ -86,6 +86,7 @@ public class PrefetchRecordsPublisher implements RecordsPublisher {
     private final MetricsFactory metricsFactory;
     private final long idleMillisBetweenCalls;
     private Instant lastSuccessfulCall;
+    private boolean isFirstGetCallTry = true;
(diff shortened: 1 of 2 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
```diff
@@ -489,7 +490,13 @@ private void callShutdownOnStrategy() {
...
+            if (lastSuccessfulCall == null && isFirstGetCallTry) {
+                isFirstGetCallTry = false;
+                return;
+            }
+            // Add a sleep if lastSuccessfulCall is still null but this is not the first try to avoid retry storm
+            if(lastSuccessfulCall == null) {
+                Thread.sleep(idleMillisBetweenCalls);
                 return;
             }
             long timeSinceLastCall = Duration.between(lastSuccessfulCall, Instant.now()).abs().toMillis();
```
joshua-kim, line 502: Is lastSuccessfulCall set correctly when we're getting throttled? If we make a successful call and only receiving throttling events onwards, I think we can get stuck in a throttling loop here.

joshua-kim (approved):
LGTM

PR #776, opened 2021-01-23
KCL 1.X Guardrails for ShardEnd corruption and preventing lease table interference in multi-app JVM

ashwing:
1. Updating HashRange directly while creating leases and including HashRange in serializing and deserializing to/from DDB
2. Fix for making LeaseCleanupManager non-singleton to avoid cross-table interference in multiple apps running in same JVM
3. Fixing updateMetaInfo method to not update other lease table fields as this might lead to stale info persistence
4. Preventing shard deletion in LeaseCleanupManager if a valid shard does not have child shards in lease table and in Kinesis Service
5. Adding childshards update support in updateMetaInfo
6. Making LeaseCleanupManager to call updateMetaInfo instead of update for childshard update in lease

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
@@ -96,9 +96,9 @@ public class Worker implements Runnable {
     // Default configs for periodic shard sync
     private static final int SHARD_SYNC_SLEEP_FOR_PERIODIC_SHARD_SYNC = 0;
     private static final int PERIODIC_SHARD_SYNC_MAX_WORKERS_DEFAULT = 1; //Default for KCL.
-    static final long LEASE_TABLE_CHECK_FREQUENCY_MILLIS = 3 * 1000L;
-    static final long MIN_WAIT_TIME_FOR_LEASE_TABLE_CHECK_MILLIS = 1 * 1000L;
-    static final long MAX_WAIT_TIME_FOR_LEASE_TABLE_CHECK_MILLIS = 30 * 1000L;
+    static long LEASE_TABLE_CHECK_FREQUENCY_MILLIS = 3 * 1000L;
+    static long MIN_WAIT_TIME_FOR_LEASE_TABLE_CHECK_MILLIS = 1 * 1000L;
+    static long MAX_WAIT_TIME_FOR_LEASE_TABLE_CHECK_MILLIS = 30 * 1000L;
 
     private static final WorkerStateChangeListener DEFAULT_WORKER_STATE_CHANGE_LISTENER = new NoOpWorkerStateChangeListener();
     private static final LeaseCleanupValidator DEFAULT_LEASE_CLEANUP_VALIDATOR = new KinesisLeaseCleanupValidator();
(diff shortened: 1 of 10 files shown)


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -96,9 +96,9 @@
     // Default configs for periodic shard sync
     private static final int SHARD_SYNC_SLEEP_FOR_PERIODIC_SHARD_SYNC = 0;
     private static final int PERIODIC_SHARD_SYNC_MAX_WORKERS_DEFAULT = 1; //Default for KCL.
-    static final long LEASE_TABLE_CHECK_FREQUENCY_MILLIS = 3 * 1000L;
-    static final long MIN_WAIT_TIME_FOR_LEASE_TABLE_CHECK_MILLIS = 1 * 1000L;
-    static final long MAX_WAIT_TIME_FOR_LEASE_TABLE_CHECK_MILLIS = 30 * 1000L;
+    static long LEASE_TABLE_CHECK_FREQUENCY_MILLIS = 3 * 1000L;
+    static long MIN_WAIT_TIME_FOR_LEASE_TABLE_CHECK_MILLIS = 1 * 1000L;
+    static long MAX_WAIT_TIME_FOR_LEASE_TABLE_CHECK_MILLIS = 30 * 1000L;
```
joshua-kim, line 101: Why are we marking these as non-final?

src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseCleanupManager.java
```diff
@@ -189,62 +185,72 @@ public LeaseCleanupResult cleanupLease(LeasePendingDeletion leasePendingDeletion
                     Set<String> childShardKeys = leaseFromDDB.getChildShardIds();
                     if (CollectionUtils.isNullOrEmpty(childShardKeys)) {
                         try {
+                            // throws ResourceNotFoundException
                             childShardKeys = getChildShardsFromService(shardInfo);
-
                             if (CollectionUtils.isNullOrEmpty(childShardKeys)) {
                                 LOG.error("No child shards returned from service for shard " + shardInfo.getShardId());
+                                // If no children shard is found in DDB and from service, then do not delete the lease
```
joshua-kim, line 192: I don't think this offers any safety since this is just retried either way on the next deletion run, are we adding this for unit testing?

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
joshua-kim: Nvm, I see it's for unit test. Ideally we would be using builder pattern to avoid lengthy constructors, how bad is the build time if we don't override these values in tests?

src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseCleanupManager.java
ashwing: This will prevent the the lease deletion and log error as this closed and valid shard could not retrieve children info from anywhere. This is a final guard to protect against bad server response.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
ashwing: Yes, we should be refactoring the long arg constructors. temporarily giving this package level access. this saves more than 3 minutes of build time.

joshua-kim (commented):
LGTM

joshua-kim (approved):

PR #779, opened 2021-01-26
KCL 1.X - Guardrails to avoid checkpoint corruption during resharding

ashwing:
1. Update the checkpoint with non SHARD_END sequence number only if t…he DDB sequence number is not SHARD_END.
2. Verify the shard end checkpointing by directly looking up the ddb lease entry

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseManager.java
+++ b/src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseManager.java
@@ -20,6 +20,7 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 import com.amazonaws.services.dynamodbv2.model.BillingMode;
+import com.amazonaws.services.dynamodbv2.model.ExpectedAttributeValue;
 import com.amazonaws.services.kinesis.clientlibrary.lib.worker.KinesisClientLibConfiguration;
 import com.amazonaws.services.kinesis.leases.util.DynamoUtils;
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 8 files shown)


src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseManager.java
```diff
@@ -582,7 +583,9 @@ public boolean updateLease(T lease)
         UpdateItemRequest request = new UpdateItemRequest();
         request.setTableName(table);
         request.setKey(serializer.getDynamoHashKey(lease));
-        request.setExpected(serializer.getDynamoLeaseCounterExpectation(lease));
+        Map<String, ExpectedAttributeValue> expectations = serializer.getDynamoLeaseCounterExpectation(lease);
+        expectations.putAll(serializer.getDynamoLeaseCheckpointExpectation(lease));
+        request.setExpected(expectations);
```
joshua-kim, line 588: Won't this fail? I believe this is also used for leaseRenewal codepath, where we're heartbeating the lease counter. We don't necessarily expect `SHARD_END` in this case.

src/main/java/com/amazonaws/services/kinesis/leases/interfaces/ILeaseSerializer.java
```diff
@@ -74,6 +74,14 @@
      */
     public Map<String, ExpectedAttributeValue> getDynamoLeaseOwnerExpectation(T lease);
 
+    /**
+     * @param lease
+     * @return the attribute value map asserting that the checkpoint state is as expected.
+     */
+    default Map<String, ExpectedAttributeValue> getDynamoLeaseCheckpointExpectation(T lease) {
```
joshua-kim, line 81: Maybe good to rename this to something like `getDynamoLeaseShardEndCheckpointExpectation`?

src/main/java/com/amazonaws/services/kinesis/leases/impl/LeaseManager.java
ashwing: I don't see this being used anywhere other than checkpoint code path. And we don't expect SHARD_END here. We expect it not to present when we write a non SHARD_END checkpoint.

src/main/java/com/amazonaws/services/kinesis/leases/interfaces/ILeaseSerializer.java
ashwing: I just want to keep this generic, so that we can add any future state expectations. Also this would set expectation based on checkpoint in the lease.

joshua-kim (approved):
LGTM

PR #781, opened 2021-01-26
Fixing record processor processing deleted leases in cycle

ashwing:
Fixing record processor processing deleted leases in cycle

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/InitializeTask.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/InitializeTask.java
@@ -14,6 +14,7 @@
  */
 package com.amazonaws.services.kinesis.clientlibrary.lib.worker;
 
+import com.amazonaws.services.kinesis.clientlibrary.exceptions.KinesisClientLibException;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/InitializeTask.java
```diff
@@ -79,7 +80,15 @@ public TaskResult call() {
 
         try {
             LOG.debug("Initializing ShardId " + shardInfo.getShardId());
-            Checkpoint initialCheckpointObject = checkpoint.getCheckpointObject(shardInfo.getShardId());
+            Checkpoint initialCheckpointObject;
+            try {
+                initialCheckpointObject = checkpoint.getCheckpointObject(shardInfo.getShardId());
+            } catch (KinesisClientLibException e) {
```
joshua-kim, line 86: 1. Looks like a typo, looks like we actually throw `KinesisClientLibIOException`, not `KinesisClientLibException`. 
2. The same exception type is thrown for a lease not existing vs throttling/transient ddb issues, can we make a custom exception for this? Should be unit testable

ashwing: Nice catch. The API `getCheckpointObject` throws `KinesisClientLibException`, but the impl throws KinesisClientLibIOException which is a RuntimeException. 
Fixing to throw InvalidStateException which is a KinesisClientLibNonRetryableException, appropriate in this case and handling in the caller

joshua-kim (approved):
LGTM

PR #794, opened 2021-03-10
Silence warning when there are no stale streams to delete

tiaanl:
A warning is logged when stale steams are to be deleted, but the log is still emitted even if the map of streams to delete is empty.  This just silences the warning as it is emitted at least a few times per minute.


antoshen (approved):
The change looks good to me, we will merge the PR and the change will be released in 1-2 months.

PR #801, opened 2021-03-30
Add acquiredLock flag to prevent unlocking when no lock is held

Renjuju:
### Description of changes
* On exceptions prior to locking, the electLeaders will jump to the finally code block and *block* indefinitely
* This change adds an acquiredLock flag to ensure we only unlock in the finally code block if the lock was acquired.

### Testing

```java
public class Foo {

    public static void main(String[] args) throws IOException {
        final ScheduledExecutorService scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
        final Runner runner = new Runner();
        scheduledExecutorService.scheduleAtFixedRate(runner, 10L,10 ,TimeUnit.MILLISECONDS);
        System.out.println("Done");
    }

    static class Runner implements Runnable {

        ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
        @Override
        public void run() {
            System.out.println("Running");
            try {
                throw new TimeoutException("timeout");
            } catch (Throwable e) {
                e.printStackTrace();
            } finally {
                System.out.println("Forever blocking");
                readWriteLock.readLock().unlock();
                System.out.println("Did I get here?");
            }
        }
    }
}
```

**Output**:

```java
Done
Running
Forever blocking
java.util.concurrent.TimeoutException: timeout
    at Foo$Runner.run(Foo.java:26)
    at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
    at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
    at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
    at java.lang.Thread.run(Thread.java:748)
```

This shows that trying to unlock a *lock* that isn't *locked* will block indefinitely; one could say we're in a *deadlock* sort of situation here

### Other stuff

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DeterministicShuffleShardSyncLeaderDecider.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DeterministicShuffleShardSyncLeaderDecider.java
@@ -14,14 +14,6 @@
  */
 package software.amazon.kinesis.coordinator;
 
-import lombok.extern.slf4j.Slf4j;
-import software.amazon.awssdk.utils.CollectionUtils;
-import software.amazon.kinesis.leases.Lease;
-import software.amazon.kinesis.leases.LeaseRefresher;
-import software.amazon.kinesis.leases.exceptions.DependencyException;
-import software.amazon.kinesis.leases.exceptions.InvalidStateException;
-import software.amazon.kinesis.leases.exceptions.ProvisionedThroughputException;
-
 import java.time.Instant;
 import java.util.Collections;
 import java.util.HashSet;
@@ -34,6 +26,13 @@ import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.BooleanSupplier;
 import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
+import software.amazon.awssdk.utils.CollectionUtils;
+import software.amazon.kinesis.leases.Lease;
+import software.amazon.kinesis.leases.LeaseRefresher;
+import software.amazon.kinesis.leases.exceptions.DependencyException;
+import software.amazon.kinesis.leases.exceptions.InvalidStateException;
+import software.amazon.kinesis.leases.exceptions.ProvisionedThroughputException;
 
 /**
  * An implementation of the {@code LeaderDecider} to elect leader(s) based on workerId.
@@ -55,7 +54,7 @@ class DeterministicShuffleShardSyncLeaderDecider
     private static final long ELECTION_SCHEDULING_INTERVAL_MILLIS = 5 * 60 * 1000;
     private static final int AWAIT_TERMINATION_MILLIS = 5000;
 
-    private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
+    private final ReadWriteLock readWriteLock;
 
     private final LeaseRefresher leaseRefresher;
     private final int numPeriodicShardSyncWorkers;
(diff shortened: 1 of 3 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DeterministicShuffleShardSyncLeaderDecider.java
```diff
@@ -81,16 +81,18 @@
      */
     private void electLeaders() {
         try {
+            // In case value is currently being read, we wait for reading to complete before updating the variable.
+            // This is to prevent any ConcurrentModificationException exceptions.
+            readWriteLock.writeLock().lock();
```
ashwing, line 86: Can we refactor this to,
```
private void executeWithWriteLock(Runnable runnable) {
try {
 readWriteLock.writeLock().lock();
 runnable.run();
} finally {
 readWriteLock.writeLock().unlock();
}
}
```

and then call
`executeWithWriteLock(()-> electLeaders())` where now electLeaders() will have no lock code in it

Also check for other similar references. As an alternative, consider using tryLock as well.

     * <p>A typical usage idiom for this method would be:
     *  <pre> {@code
     * Lock lock = ...;
     * if (lock.tryLock()) {
     *   try {
     *     // manipulate protected state
     *   } finally {
     *     lock.unlock();
     *   }
     * } else {
     *   // perform alternative actions
     * }}</pre>
     *

Renjuju: Refactored! Thanks for the feedback. All other cases where we use locks are sufficiently protected against exceptions (they all use try catch without a route for getting in a blocked state)

See examples below from the code


```java
    public boolean submitShardSyncTask() {
        try {
            lock.lock();
            return checkAndSubmitNextTask();
        } finally {
            lock.unlock();
        }
    }
```

```java
    @Override
    public void restartFrom(RecordsRetrieved recordsRetrieved) {
        if (!(recordsRetrieved instanceof PrefetchRecordsRetrieved)) {
                    "Provided RecordsRetrieved was not produced by the PrefetchRecordsPublisher");
        }
        resetLock.writeLock().lock();
        try {
            publisherSession.reset((PrefetchRecordsRetrieved)recordsRetrieved);
            wasReset = true;
        } finally {
            resetLock.writeLock().unlock();
        }
    }
```

```java
    private void addArrivedRecordsInput(PrefetchRecordsRetrieved recordsRetrieved) throws InterruptedException {
        wasReset = false;
        while (!publisherSession.offerRecords(recordsRetrieved, idleMillisBetweenCalls)) {
            //
            // Unlocking the read lock, and then reacquiring the read lock, should allow any waiters on the write lock a
            // chance to run. If the write lock is acquired by restartFrom than the readLock will now block until
            // restartFrom(...) has completed. This is to ensure that if a reset has occurred we know to discard the
            // data we received and start a new fetch of data.
            //
            resetLock.readLock().unlock();
            resetLock.readLock().lock();
            if (wasReset) {
                throw new PositionResetException();
            }
        }
        publisherSession.prefetchCounters().added(recordsRetrieved.processRecordsInput);
    }
```

ashwing:
Ashwin Giridharan  3:19 PM
Hey Renju
3:19
sorry took another look at the PR
3:19
we dont need the lock to be moved to the top

Renju Radhakrishnan  3:20 PM
Ah I see. Why is that?

Ashwin Giridharan  3:20 PM
the listleases is a time consuming operation
3:20
we dont want the lock to the held for that much time

Renju Radhakrishnan  3:21 PM
Ahh gotcha. That makes sense
3:21
So removing the lock would be better

Ashwin Giridharan  3:21 PM
We only care about this assignment
3:21
leaders = new HashSet<>(uniqueHosts.subList(0, numShardSyncWorkers));

Renju Radhakrishnan  3:21 PM
Ah ok
3:21
Got it

Ashwin Giridharan  3:21 PM
no we have to keep the lock in its previous place
3:21
but we can add a variable like if lock acquired

Renju Radhakrishnan  3:21 PM
Ah ok cool

Ashwin Giridharan  3:22 PM
or the best way is to use tryLock(waitTimeMillis)
3:22
this would return boolean if the lock was acquired
3:22
release it in finally only if this is true
3:23
I posted an example in the comments

Renju Radhakrishnan  3:23 PM
Nice, that is better
3:23
I’ll use that

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DeterministicShuffleShardSyncLeaderDecider.java
```diff
@@ -68,18 +68,22 @@
      * @param leaderElectionThreadPool    Thread-pool to be used for leaderElection.
      * @param numPeriodicShardSyncWorkers Number of leaders that will be elected to perform periodic shard syncs.
      */
-    DeterministicShuffleShardSyncLeaderDecider(LeaseRefresher leaseRefresher, ScheduledExecutorService leaderElectionThreadPool,
-            int numPeriodicShardSyncWorkers) {
+    DeterministicShuffleShardSyncLeaderDecider(LeaseRefresher leaseRefresher,
```
ashwing, line 71: We need a new constructor that allows to take in the dedicated arg. We should not change constructors for public classes.

Renjuju: Sounds good. Added a new constructor

ashwing (approved):
lgtm

PR #841, opened 2021-07-26
Change hierarchicalShardSyncer getter to be public.

giancarlokc:
*Description of changes:*

This PR to make the `hierarchicalShardSyncer` getter in `LeaseManagementConfig.java` public again.

In v2.2.x the `hierarchicalShardSyncer` could be get and set via the lombok accessors. In v2.3.x a custom getter was introduced for accessing `hierarchicalShardSyncer`. The getter overwrote the lombok acessor, which is fine, but defined the access as `private`, preventing access to `hierarchicalShardSyncer` from outside the instance. I believe to be consistent we want the getter to be `public`, as are the getters of other variables.


joshua-kim (approved):
LGTM

Issue #845, opened 2021-08-06
Faster acquisition of leases from workers that have been gracefully shutdown?


rtyley:
New instances seem to take a relatively long time to acquire leases after old instances have stopped renewing them (seen in KCL 2.3.6, probably other versions too) - apart from incremental [lease stealing](https://github.com/awslabs/amazon-kinesis-client/blob/bd96580b0ac2ae45f43325e93b20d0d8b6be3ecf/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseTaker.java#L112-L119), the new instances seem to have to wait for the full `failoverTimeMillis` (which is referred to as `leaseDurationMillis` within `DynamoDBLeaseCoordinator` & `leaseDurationNanos` within `DynamoDBLeaseTaker`) before a lease is [considered expired](https://github.com/awslabs/amazon-kinesis-client/blob/0c5042dadf794fe988438436252a5a8fe70b6b0b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/Lease.java#L168-L185) and they can take it.

Although failover timeout is obviously a good for handling `ShardRecordProcessor`s that become unresponsive, if an instance is smoothly shutting down, and [`scheduler.startGracefulShutdown()`](https://github.com/awslabs/amazon-kinesis-client/blob/a25835678a0ca0e43caf03b0b326afeeb7e9e658/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java#L688-L730) has been called, couldn't it be clearly indicated that the old scheduler is no longer responsible for the lease by invoking [`evictLease()`](https://github.com/awslabs/amazon-kinesis-client/blob/0c5042dadf794fe988438436252a5a8fe70b6b0b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresher.java#L574-L606) (setting the lease owner to `null`) on the leases it still holds during graceful shutdown? This would be after `shutdownRequested()` is called on the `ShardRecordProcessor` & before `leaseLost()`. It could possibly be in [`s.a.k.lifecycle.ShutdownTask.call()`](https://github.com/awslabs/amazon-kinesis-client/blob/0c5042dadf794fe988438436252a5a8fe70b6b0b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShutdownTask.java#L125-L126)?

[...]

PR #846, opened 2021-08-06
Fix `DynamoDBLeaseTaker` logging of available leases

rtyley:
When `DynamoDBLeaseTaker` is deciding what leases to take to satisfy the target lease-count of it's worker, it can take either *expired* leases that haven't been updated by any other worker for a while, or failing that, 'steal' a limited number (by default, 1) of leases from other workers, in order to ensure its worker is gradually getting closer to its target number.

It turns out that the log message it prints out when taking expired leases is misleading. For instance, in this message - the leases weren't stolen, they were taken from the available 'expired' leases:

> Worker xxxxxx saw 68 total leases, 0 available leases, 6 workers. Target is 12 leases, I have 2 leases, I will take 10 leases

...yet the message says there were 0 available leases, so how is that possible? The truth is, there _were_ 10 available expired leases, but the log message was printing out the value of `expiredLeases.size()` - and `expiredLeases` was mutated over the course of the method, with entries removed and added to `leasesToTake`:

https://github.com/awslabs/amazon-kinesis-client/blob/0c5042dadf794fe988438436252a5a8fe70b6b0b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseTaker.java#L421

The fix to logging in this commit just records `numAvailableLeases` at the start of the method, so we can give an accurate log message:

> Worker xxxxxx saw 68 total leases, 10 available leases, 6 workers. Target is 12 leases, I have 2 leases, I will take 10 leases

While I was investigating #845 I found the broken log messaging quite confusing! 

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseTaker.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseTaker.java
@@ -347,6 +347,7 @@ public class DynamoDBLeaseTaker implements LeaseTaker {
         MetricsUtil.addWorkerIdentifier(scope, workerIdentifier);
         List<Lease> veryOldLeases = new ArrayList<>();
 
+        final int numAvailableLeases = expiredLeases.size();
         int numLeases = 0;
         int numWorkers = 0;
         int numLeasesToReachTarget = 0;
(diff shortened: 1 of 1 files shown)


rtyley:
Although the Travis CI build is failing for this small change, it looks like this is probably due to some kind of flakiness in the build, rather than a problem with the PR:

https://github.com/awslabs/amazon-kinesis-client/pull/846/checks?check_run_id=3266235711

image

joshua-kim:
LGTM. Thanks for the fix.

joshua-kim (approved):

PR #847, opened 2021-08-06
Make use of Java 8 to simplify `computeLeaseCounts()`

rtyley:
It's possible to remove quite a few lines of code from the method using Java 8 features (`groupingBy` Collector, etc).

Java 8 is the minimum supported version of the KCL since PR https://github.com/awslabs/amazon-kinesis-client/pull/176 was merged in July 2017.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #848, opened 2021-08-08
Treat unassigned leases (as well as expired ones) as available-to-be-taken

rtyley:
If a lease is 'unassigned' (it has no lease owner) then it should be considered available for taking in `DynamoDBLeaseTaker`. Prior to this change, the only ways `DynamoDBLeaseTaker` could take leases for a scheduler was either by incremental lease stealing, or waiting for the lease to expire by not having been updated in `failoverTimeMillis` - which could be slow if `failoverTimeMillis` was set reasonably high (with it set to just 30s, I've seen new instances take over 3 minutes to take all leases from old instances in a deployment).

This would be one half of a fix for #845 - the other half is invoking `evictLease()` (setting the lease owner to null) on graceful shutdown of a scheduler.


rtyley:
I've rebased this PR on top of the latest `master` commit, resolving the merge conflict with https://github.com/awslabs/amazon-kinesis-client/pull/861, the PR that reverted https://github.com/awslabs/amazon-kinesis-client/pull/847 - and also fixed the embarrassing problem with null `leaseOwner`s in the Java 8 code that I introduced with that PR (sincere apologies for that mistake).

The tests that @shanmsac introduced with #861 still pass with this updated PR, though I have tweaked one assertion to cope with the fact that the updated `computeActiveLeaseCountsByWorker()` method has changed with this PR to not return an entry for unassigned leases (so that a good active-worker count is simply given by the size of the map).

I hope this PR is still worth looking at, though obviously, having already managed to break things once with #847, I can understand why you might not trust this subsequent PR!

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseTaker.java
```diff
@@ -536,40 +528,28 @@ private void updateAllLeases(Callable<Long> timeProvider)
...
-                if (oldCount == null) {
-                    leaseCounts.put(leaseOwner, 1);
-                } else {
-                    leaseCounts.put(leaseOwner, oldCount + 1);
-                }
-            }
-        }
+        Map<String, Integer> activeLeaseCountsByWorker = allLeases.values().stream()
+            .filter(lease -> !lease.isUnassigned() && !availableLeasesSet.contains(lease))
+            .collect(groupingBy(Lease::leaseOwner, summingInt(lease -> 1)));
```
rtyley, line 547: Here filtering for `!lease.isUnassigned()` is what now stops the groupingBy `Lease::leaseOwner` from failing with the `NullPointerException: element cannot be mapped to a null key` error reported in https://github.com/awslabs/amazon-kinesis-client/pull/861 - given that unassigned leases are filtered out, there is no attempt to insert a `null` key entry in the map, so the code doesn't crash.

bjrara: Do we still want to keep this change given that #861 has been merged?

bjrara (commented):
Evicting leases during graceful shutdown would break the assumption here:
https://github.com/awslabs/amazon-kinesis-client/blob/a3e51d55959c0ce364cc6816386fb1b7cbc8a852/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseTaker.java#L334-L336

I'm also wondering if we can make use of resetting `lastCounterIncrementNanos` so that we don't need extra checks on `leaseOwner`?

PR #854, opened 2021-08-27
Create DynamoDB tables on On-Demand billing mode by default.

shanmsac:
* This will enable KCL to create the DynamoDB tables - Lease and ShardProgress with the On-Demand billing mode instead of provisioned billing mode.
* Updated the LeaseRefresher interface function createLeaseTableIfNotExists() to remove the provision mode parameters.
* Removed the integration test "DynamoDBLeaseRenewerBillingModePayPerRequestIntegrationTest.java" as PayPerRequest mode is tested by default integration test.

Signed-off-by: Sachin Sundar P S <shanmsac@amazon.com>

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseRefresher.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseRefresher.java
@@ -29,7 +29,7 @@ public interface LeaseRefresher {
 
     /**
      * Creates the table that will store leases. Succeeds if table already exists.
-     * Deprecated. Use createLeaseTableIfNotExists().
+     * Deprecated. Use {@link #createLeaseTableIfNotExists()}.
      *
      * @param readCapacity
      * @param writeCapacity
@@ -45,7 +45,8 @@ public interface LeaseRefresher {
             throws ProvisionedThroughputException, DependencyException;
 
     /**
-     * Creates the table that will store leases. Succeeds if table already exists.
+     * Creates the table that will store leases. Table is now created in PayPerRequest billing mode by default.
+     * Succeeds if table already exists.
      * 
      * @return true if we created a new table (table didn't exist before)
      * 
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresher.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresher.java
@@ -162,7 +162,6 @@ public class DynamoDBLeaseRefresher implements LeaseRefresher {
     @Override
     public boolean createLeaseTableIfNotExists(@NonNull final Long readCapacity, @NonNull final Long writeCapacity)
             throws ProvisionedThroughputException, DependencyException {
-        // DynamoDB is now created in PayPerRequest billing mode by default. Keeping this for backward compatibility.
         ProvisionedThroughput throughput = ProvisionedThroughput.builder().readCapacityUnits(readCapacity)
                 .writeCapacityUnits(writeCapacity).build();
         final CreateTableRequest request;
--- a/amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresherTest.java
+++ b/amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresherTest.java
@@ -21,6 +21,8 @@ import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.anyBoolean;
 import static org.mockito.Matchers.anyLong;
 import static org.mockito.Matchers.anyString;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
@@ -50,10 +52,14 @@ import software.amazon.awssdk.services.dynamodb.model.DeleteItemRequest;
 import software.amazon.awssdk.services.dynamodb.model.DeleteItemResponse;
 import software.amazon.awssdk.services.dynamodb.model.DescribeTableRequest;
 import software.amazon.awssdk.services.dynamodb.model.DescribeTableResponse;
+import software.amazon.awssdk.services.dynamodb.model.DynamoDbException;
 import software.amazon.awssdk.services.dynamodb.model.GetItemRequest;
 import software.amazon.awssdk.services.dynamodb.model.GetItemResponse;
+import software.amazon.awssdk.services.dynamodb.model.LimitExceededException;
+import software.amazon.awssdk.services.dynamodb.model.ProvisionedThroughput;
 import software.amazon.awssdk.services.dynamodb.model.PutItemRequest;
 import software.amazon.awssdk.services.dynamodb.model.PutItemResponse;
+import software.amazon.awssdk.services.dynamodb.model.ResourceInUseException;
 import software.amazon.awssdk.services.dynamodb.model.ResourceNotFoundException;
 import software.amazon.awssdk.services.dynamodb.model.ScanRequest;
 import software.amazon.awssdk.services.dynamodb.model.ScanResponse;
@@ -67,6 +73,8 @@ import software.amazon.kinesis.leases.Lease;
 import software.amazon.kinesis.leases.LeaseManagementConfig;
 import software.amazon.kinesis.leases.LeaseSerializer;
 import software.amazon.kinesis.leases.exceptions.DependencyException;
+import software.amazon.kinesis.leases.exceptions.ProvisionedThroughputException;
+
 @RunWith(MockitoJUnitRunner.class)
 public class DynamoDBLeaseRefresherTest {
 
@@ -100,6 +108,8 @@ public class DynamoDBLeaseRefresherTest {
     public ExpectedException expectedException = ExpectedException.none();
 
     private DynamoDBLeaseRefresher leaseRefresher;
+    private DescribeTableRequest describeTableRequest;
+    private CreateTableRequest createTableRequest;
 
     private Map<String, AttributeValue> serializedLease;
 
@@ -109,6 +119,13 @@ public class DynamoDBLeaseRefresherTest {
                 tableCreatorCallback);
         serializedLease = new HashMap<>();
 
+        describeTableRequest = DescribeTableRequest.builder().tableName(TABLE_NAME).build();
+        createTableRequest = CreateTableRequest.builder()
+                .tableName(TABLE_NAME)
+                .keySchema(leaseSerializer.getKeySchema())
+                .attributeDefinitions(leaseSerializer.getAttributeDefinitions())
+                .billingMode(BillingMode.PAY_PER_REQUEST)
+                .build();
     }
 
     @Test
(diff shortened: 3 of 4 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseRefresher.java
```diff
@@ -30,16 +30,13 @@
...
-     * @param readCapacity
-     * @param writeCapacity
-     * 
      * @return true if we created a new table (table didn't exist before)
      * 
      * @throws ProvisionedThroughputException if we cannot create the lease table due to per-AWS-account capacity
      *         restrictions.
      * @throws DependencyException if DynamoDB createTable fails in an unexpected way
      */
-    boolean createLeaseTableIfNotExists(Long readCapacity, Long writeCapacity)
```
joshua-kim, line 42: This is a public interface customers might be using; let's not mess with these signatures. We can make `createLeaseTableIfNotExists()` a separate signature to preserve backwards-compatible behavior.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRenewerBillingModePayPerRequestIntegrationTest.java
```diff
@@ -1,281 +0,0 @@
...
-import java.util.Map;
-import java.util.concurrent.Executors;
-
-import static org.hamcrest.CoreMatchers.equalTo;
-import static org.hamcrest.CoreMatchers.notNullValue;
-import static org.hamcrest.CoreMatchers.nullValue;
-import static org.junit.Assert.assertThat;
-
-@RunWith(MockitoJUnitRunner.class)
-public class DynamoDBLeaseRenewerBillingModePayPerRequestIntegrationTest extends
```
joshua-kim, line 38: Why are we deleting this test suite?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/LeaseIntegrationBillingModePayPerRequestTest.java
```diff
@@ -1,37 +0,0 @@
...
-import org.mockito.Mock;
-import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
-import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient;
-import software.amazon.awssdk.services.dynamodb.model.BillingMode;
-import software.amazon.kinesis.leases.dynamodb.DynamoDBLeaseRefresher;
-import software.amazon.kinesis.leases.dynamodb.DynamoDBLeaseSerializer;
-import software.amazon.kinesis.leases.dynamodb.TableCreatorCallback;
-
-@Slf4j
-public class LeaseIntegrationBillingModePayPerRequestTest extends LeaseIntegrationTest {
```
joshua-kim, line 30: Let's keep the existing integ tests

joshua-kim (changes requested):
Left one comment to not remove public interface method

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseRefresher.java
shanmsac: ya, kept this and marked as deprecated.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/LeaseIntegrationBillingModePayPerRequestTest.java
shanmsac: since keeping the old function signature, will keep the integ tests

shanmsac:
Updated as per comments

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresher.java
```diff
@@ -162,6 +161,16 @@ public DynamoDBLeaseRefresher(final String table, final DynamoDbAsyncClient dyna
     @Override
     public boolean createLeaseTableIfNotExists(@NonNull final Long readCapacity, @NonNull final Long writeCapacity)
             throws ProvisionedThroughputException, DependencyException {
+        // DynamoDB is now created in PayPerRequest billing mode by default. Keeping this for backward compatibility.
+        return createLeaseTableIfNotExists();
```
joshua-kim, line 165: Let's not change the old code path - this will change the behavior for any customer classes that use this.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/ExceptionThrowingLeaseRefresher.java
```diff
@@ -110,11 +110,17 @@ private void throwExceptions(String methodName, ExceptionThrowingLeaseRefresherM
 
     @Override
     public boolean createLeaseTableIfNotExists(Long readCapacity, Long writeCapacity)
+            throws ProvisionedThroughputException, DependencyException {
+        return createLeaseTableIfNotExists();
```
joshua-kim, line 114: See earlier comment

joshua-kim:
How did we test this? Can we add unit tests?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresher.java
shanmsac: addressed this.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/ExceptionThrowingLeaseRefresher.java
shanmsac: done.

shanmsac:
added unit tests for both table creation methods

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresher.java
```diff
@@ -162,16 +162,7 @@ public DynamoDBLeaseRefresher(final String table, final DynamoDbAsyncClient dyna
     @Override
     public boolean createLeaseTableIfNotExists(@NonNull final Long readCapacity, @NonNull final Long writeCapacity)
             throws ProvisionedThroughputException, DependencyException {
-        try {
```
joshua-kim, line 165: Why are we deleting this block of code? This also changes customer-facing behavior - let's remove this from the diff.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseRefresher.java
```diff
@@ -29,17 +29,31 @@
 
     /**
      * Creates the table that will store leases. Succeeds if table already exists.
-     * 
+     * Deprecated. Use createLeaseTableIfNotExists().
```
joshua-kim, line 32: You can use @link or @see javadoc annotation here to link people over to the new method.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseCoordinator.java
```diff
@@ -211,10 +211,9 @@ public void run() {
     @Override
     public void initialize() throws ProvisionedThroughputException, DependencyException, IllegalStateException {
         final boolean newTableCreated =
-                leaseRefresher.createLeaseTableIfNotExists(initialLeaseTableReadCapacity, initialLeaseTableWriteCapacity);
```
joshua-kim, line 214: There should be a unit test in `DynamoDBLeaseCoordinatorTest` to make sure that `createLeaseIfNotExists()` gets called in `initialize()` as part of validating this change.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresher.java
```diff
@@ -162,16 +162,7 @@ public DynamoDBLeaseRefresher(final String table, final DynamoDbAsyncClient dyna
...
-            if (tableStatus() != null) {
-                return newTableCreated;
-            }
-        } catch (DependencyException de) {
-            //
-            // Something went wrong with DynamoDB
-            //
-            log.error("Failed to get table status for {}", table, de);
-        }
+        // DynamoDB is now created in PayPerRequest billing mode by default. Keeping this for backward compatibility.
```
joshua-kim, line 165: Leaving this up to your preference, but I personally don't think this comment is needed since it's mentioned the interface's javadoc.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresherTest.java
```diff
@@ -265,7 +266,40 @@ public void testLeaseTableExistsTimesOut() throws Exception {
     }
 
     @Test
-    public void testCreateLeaseTableTimesOut() throws Exception {
+    public void testCreateLeaseTableProvisionedBillingModeIfNotExists() throws Exception {
+        leaseRefresher = new DynamoDBLeaseRefresher(TABLE_NAME, dynamoDbClient, leaseSerializer, CONSISTENT_READS,
+                tableCreatorCallback, LeaseManagementConfig.DEFAULT_REQUEST_TIMEOUT, BillingMode.PROVISIONED);
+
+        when(dynamoDbClient.describeTable(any(DescribeTableRequest.class))).thenReturn(mockDescribeTableFuture);
+        when(mockDescribeTableFuture.get(anyLong(), any()))
```
joshua-kim, line 274: Why `any()` here? Can we use the specific class we're expecting (if it's not possible to use the exact value).

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresherTest.java
```diff
@@ -265,7 +266,40 @@ public void testLeaseTableExistsTimesOut() throws Exception {
     }
 
     @Test
-    public void testCreateLeaseTableTimesOut() throws Exception {
+    public void testCreateLeaseTableProvisionedBillingModeIfNotExists() throws Exception {
+        leaseRefresher = new DynamoDBLeaseRefresher(TABLE_NAME, dynamoDbClient, leaseSerializer, CONSISTENT_READS,
+                tableCreatorCallback, LeaseManagementConfig.DEFAULT_REQUEST_TIMEOUT, BillingMode.PROVISIONED);
+
+        when(dynamoDbClient.describeTable(any(DescribeTableRequest.class))).thenReturn(mockDescribeTableFuture);
```
joshua-kim, line 273: I don't think we need to use `any(DescribeTableRequest.class)` since the table name is constant across tests, we can mock against a concrete object like `DescribeTableRequest.builder().tableName(TABLE_NAME).build()`.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresher.java
```diff
@@ -185,6 +176,34 @@ public boolean createLeaseTableIfNotExists(@NonNull final Long readCapacity, @No
...
+        return createTableIfNotExists(request);
+    }
+
+    private boolean createTableIfNotExists(CreateTableRequest request)
+            throws ProvisionedThroughputException, DependencyException {
+        try {
+            if (tableStatus() != null) {
+                return newTableCreated;
+            }
+        } catch (DependencyException de) {
```
joshua-kim, line 201: I know this is the existing behavior but I'm wondering why we don't throw this exception here. Seems like it makes more sense to fail here if we're not able to pull the lease table's status and let the caller decide how to proceed.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresherTest.java
```diff
@@ -265,7 +266,40 @@ public void testLeaseTableExistsTimesOut() throws Exception {
...
+        when(mockDescribeTableFuture.get(anyLong(), any()))
+                .thenThrow(ResourceNotFoundException.builder().message("Table doesn't exist").build());
+
+        when(dynamoDbClient.createTable(any(CreateTableRequest.class))).thenReturn(mockCreateTableFuture);
```
joshua-kim, line 277: Same with earlier comment

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresherTest.java
```diff
@@ -265,7 +266,40 @@ public void testLeaseTableExistsTimesOut() throws Exception {
     }
 
     @Test
-    public void testCreateLeaseTableTimesOut() throws Exception {
+    public void testCreateLeaseTableProvisionedBillingModeIfNotExists() throws Exception {
```
joshua-kim, line 269: Let's have more descriptive names for these tests - for example this test is testing behavior if the lease table already exists we should still return true when we catch the RNF exception.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresherTest.java
```diff
@@ -265,7 +266,40 @@ public void testLeaseTableExistsTimesOut() throws Exception {
...
+
+        when(dynamoDbClient.describeTable(any(DescribeTableRequest.class))).thenReturn(mockDescribeTableFuture);
+        when(mockDescribeTableFuture.get(anyLong(), any()))
+                .thenThrow(ResourceNotFoundException.builder().message("Table doesn't exist").build());
+
+        when(dynamoDbClient.createTable(any(CreateTableRequest.class))).thenReturn(mockCreateTableFuture);
+        when(mockCreateTableFuture.get(anyLong(), any())).thenReturn(null);
+
+        final boolean result = leaseRefresher.createLeaseTableIfNotExists(10L, 10L);
+
```
joshua-kim, line 281: Can we verify interactions w/ the ddb client?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRefresherTest.java
```diff
@@ -265,7 +266,40 @@ public void testLeaseTableExistsTimesOut() throws Exception {
...
+                .thenThrow(ResourceNotFoundException.builder().message("Table doesn't exist").build());
+
+        when(dynamoDbClient.createTable(any(CreateTableRequest.class))).thenReturn(mockCreateTableFuture);
+        when(mockCreateTableFuture.get(anyLong(), any())).thenReturn(null);
+
+        final boolean result = leaseRefresher.createLeaseTableIfNotExists();
+
+        Assert.assertTrue(result);
+    }
+
```
joshua-kim, line 298: We've only tested the cases where the table doesn't exist (i.e `ddbClient.describeTable()` throws a `ResourceNotFoundException`. Let's add a few other edge cases which can happen:

**DescribeTable**
1. If we throw a `DependencyException`, we should either continue trying to create the table (or throw the exception if we decide to propagate it up as per my earlier comment. I don't have a strong opinion on this since it's existing behavior).
**CreateTable**
2. If we throw a `ResourceNotFoundException`, we should expect to return true.
3. If we throw a `LimitExceededException`, we should propagate up a `ProvisionedThroughputException`.
4. If we throw a `DynamoDBException` we should throw a `DependencyException`.
5. If we throw a `TimeoutException` we should throw a `DependencyException`.

joshua-kim (changes requested):
Left a few comments on the unit tests

Issue #249, opened 2017-10-19
MillisBehindLatest metric across _all_ shards


usrenmae:
Currently several metrics, including `MillisBehindLatest` are reported to CloudWatch with one of the dimensions being a shard id. On the other side we find it very convenient to set CloudWatch alarms on top of this metric to be able to react, if _any_ shard starts to lag behind. Now it is not possible to set up alerts without specifying the exact name of the shard. This is a limiting factor, because once you add and remove shards constantly, the shard names are being very dynamic and each time they change, you need to change the alarms accordingly, which is frustrating. In general as one want to react to any shard lagging behind, it would be very nice to have a global `MillisBehindLatest` without relating it to any shard in its dimensions. This can be the maximum across all shards, like `MaxMillisBehindLatest`.

sahilpalvia:
Kinesis does emit a Stream level metrics for iterator age, called GetRecords.IteratorAgeMillis. You should be able to setup alarm on that metric. That metric can be found under the Kinesis namespace in CloudWatch. If you set the statistic for that metric to Maximum it'll map the maximum millisBehindLatest from all the shards for that given period. Please feel free to reopen the issue, if you still have questions.

usrenmae:
Thanks for informing about the `GetRecords.IteratorAgeMilliseconds` metric. I wasn't aware of this one. After a closer look into it I figured out it's a global per-stream metric of the Kinesis service. What I'm interested in is a per-consumer metric. We have multiple consumers running on the same stream, some of them may catch up the event feed perfectly, but others may lag. My idea was to have a metric which can tell you which particular consumer is lagging behind. It's not possible to get this information out of the `GetRecords.IteratorAgeMilliseconds` metric of Kinesis stream itself, but KCL could provide this metric similar way it provides the `MillisBehindLatest`, but without the shardId dimension.
Actually it is not convenient at all to have automation built around any shard-specific metrics, as shards are very dynamic on their own and may change in time, considering the fact that it is not possible to have an alarm on a metric with dimensions, but not specifying the dimension value. When monitoring is build on per-consumer basis, it's much more useful: one can setup permanent alarms on it and only in case of incident it's possible to trace back the particular shard with the shard-specific metrics already. 
Please re-opening the issue as suggested above.

sahilpalvia:
Thank you for the feedback. We agree with the change you have suggested, and will prioritize it accordingly against the other customer requests we receive.

StevenYCChou:
[...]

PR #868, opened 2021-11-23
Adding a new metric: Application-level MillisBehindLatest

QAQJ:
Feature request discussion: https://github.com/awslabs/amazon-kinesis-client/issues/249

According the [document](https://docs.aws.amazon.com/streams/latest/dev/monitoring-with-kcl.html#kcl-metrics-per-shard), `MillisBehindLatest` metric is used for monitoring how far the KCL application is behind to the latest record in the shard it's working with, and this is a shard-level metric. However a KCL application may work with multiple shards (or even streams), and as per customer's request, an application-level `MillisBehindLatest` metric would be helpful so that they only need to setup 1 alarm for monitoring the latencies of their KCL application, instead of setting an alarm for each shard. 

This PR implements the application-level `MillisBehindLatest` metric, simply by creating a new application-level metrics scope, along with the existing shard-level metrics scope, in `ProcessTask`, and use it to collect and publish the `MillisBehindLatest` data. This scope doesn't have `ShardId` or `StreamId` as its dimensions, so all the `MillisBehindLatest` will be published to a general metric in CloudWatch.

To view this metric, customer can log into their AWS console -> CloudWatch -> All Metrics -> name-of-their-application under Custom Namespaces section -> Operation -> then they should see the metric like below.
Screen Shot 2021-11-22 at 4 54 38 PM

The metric level is set to `SUMMARY`, same as shard-level `MillisBehindLatest` metric.

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ProcessTask.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ProcessTask.java
@@ -45,7 +45,7 @@ import software.amazon.kinesis.schemaregistry.SchemaRegistryDecoder;
 @KinesisClientInternalApi
 public class ProcessTask implements ConsumerTask {
     private static final String PROCESS_TASK_OPERATION = "ProcessTask";
-    private static final String APPLICATION_LEVEL_METRICS = "ApplicationLevelMetrics";
+    private static final String APPLICATION_TRACKER_OPERATION = "ApplicationTracker";
     private static final String DATA_BYTES_PROCESSED_METRIC = "DataBytesProcessed";
     private static final String RECORDS_PROCESSED_METRIC = "RecordsProcessed";
     private static final String RECORD_PROCESSOR_PROCESS_RECORDS_METRIC = "RecordProcessor.processRecords";
@@ -118,7 +118,7 @@ public class ProcessTask implements ConsumerTask {
          * therefore all data added to appScope, although from different shard consumer, will be sent to the same metric,
          * which is the app-level MillsBehindLatest metric.
          */
-        final MetricsScope appScope = MetricsUtil.createMetricsWithOperation(metricsFactory, APPLICATION_LEVEL_METRICS);
+        final MetricsScope appScope = MetricsUtil.createMetricsWithOperation(metricsFactory, APPLICATION_TRACKER_OPERATION);
         final MetricsScope shardScope = MetricsUtil.createMetricsWithOperation(metricsFactory, PROCESS_TASK_OPERATION);
         shardInfo.streamIdentifierSerOpt()
                 .ifPresent(streamId -> MetricsUtil.addStreamId(shardScope, StreamIdentifier.multiStreamInstance(streamId)));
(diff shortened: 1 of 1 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ProcessTask.java
```diff
@@ -45,6 +45,7 @@
 @KinesisClientInternalApi
 public class ProcessTask implements ConsumerTask {
     private static final String PROCESS_TASK_OPERATION = "ProcessTask";
+    private static final String APPLICATION_LEVEL_METRICS = "ApplicationLevelMetrics";
```
avahuang0429, line 48: Naming: doesn't say explicitly what this metric is about. 
Suggestion: AppMillisBehindLatest / something else that speaks about the metric and the visibility level

avahuang0429: I read the public documentation and found "ApplicationLevelMetrics" not descriptive enough. This new metric we are adding belongs to https://docs.aws.amazon.com/streams/latest/dev/monitoring-with-kcl.html#kcl-metrics-per-app. Refer to this naming convention on Operation field, and get feedback from PM on the naming.

QAQJ: I showed a demo to the PM previously, but yeah I agree, will talk with the PM to confirm the names.

avahuang0429:
General comment, can we try test the correctness of the metrics? I tried to look for test cases but there's none. Setting up brand new test case could be an option, but evaluate either that or manual testing

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ProcessTask.java
QAQJ: After talking with Nihar, we decided to use `ApplicationTracker` as the operation, and the code is updated now.

QAQJ:
@avahuang0429 I do have a testing script in my personal repo: https://github.com/QAQJ/amazon-kinesis-client/blob/dev-jiqilin/amazon-kinesis-client/src/test/java/software/amazon/kinesis/ApplicationTest.java, which runs a 2-stream application and the metric can be published to the cloudwatch. However the concern is that I wasn't able to create any latency to that application, so all MillsBehindLatest (including all shards & app-level) are all 0 my test. But theoretically, if any latency happens in any shard, by go to **Graphed Metrics** and choose "Maximum" in the **Statistic** dimension, like the second screenshot, customer should be able to see it.
Screen Shot 2021-12-02 at 10 58 08 AM
Screen Shot 2021-12-02 at 11 04 22 AM
o

avahuang0429 (approved):
LGTM, make sure to paste offline comments + retry build till pass before merging

QAQJ:
In my personal branch I've manually created some MillisBehindLatest latency: https://github.com/QAQJ/amazon-kinesis-client/tree/dev-jiqilin
When we set the Statistics to `Average` the metrics will show as the first image, we can see the ApplicationTracker.MillisBehindLatest is the average of the 2 shards.
Screen Shot 2021-12-06 at 8 47 13 AM
When we set the Statistics to `Maximum`, ApplicationTracker.MillisBehindLatest will be the same as the shard with largest latency. 
Screen Shot 2021-12-06 at 8 47 27 AM
Thus, KCL users can decide which statistic they want for their alarms.

Issue #900, opened 2022-01-12
Unable to shutdown before getting lease 


AndriiChuzhynov:
Hi, we have a problem with Kinesis when it tries to stop before the lease was taken. 
In this case, NullPointerException happens which prevents worker to shut down and it continues operating in a corrupted state

`java.util.concurrent.ExecutionException: java.lang.NullPointerException
	at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
	at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:205)
	at com.myapp.events.CompletableKinesisWorker.lambda$shutdown$0(CompletableKinesisWorker.java:65)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1736)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.exec(CompletableFuture.java:1728)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
**Caused by: java.lang.NullPointerException: null**
	at **software.amazon.kinesis.leases.dynamodb.DynamoDBLeaseCoordinator.stopLeaseTaker(DynamoDBLeaseCoordinator.java:332)**
	at software.amazon.kinesis.coordinator.Scheduler.lambda$createWorkerShutdownCallable$20(Scheduler.java:761)
	at software.amazon.kinesis.coordinator.GracefulShutdownCoordinator$GracefulShutdownCallable.call(GracefulShutdownCoordinator.java:152)
	at software.amazon.kinesis.coordinator.GracefulShutdownCoordinator$GracefulShutdownCallable.call(GracefulShutdownCoordinator.java:38)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.lang.Thread.run(Thread.java:829)
`

Issue #745, opened 2020-10-20
NullPointerException on Scheduler.startGreacefulShutdown


csgero:
Calling `Scheduler.startGracefulShutdown` on a scheduler too fast after the scheduler has been started results in a NullPointerException, and the scheduler is not shut down.
```
java.lang.NullPointerException
DynamoDBLeaseCoordinator.java:  332  software.amazon.kinesis.leases.dynamodb.DynamoDBLeaseCoordinator/stopLeaseTaker
            Scheduler.java:  755  software.amazon.kinesis.coordinator.Scheduler/lambda$createWorkerShutdownCallable$20
GracefulShutdownCoordinator.java:  152  software.amazon.kinesis.coordinator.GracefulShutdownCoordinator$GracefulShutdownCallable/call
GracefulShutdownCoordinator.java:   38  software.amazon.kinesis.coordinator.GracefulShutdownCoordinator$GracefulShutdownCallable/call
           FutureTask.java:  264  java.util.concurrent.FutureTask/run
               Thread.java:  834  java.lang.Thread/run
```

tom-beatdapp:
I have this exact issue too:

```
2020-10-22T22:50:53.8824996Z Exception while executing graceful shutdown: java.lang.NullPointerException
2020-10-22T22:50:53.8826886Z Oct 22, 2020 10:50:35 PM software.amazon.kinesis.coordinator.GracefulShutdownCoordinator$GracefulShutdownCallable call
2020-10-22T22:50:53.8828597Z WARNING: Caught exception while requesting initial worker shutdown.
2020-10-22T22:50:53.8829298Z java.lang.NullPointerException
2020-10-22T22:50:53.8831484Z 	at software.amazon.kinesis.leases.dynamodb.DynamoDBLeaseCoordinator.stopLeaseTaker(DynamoDBLeaseCoordinator.java:332)
2020-10-22T22:50:53.8834249Z 	at software.amazon.kinesis.coordinator.Scheduler.lambda$createWorkerShutdownCallable$20(Scheduler.java:755)
2020-10-22T22:50:53.8836790Z 	at software.amazon.kinesis.coordinator.GracefulShutdownCoordinator$GracefulShutdownCallable.call(GracefulShutdownCoordinator.java:152)
2020-10-22T22:50:53.8839892Z 	at software.amazon.kinesis.coordinator.GracefulShutdownCoordinator$GracefulShutdownCallable.call(GracefulShutdownCoordinator.java:38)
[...]

Issue #108, opened 2016-09-30
Checkpointing seems to fail after other worker has taken lease


scho:
Hi,

from time to time (every couple of days) we encounter the following situation (I've shortened all shard ids/sequence numbers to improve readability):

- `worker-a` processes `shard-01` and `shardId-02`.
- `worker-b` is started and takes lease for `shardId-02`.
- `worker-a` tries to checkpoint sequence number `00001` of `shardId-02`, but fails with the following reason:
```
com.amazonaws.services.kinesis.leases.impl.LeaseRenewer: Worker worker-a lost lease with key shardId-02 - discovered during update
```
- `worker-b` says:
```
com.amazonaws.services.kinesis.leases.impl.LeaseTaker: Worker worker-b successfully took 1 leases: shardId-02
com.amazonaws.services.kinesis.clientlibrary.lib.worker.KinesisDataFetcher: Initializing shard shardId-02 with 00001
```

The way I see it is, that `worker-a` seems to fail to checkpoint, but is successful instead.
The fact, that `worker-b` then initializes itself with the same sequence number, tells me, that checkpointing succeeded anyways. Otherwise, `worker-b` would not know anything about that particular sequence number.

Could this be a bug in the KPL?

We are using version `1.6.3` of the KPL.

pfifer:
It sounds like it did succeed it's last checkpoint.  The error message you would see when the checkpoint fails due to lease loss is [Can't update checkpoint - instance doesn't hold the lease for this shard](https://github.com/awslabs/amazon-kinesis-client/blob/master/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisClientLibLeaseCoordinator.java#L172)

The message you're seeing is likely related to the lease renewer finding that the record processor doesn't hold the lease anymore.  At that point in time the record processor will be terminated with the shutdown reason of ZOMBIE.

scho:
Sorry, I forgot to mention, that we get an exception on checkpointing as well. 

What we see on when `worker-a` is trying to checkpoint is:

The Log message (INFO) 2ms prior checkpointing:
```
com.amazonaws.services.kinesis.leases.impl.LeaseRenewer: Worker worker-a lost lease with key shardId-02 - discovered during update
```
The following exception when calling `IRecordProcessorCheckpointer::checkpoint(record)`:
```
[...]

PR #901, opened 2022-01-12
Fix null pointer exception that happens in shutdown before lease have got

AndriiChuzhynov:
*Issue #, if available:*
https://github.com/awslabs/amazon-kinesis-client/issues/900
https://github.com/awslabs/amazon-kinesis-client/issues/745

*Description of changes:*
This change adds a null check to avoid exceptions, in conditions when the worker didn't get the lease. 
Without this change, the shutdown process interrupts and workers are continuing to operate.

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
@@ -763,9 +763,12 @@ public class Scheduler implements Runnable {
             // Stop accepting new leases. Once we do this we can be sure that
             // no more leases will be acquired.
             //
-            leaseCoordinator.stopLeaseTaker();
+            Collection<Lease> leases = Collections.emptyList();
+            if (leaseCoordinator != null) {
+                leaseCoordinator.stopLeaseTaker();
+                leases = leaseCoordinator.getAssignments();
+            }
 
-            Collection<Lease> leases = leaseCoordinator.getAssignments();
             if (leases == null || leases.isEmpty()) {
                 //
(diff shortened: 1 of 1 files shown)


Novaes:
@pfifer @sahilpalvia is #900 #745  is a known issue? Any ideas what else we could do? Any extra insight would be greatly appreciated!

Novaes:
@pfifer @sahilpalvia is #900 #745 is a known issue? Any ideas what else we could do besides completing this client fix? Any extra insight would be greatly appreciated!

Novaes:
In summary, coordinator can't stop lease taker, so it impact us on end up with a lease unbalance between our instances for a new deploy. We thought the coordinator would re-balance, what also doesn't do: `software.amazon.kinesis.leases.dynamodb.DynamoDBLeaseCoordinator.stopLeaseTaker(DynamoDBLeaseCoordinator.java:332)` . On our end seems we can also enable the debugging mode mentioned [here](https://github.com/awslabs/amazon-kinesis-client/issues/108#issuecomment-256578391), to have more insights on it.

Novaes:
In summary, coordinator can't stop lease taker `software.amazon.kinesis.leases.dynamodb.DynamoDBLeaseCoordinator.stopLeaseTaker(DynamoDBLeaseCoordinator.java:332)`, so during new deploys we sometimes have a lease unbalance between our instances for a new deploy. We thought the coordinator would re-balance, what also doesn't do. On our end, seems we can also enable the debugging mode mentioned [here](https://github.com/awslabs/amazon-kinesis-client/issues/108#issuecomment-256578391), to have more insights on the re-balance. The NPE should be handled.

Issue #60, opened 2016-03-18
MultiLang Daemon


jasonrichardsmith:
Add new call to Multilang KCL

 * { "action" : "stdlogger",
 *   "message" : "my log message";
 * }

that will just pass the output to stdlog

pfifer:
I'm looking into adding this capability, but I don't currently have an ETA.

ababushkin:
Any update on this one @pfifer ?

mmatviyiv:
is there any legit way to pass through std logging from the worker?

CsiPeT:
This would be a great feature.. still...

mwarkentin:
Agreed

jamesritter15:
Any update on this? 👍

martyphee:
[I started playing around with adding a `LogMessage` to messages.](https://github.com/awslabs/amazon-kinesis-client/pull/962/files). Could use some cleanup and a helper class on the python KCL library to make things easier.

PR #962, opened 2022-06-17
MPHEE: first pass at adding logging for applications

martyphee:
[MultiLang Daemon StdLogger call](https://github.com/awslabs/amazon-kinesis-client/issues/60)

Adding the ability for the clients to send logging messages through the normal stdout by having the KCL process new LogMessages

WIP right now

```
2022-06-17 13:31:49,176 [ShardRecordProcessor-0001] INFO  s.a.k.multilang.messages.LogMessage [NONE] - Processing record with key AN_ID_OF_A_POLICY 
2022-06-17 13:31:49,176 [multi-lang-daemon-0003] INFO  s.a.kinesis.multilang.LineReaderTask [NONE] - Starting: Reading next message from STDIN for shardId-000000000000 
2022-06-17 13:31:49,177 [ShardRecordProcessor-0001] ERROR s.a.k.multilang.messages.LogMessage [NONE] - Processing record with key AN_ID_OF_A_POLICY 
```

Python side maybe?
```
class KclLogger:
    class LogLevel(Enum):
        INFO = "info"
        DEBUG = "debug"
        ERROR = "error"
        WARN = "warn"

    def debug(self, message: str):
        self._write_action(self.LogLevel.DEBUG.value, message)

    def info(self, message: str):
        self._write_action(self.LogLevel.INFO.value, message)

    def warn(self, message: str):
        self._write_action(self.LogLevel.WARN.value, message)

    def error(self, message: str):
        self._write_action(self.LogLevel.ERROR.value, message)

    @staticmethod
    def _write_action(log_level: LogLevel, message: str):
        kclprocess.io_handler.write_action(KclLogger._create_message(log_level, message))

    @staticmethod
    def _create_message(log_level: LogLevel, message: str):
        return {
            "action": "log",
            "logLevel": log_level,
            "message": message
        }
```

### DIFF
--- a/amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/MultiLangProtocol.java
+++ b/amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/MultiLangProtocol.java
@@ -27,14 +27,7 @@ import software.amazon.kinesis.lifecycle.events.LeaseLostInput;
 import software.amazon.kinesis.lifecycle.events.ProcessRecordsInput;
 import software.amazon.kinesis.lifecycle.events.ShardEndedInput;
 import software.amazon.kinesis.multilang.config.MultiLangDaemonConfiguration;
-import software.amazon.kinesis.multilang.messages.CheckpointMessage;
-import software.amazon.kinesis.multilang.messages.InitializeMessage;
-import software.amazon.kinesis.multilang.messages.LeaseLostMessage;
-import software.amazon.kinesis.multilang.messages.Message;
-import software.amazon.kinesis.multilang.messages.ProcessRecordsMessage;
-import software.amazon.kinesis.multilang.messages.ShardEndedMessage;
-import software.amazon.kinesis.multilang.messages.ShutdownRequestedMessage;
-import software.amazon.kinesis.multilang.messages.StatusMessage;
+import software.amazon.kinesis.multilang.messages.*;
 import software.amazon.kinesis.processor.RecordProcessorCheckpointer;
 
 /**
@@ -193,6 +186,10 @@ class MultiLangProtocol {
             if (checkpointFailed.orElse(false)) {
                 return false;
             }
+
+            Optional<Boolean> messageLogged = message.filter(m -> m instanceof LogMessage)
+                    .map(m -> (LogMessage) m)
+                    .map(m -> m.getLogger().apply(m.getMessage()));
 
             statusMessage = message.filter(m -> m instanceof StatusMessage).map(m -> (StatusMessage) m );
         }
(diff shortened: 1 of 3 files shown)


amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/MultiLangProtocol.java
```diff
@@ -27,14 +27,7 @@
...
 import software.amazon.kinesis.multilang.config.MultiLangDaemonConfiguration;
-import software.amazon.kinesis.multilang.messages.CheckpointMessage;
-import software.amazon.kinesis.multilang.messages.InitializeMessage;
-import software.amazon.kinesis.multilang.messages.LeaseLostMessage;
-import software.amazon.kinesis.multilang.messages.Message;
-import software.amazon.kinesis.multilang.messages.ProcessRecordsMessage;
-import software.amazon.kinesis.multilang.messages.ShardEndedMessage;
-import software.amazon.kinesis.multilang.messages.ShutdownRequestedMessage;
-import software.amazon.kinesis.multilang.messages.StatusMessage;
+import software.amazon.kinesis.multilang.messages.*;
```
brendan-p-lynch, line 30: We should not be adding wildcards for imports. This should explictly state all the things we need to import

amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/messages/LogMessage.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Default constructor.
+     */
+    public LogMessage(String message) {
+        this.message = message;
+        log.info("Client logging: " + this.message);
+    }
+
+    public Function<String, Boolean> getLogger() {
```
brendan-p-lynch, line 37: This looks to be returning only a boolean, not a Function<String, Boolean>. Also everything returns true. What is the benefit of returning a value

brendan-p-lynch (changes requested):
Please update this draft if you still would like this to be added

amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/messages/LogMessage.java
martyphee: It's returning a function

```
return (m) -> {
                     log.debug(m);
                     return true;
                 };
```

Not sure a better return type since `Function` requires one.

martyphee (commented):
I haven't looked at this for a couple years now. If you find it useful I can fix things up but I'm not using kinesis anymore.

PR #970, opened 2022-08-02
PeriodicShardSyncManager Changes Needed for DynamoDBStreamsKinesisAdapter Compatibility

nichgu:
Description of changes: Created the IPeriodicShardSync Manager and renamed the PeriodicShardSyncManager to the KinesisShardSyncManager to allow for compatibility with the DynamoDBStreamsKinesisAdapter

Testing: Testing was done with the [StreamsKCLAdapterCheck](https://code.amazon.com/packages/StreamsKCLAdapterCheck/trees/mainline) package which contains Kinesis KCL tests. Changes were also verified with the built in unit tests using mvn test.

### DIFF
--- a/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisPeriodicShardSyncManager.java
+++ b/src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisPeriodicShardSyncManager.java
@@ -1,0 +1,403 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates.
+ * Licensed under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.amazonaws.services.kinesis.clientlibrary.lib.worker;
+
+import java.io.Serializable;
+import java.math.BigInteger;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+import com.amazonaws.services.cloudwatch.model.StandardUnit;
+import com.amazonaws.services.kinesis.clientlibrary.proxies.IKinesisProxy;
+import com.amazonaws.services.kinesis.leases.exceptions.DependencyException;
+import com.amazonaws.services.kinesis.leases.exceptions.InvalidStateException;
+import com.amazonaws.services.kinesis.leases.exceptions.ProvisionedThroughputException;
+import com.amazonaws.services.kinesis.leases.impl.HashKeyRangeForLease;
+import com.amazonaws.services.kinesis.leases.impl.KinesisClientLease;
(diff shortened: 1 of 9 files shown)


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -1546,7 +1554,7 @@ public Worker build() {
                     workerStateChangeListener,
                     shardSyncer,
                     leaderDecider,
-                    null /* PeriodicShardSyncManager */);
```
shanmsac, line 1549: why are we passing this now? will this change any behavior?

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -1421,6 +1423,12 @@ public Worker build() {
                 throw new IllegalArgumentException(
                         "Kinesis Client Library configuration needs to be provided to build Worker");
             }
+            if (periodicShardSyncManager != null) {
+                if (leaseManager == null || shardSyncer == null || metricsFactory == null || leaderDecider == null) {
+
+                    throw new IllegalArgumentException("LeaseManager, ShardSyncer, MetricsFactory, and LeaderDecider must be provided if PeriodicShardSyncManager is provided");
+                }
+            }
```
shanmsac, line 1431: Why do we need this check? will the worker not start if not passed? will current implementations be affect with this?

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisPeriodicShardSyncManager.java
```diff
@@ -0,0 +1,403 @@
...
+
+/**
+ * The top level orchestrator for coordinating the periodic shard sync related activities. If the configured
+ * {@link ShardSyncStrategyType} is PERIODIC, this class will be the main shard sync orchestrator. For non-PERIODIC
+ * strategies, this class will serve as an internal auditor that periodically checks if the full hash range is covered
+ * by currently held leases, and initiates a recovery shard sync if not.
+ */
+@Getter
+@EqualsAndHashCode
+class KinesisPeriodicShardSyncManager implements IPeriodicShardSyncManager{
```
shanmsac, line 63: Just confirming, is this a copy of PeriodicShardSyncManager?

shanmsac (changes requested):
indent fixed part of this is making it hard to review, it will help if you can revert that

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
nichgu: That needs to be passed now because the StreamsWorkerFactory now passes the DynamoDBPeriodicShardSyncManager to the Worker, so the periodicShardSyncManager object now needs to be passed to the Worker. It shouldn't change behavior because if a PSSM is not passed, the object will be null and the behavior will be as normal with the getOrCreatePeriodicShardSyncManager function creating a KinesisPeriodicShardSyncManager

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
nichgu: All of those elements are used in the construction of a PeriodicShardSyncManager, so I added that check to make sure no duplicate elements like the shardSyncer are created if a PeriodicShardSyncManager has already passed to the Worker. Current implementations should not be affected as this check only happens if a PeriodicShardSyncManager is passed to the Worker which did not happen prior to my changes.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisPeriodicShardSyncManager.java
nichgu: Yes it is. Just a rename to differentiate from the DynamoDBStreamsPeriodicShardSyncManager

shanmsac (approved):
lgtm

PR #971, opened 2022-08-02
Everything else

nichgu:
Description of changes: The remaining changes require to allow for re-compatibility with the DynamoDBStreamsKinesisAdapter. The majority of changes consist of creating interfaces for key classes like the ShardConsumer, renaming these preexisting classes to include the Kinesis prefix (ShardConsumer -> KinesisShardConsumer) to distinguish from any classes that will use the new interfaces (i.e. DynamoDBStreamsShardConsumer), and making several utility classes and functions public so they could be accessed by the adapter or interfaces, such as the InitializationTask

Testing: Testing was done with the [StreamsKCLAdapterCheck](https://code.amazon.com/packages/StreamsKCLAdapterCheck/trees/mainline) package which contains Kinesis KCL tests. Changes were also verified with the built in unit tests using mvn test.


src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisShardConsumer.java
```diff
@@ -64,24 +61,36 @@ class ShardConsumer {
     private final long taskBackoffTimeMillis;
     private final boolean skipShardSyncAtWorkerInitializationIfLeasesExist;
 
-    @Getter
+    //@Getter
```
shanmsac, line 64: why are we commenting it out and adding separate get function? same for below two commented gets

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisShardConsumer.java
```diff
@@ -373,9 +382,9 @@ public boolean isSkipShardSyncAtWorkerInitializationIfLeasesExist() {
         return skipShardSyncAtWorkerInitializationIfLeasesExist;
     }
 
-    private enum TaskOutcome {
+    /*public enum TaskOutcome {
```
shanmsac, line 385: Since you have moved this to IShardConsumer, why not just remove this?

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
```diff
@@ -1134,7 +1138,12 @@ protected ShardConsumer buildConsumer(ShardInfo shardInfo, IRecordProcessorFacto
                         streamConfig.shouldValidateSequenceNumberBeforeCheckpointing()),
                 metricsFactory);
 
-        return new ShardConsumer(shardInfo,
+        if(shardConsumerFactory == null){
+
+            shardConsumerFactory = new KinesisShardConsumerFactory();
+        }
+
+        return shardConsumerFactory.createShardConsumer(shardInfo,
```
shanmsac, line 1146: Will there be any conflict if only shardConsumerFactory is set to Kinesis variant and all other are of different type?

shanmsac (changes requested):
most of the changes are just creating and using interface and updating class. had some minor comments

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisShardConsumer.java
nichgu: This is a carry over from the brazil testing as the build couldn't find the getter functions. I will verify that maven can properly read the getter, and remove these changes if tests pass.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/KinesisShardConsumer.java
nichgu: Just a carry over from testing. I will remove them.

src/main/java/com/amazonaws/services/kinesis/clientlibrary/lib/worker/Worker.java
nichgu: I am unsure what you mean by this. The chosen ShardConsumerFactory should create all ShardConsumers so they should all be the same type. Do you mean if the ShardConsumerFactory is the KinesisVariant but other classes like the PeriodicShardSyncManager are the DDB Streams Variant?

shanmsac: > Do you mean if the ShardConsumerFactory is the KinesisVariant but other classes like the PeriodicShardSyncManager are the DDB Streams Variant?

yes, will it cause any issue?

nichgu: So a normal KCL application shouldn't have access to anything outside of the standard KCL classes so the shardConsumerFactory and all other classes should be the Kinesis variant. Obviously in an adapter application, using the KinesisShardConsumerFactory with the DDB Streams classes will cause issues but standard KCL applications should never face that issue

shanmsac (approved):
lgtm, integ tests and application tests good.

PR #995, opened 2022-09-14
Everything else

gguptp:
Description of changes: The remaining changes require to allow for re-compatibility with the DynamoDBStreamsKinesisAdapter. The majority of changes consist of creating interfaces for key classes like the ShardConsumer, renaming these preexisting classes to include the Kinesis prefix (ShardConsumer -> KinesisShardConsumer) to distinguish from any classes that will use the new interfaces (i.e. DynamoDBStreamsShardConsumer), and making several utility classes and functions public so they could be accessed by the adapter or interfaces, such as the InitializationTask

Testing: Testing was done with the StreamsKCLAdapterCheck package which contains Kinesis KCL tests. Changes were also verified with the built in unit tests using mvn test.


shanmsac (approved):
lgtm

PR #1005, opened 2022-10-18
Added support for faster shard reallocation on graceful shutdown

mathieufortin01:
*Issue #, if available:* https://github.com/awslabs/amazon-kinesis-client/issues/845

*Description of changes:*
We have this latency issue in production where on graceful shutdown we have seen up to 40s before records resume processing on a new container. 

The cause of it is that on graceful shutdown, the lease stays owned. Others wait for lease to expire before picking them up. See https://github.com/awslabs/amazon-kinesis-client/issues/845 for details, this is pretty well explained.

The changes im proposing here are:
- Consider unassigned lease (without owners) as available for the taking (inspired by https://github.com/awslabs/amazon-kinesis-client/pull/848)
- Evict lease on graceful shutdown
- Expose leaseTakerIntervalMillis option so we can set a value that isnt based on the lease duration

With these changes, I was able to go below 1s before records resumed consumption. 

So I would like some opinions on this. We will test this in various envs and see the impacts. If this goes through this would be a contribution by Autodesk.

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
@@ -224,6 +224,18 @@ public class LeaseManagementConfig {
     private long listShardsCacheAllowedAgeInSeconds = 30;
     private int cacheMissWarningModulus = 250;
 
+    /**
+     * Interval at which the lease taker will execute.
+     * If unspecified, an interval will be calculated based on the lease duration.
+     */
+    private long leaseTakerIntervalMillis = -1L;
+
+    /**
+     * If leases should be evicted or not on shutdown requested.
+     * By default, leases are not evicted.
+     */
+    private boolean evictLeaseOnShutdown = false;
+
     private MetricsFactory metricsFactory = new NullMetricsFactory();
 
     @Deprecated
@@ -326,6 +338,7 @@ public class LeaseManagementConfig {
                     initialPositionInStream(),
                     failoverTimeMillis(),
                     epsilonMillis(),
+                    leaseTakerIntervalMillis,
                     maxLeasesForWorker(),
                     maxLeasesToStealAtOneTime(),
                     maxLeaseRenewalThreads(),
@@ -361,6 +374,7 @@ public class LeaseManagementConfig {
                     executorService(),
                     failoverTimeMillis(),
                     epsilonMillis(),
+                    leaseTakerIntervalMillis,
                     maxLeasesForWorker(),
                     maxLeasesToStealAtOneTime(),
                     maxLeaseRenewalThreads(),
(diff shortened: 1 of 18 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
```diff
@@ -224,6 +224,18 @@ public class LeaseManagementConfig {
     private long listShardsCacheAllowedAgeInSeconds = 30;
     private int cacheMissWarningModulus = 250;
 
+    /**
+     * Interval at which the lease taker will execute.
+     * If unspecified, an interval will be calculated based on the lease duration.
+     */
+    private long leaseTakerIntervalMillis = -1L;
```
mathieufortin01, line 231: Exposing this property to clients.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
```diff
@@ -224,6 +224,18 @@ public class LeaseManagementConfig {
...
+
+    /**
+     * If leases should be evicted or not on shutdown requested.
+     * By default, leases are not evicted.
+     */
+    private boolean evictLeaseOnShutdown = false;
```
mathieufortin01, line 237: New property. If false, same behavior as today which is wait for lease to be expired before taking it.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseTaker.java
```diff
@@ -191,9 +191,9 @@ synchronized Map<String, Lease> takeLeases(Callable<Long> timeProvider)
                 return takenLeases;
             }
 
-            List<Lease> expiredLeases = getExpiredLeases();
+            List<Lease> availableLeases = getAvailableLeases();
```
mathieufortin01, line 194: Changes in this file are all related to changing `expired` lease concept for `available` lease concept.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseTaker.java
```diff
@@ -482,7 +491,7 @@ private Set<Lease> computeLeasesToTake(List<Lease> expiredLeases) {
             }
 
         } finally {
-            scope.addData("ExpiredLeases", expiredLeases.size(), StandardUnit.COUNT, MetricsLevel.SUMMARY);
+            scope.addData("AvailableLeases", availableLeases.size(), StandardUnit.COUNT, MetricsLevel.SUMMARY);
```
mathieufortin01, line 494: Not sure whats the best way to go here.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseCoordinatorIntegrationTest.java
```diff
@@ -48,21 +45,19 @@
 import software.amazon.kinesis.retrieval.kpl.ExtendedSequenceNumber;
 
 @RunWith(MockitoJUnitRunner.class)
-public class DynamoDBLeaseCoordinatorIntegrationTest {
-    private static final int ATTEMPTS = 20;
+public class DynamoDBLeaseCoordinatorIntegrationTest extends LeaseIntegrationTest {
```
mathieufortin01, line 48: Now extends from `LeaseIntegrationTest`, so I could remove a whole bunch of test setup.

PR #1014, opened 2022-12-14
Use AFTER_SEQUENCE_NUMBER iterator type for expired iterator request

cylee99:
*Issue #, if available:*

*Description of changes:*

Updating the restartIterator logic in the KinesisDataFetcher class to use AFTER_SEQUENCE_NUMBER iterator call to prevent the last retrieved record from being retrieved again when the iterator expires. 


zengyu714 (approved):
lgtm

PR #1045, opened 2023-02-17
Fixed duplication of project version in children pom.xml

ZeyuLi-AWS:
*Issue #, if available:*

*Description of changes:*
* Removed explicit project version in children pom.xml 

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
   <artifactId>amazon-kinesis-client-pom</artifactId>
   <packaging>pom</packaging>
   <name>Amazon Kinesis Client Library</name>
-  <version>2.4.6-SNAPSHOT</version>
+  <version>${revision}</version>
   <description>The Amazon Kinesis Client Library for Java enables Java developers to easily consume and process data
     from Amazon Kinesis.
   </description>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -33,6 +33,7 @@
   </scm>
 
   <properties>
+    <revision>2.4.6-SNAPSHOT</revision>
```
stair-aws, line 36: Without context, `revision` is ambiguous. Suggest renaming to something more descriptive (e.g., `kcl.version`).

stair-aws (approved):
Verified "it blends", but recommend renaming the property to eliminate ambiguity.

@ZeyuLi-AWS Thanks for removing the duplication!

pom.xml
ZeyuLi-AWS: In https://maven.apache.org/maven-ci-friendly.html#multi-module-setup, it mentioned: `A note about the used properties. You can only use those named ${revision}, ${sha1} and/or ${changelist} and not other named properties like this` 
Other names will be prohibited.

stair-aws: That limitation appears to be from `flatten-maven-plugin` [interpolation](https://github.com/mojohaus/flatten-maven-plugin/blob/c613968d1fe7b748945e4e232ecbe66739610feb/src/main/java/org/codehaus/mojo/flatten/cifriendly/CiModelInterpolator.java#L149-L153) -- a plugin we do not use today. Other property names appear to work.

LGTM either way!

PR #1063, opened 2023-03-14
Allow leader to learn about new leases upon leader change to avoid unnecessary shardSyncs

chenylee-aws:
*Description of changes:*

Upon leader change the new leader might not have the latest streams in its in-memory map therefore would enqueue shardSyncs for these streams. It can be a waste if the shardSsyncs result in no leases created as they have been created by the previous leader. Instead, we can let the `PeriodicShardSyncManager` to take care of updating these streams by first checking the hash range completion before deciding to enqueuing them for shardSyncs to save IOPs.  

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
@@ -180,7 +180,7 @@ public class Scheduler implements Runnable {
     private final Object lock = new Object();
 
     private final Stopwatch streamSyncWatch = Stopwatch.createUnstarted();
-    private boolean leasesSyncedOnAppInit = false;
+    private boolean shouldSyncLeases = true;
 
     /**
      * Used to ensure that only one requestedShutdown is in progress at a time.
@@ -279,8 +279,6 @@ public class Scheduler implements Runnable {
                 PERIODIC_SHARD_SYNC_MAX_WORKERS_DEFAULT);
         this.failoverTimeMillis = this.leaseManagementConfig.failoverTimeMillis();
         this.taskBackoffTimeMillis = this.lifecycleConfig.taskBackoffTimeMillis();
-//        this.retryGetRecordsInSeconds = this.retrievalConfig.retryGetRecordsInSeconds();
-//        this.maxGetRecordsThreadPool = this.retrievalConfig.maxGetRecordsThreadPool();
         this.listShardsBackoffTimeMillis = this.retrievalConfig.listShardsBackoffTimeInMillis();
         this.maxListShardsRetryAttempts = this.retrievalConfig.maxListShardsRetryAttempts();
         this.shardDetectorProvider = streamConfig -> createOrGetShardSyncTaskManager(streamConfig).shardDetector();
@@ -419,6 +417,8 @@ public class Scheduler implements Runnable {
             // check for new streams and sync with the scheduler state
             if (isLeader()) {
                 checkAndSyncStreamShardsAndLeases();
+            } else {
+                shouldSyncLeases = true;
             }
 
             logExecutorState();
@@ -426,7 +426,7 @@ public class Scheduler implements Runnable {
             Thread.sleep(shardConsumerDispatchPollIntervalMillis);
         } catch (Exception e) {
             log.error("Worker.run caught exception, sleeping for {} milli seconds!",
-                    String.valueOf(shardConsumerDispatchPollIntervalMillis), e);
+                    shardConsumerDispatchPollIntervalMillis, e);
             try {
                 Thread.sleep(shardConsumerDispatchPollIntervalMillis);
             } catch (InterruptedException ex) {
@@ -454,15 +454,17 @@ public class Scheduler implements Runnable {
             final MetricsScope metricsScope = MetricsUtil.createMetricsWithOperation(metricsFactory, MULTI_STREAM_TRACKER);
 
             try {
-                // This is done to ensure that we clean up the stale streams lingering in the lease table.
-                if (!leasesSyncedOnAppInit && isMultiStreamMode) {
-                    final List<MultiStreamLease> leases = fetchMultiStreamLeases();
-                    syncStreamsFromLeaseTableOnAppInit(leases);
-                    leasesSyncedOnAppInit = true;
-                }
-
                 final Map<StreamIdentifier, StreamConfig> newStreamConfigMap = streamTracker.streamConfigList()
                         .stream().collect(Collectors.toMap(StreamConfig::streamIdentifier, Function.identity()));
+                // This is done to ensure that we clean up the stale streams lingering in the lease table.
+                if (shouldSyncLeases && isMultiStreamMode) {
+                    // Skip updating the stream map due to no new stream since last sync
+                    if (newStreamConfigMap.keySet().stream().anyMatch(s -> !currentStreamConfigMap.containsKey(s))) {
+                        syncStreamsFromLeaseTable(fetchMultiStreamLeases());
+                    }
+                    shouldSyncLeases = false;
+                }
+
                 // For new streams discovered, do a shard sync and update the currentStreamConfigMap
                 for (StreamIdentifier streamIdentifier : newStreamConfigMap.keySet()) {
                     if (!currentStreamConfigMap.containsKey(streamIdentifier)) {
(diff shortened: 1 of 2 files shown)


amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/SchedulerTest.java
```diff
@@ -974,6 +973,36 @@ public void testErrorHandlerForUndeliverableAsyncTaskExceptions() {
...
+    public void testNotUpdatingStreamMapAsItContainsAllStreams() throws Exception {
+        final List<StreamConfig> streamConfigList = createDummyStreamConfigList(1, 6);
+        retrievalConfig = new RetrievalConfig(kinesisClient, multiStreamTracker, applicationName)
+                .retrievalFactory(retrievalFactory);
+        scheduler = spy(new Scheduler(checkpointConfig, coordinatorConfig, leaseManagementConfig, lifecycleConfig,
+                metricsConfig, processorConfig, retrievalConfig));
+        when(multiStreamTracker.streamConfigList()).thenReturn(streamConfigList);
+        when(scheduler.shouldSyncStreamsNow()).thenReturn(true);
+        // Populate currentStreamConfigMap to simulate that the leader has the latest streams.
+        streamConfigList.forEach(s -> scheduler.currentStreamConfigMap().put(s.streamIdentifier(), s));
```
stair-aws, line 1000: [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). With the exception of this line and the `VerificationMode` (line ~1003), this test is identical to the one above it. Create a reusable method.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/SchedulerTest.java
```diff
@@ -974,6 +973,36 @@ public void testErrorHandlerForUndeliverableAsyncTaskExceptions() {
...
+
+        scheduler.checkAndSyncStreamShardsAndLeases();
+        verify(scheduler, times(0)).syncStreamsFromLeaseTable(any());
```
stair-aws, line 1003: Nit: `s/times(0)/never()/`

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -180,7 +180,7 @@ public class Scheduler implements Runnable {
     private final Object lock = new Object();
 
     private final Stopwatch streamSyncWatch = Stopwatch.createUnstarted();
-    private boolean leasesSyncedOnAppInit = false;
+    private boolean shouldSyncLeases = true;
```
stair-aws, line 183: Unfortunately, this is not a backwards-compatible change since `Scheduler` has exposed its member variables via accessors. I'll reach out to you internally so we can discuss release strategy.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -454,15 +454,17 @@ Set<StreamIdentifier> checkAndSyncStreamShardsAndLeases()
...
-                if (!leasesSyncedOnAppInit && isMultiStreamMode) {
-                    final List<MultiStreamLease> leases = fetchMultiStreamLeases();
-                    syncStreamsFromLeaseTableOnAppInit(leases);
-                    leasesSyncedOnAppInit = true;
+                if (shouldSyncLeases && isMultiStreamMode) {
+                    // Skip updating the stream map due to no new stream since last sync
+                    if (newStreamConfigMap.keySet().stream().anyMatch(s -> !currentStreamConfigMap.containsKey(s))) {
+                        syncStreamsFromLeaseTable(fetchMultiStreamLeases());
+                    }
+                    shouldSyncLeases = false;
```
abhit17, line 469: this would mean we will be doing the perfoming leaseSyncs on every host now which would lead to increase in the Lease table IOPs significantly, specifically on deployment.

I think we should consider evaluating the solution where on new leader itself we evaluate if shardSync is needed (rather than doing shardSync blindly) if not needed then simply update the in-memory state else perform shardSync(as this is requirement anyways). 

Check the logic in PeriodicShardSyncManager which is optimized logic and does not lead to many scan operations but also perform shardSync only when needed. Today in Scheduler we are assuming if leader is not info of stream then shardSync is needed which is not correct and can be optimized IMHO. Thoughts?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
chenylee-aws: Created separate non-assessable variable and kept the current logic untouched

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
chenylee-aws: Letting `PeriodicShardSyncManager` to do all the shard sync in one place is also the ideal solution I have in my mind as pointed out in the overview that PeriodicShardSyncManager already has all the functionality but here are my concerns around going to that approach directly. 

1. Incremental change - this change has smaller blast radius and provides IOPs improvements already in most of the cases.
2. Backward compatibility - the logic we have now is that leader learns new streams and enqueues directly which might be faster than letting the periodic shard sync manager to start doing shard sync. KCL users might set really high wait time between each periodic shard sync runs and this would cause delay in learning and crating lease for the new streams.

abhit17: As per offline discussion, looks good and this should not have the IOPs increase with deployment as the change is on the leader side only and not on each host. 

Logic looks good, Lets test this change also properly to ensure we do not miss or add any edge condition.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/SchedulerTest.java
```diff
@@ -974,6 +958,54 @@ public void testErrorHandlerForUndeliverableAsyncTaskExceptions() {
...
+        final List<StreamConfig> streamConfigList = createDummyStreamConfigList(1, 6);
+        prepareMultiStreamScheduler(streamConfigList);
+        // Populate currentStreamConfigMap to simulate that the leader has the latest streams.
+        streamConfigList.forEach(s -> scheduler.currentStreamConfigMap().put(s.streamIdentifier(), s));
+        scheduler.checkAndSyncStreamShardsAndLeases();
+        verify(scheduler, never()).syncStreamsFromLeaseTableOnAppInit(any());
+    }
+
+    @SafeVarargs
+    private final void prepareMultiStreamScheduler(List<StreamConfig>... streamConfigs) {
```
stair-aws, line 979: FYI: `private final` can be simplified to `private` (and some IDEs will flag this). Private methods are not visible for subclasses to override.

stair-aws (approved):

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/SchedulerTest.java
chenylee-aws: My IDE doesn't not like the removal of the `final` keyword. `@SafeVarargs is not allowed on non-final instance methods`

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -460,7 +460,7 @@ Set<StreamIdentifier> checkAndSyncStreamShardsAndLeases()
                 final Map<StreamIdentifier, StreamConfig> newStreamConfigMap = streamTracker.streamConfigList()
                         .stream().collect(Collectors.toMap(StreamConfig::streamIdentifier, Function.identity()));
                 // This is done to ensure that we clean up the stale streams lingering in the lease table.
-                if (isMultiStreamMode && (shouldSyncLeases || leasesSyncedOnAppInit)) {
+                if (isMultiStreamMode && (shouldSyncLeases || !leasesSyncedOnAppInit)) {
```
stair-aws, line 463: +1 Great catch.

stair-aws (approved):

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
chenylee-aws: Have performed and documented two tests.

T1: Once a worker is being re-elected for leader it can do DDB lookup to update its in memory map if detecting  new stream.

T2: During the lifetime of a leader the DDB lookup is executed at most once.

PR #1087, opened 2023-04-07
Internally construct StreamARN using STS

zengyu714:
*Issue #, if available:*

*Description of changes:*

- Internally, this change constructs a StreamARN using the accountId and aws partition obtained from STS `getCallerIdentity()`, while the other fields in the StreamARN, such as the streamName and region, are provided by the customer and extracted from the `kinesisClient` respectively
- This change maintains backward compatibility by not failing the consumer application when an STS call fails. However, if it is desired to fail the customer in such a scenario, a major version bump may be necessary to indicate the incompatibility.

*Testing*
- Added unit tests for the helper class`StreamARNUtil.java`
- `mvn clean install`
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Amazon Kinesis Client Library 2.5.0-SNAPSHOT:
[INFO]
[INFO] Amazon Kinesis Client Library ...................... SUCCESS [  0.170 s]
[INFO] Amazon Kinesis Client Library for Java ............. SUCCESS [02:25 min]
[INFO] amazon-kinesis-client-multilang .................... SUCCESS [  5.718 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  02:31 min
[INFO] Finished at: 2023-04-13T23:46:33Z
[INFO] ------------------------------------------------------------------------
```
- Manual sanity test using `AWSKinesisClientLibraryTester` and observed that the requests were using StreamARN as expected

```
2023-04-13 16:54:31,850 [ShardRecordProcessor-0001] INFO  s.a.k.leases.KinesisShardDetector [NONE] - Stream KCLGetRegion: listing shards with list shards request ListShardsRequest(StreamName=KCLGetRegion, StreamARN=arn:aws:kinesis:us-west-2:448057645220:stream/KCLGetRegion) 
```

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/ConfigsBuilder.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/ConfigsBuilder.java
@@ -142,7 +142,7 @@ public class ConfigsBuilder {
             @NonNull KinesisAsyncClient kinesisClient, @NonNull DynamoDbAsyncClient dynamoDBClient,
             @NonNull CloudWatchAsyncClient cloudWatchClient, @NonNull String workerIdentifier,
             @NonNull ShardRecordProcessorFactory shardRecordProcessorFactory) {
-        this(new SingleStreamTracker(streamName),
+        this(new SingleStreamTracker(streamName, kinesisClient.serviceClientConfiguration().region()),
                 applicationName,
                 kinesisClient,
                 dynamoDBClient,
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamARNUtil.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamARNUtil.java
@@ -1,0 +1,86 @@
+package software.amazon.kinesis.common;
+
+import com.google.common.base.Joiner;
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
+import software.amazon.awssdk.arns.Arn;
+import software.amazon.awssdk.awscore.exception.AwsServiceException;
+import software.amazon.awssdk.core.exception.SdkClientException;
+import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.sts.StsClient;
+import software.amazon.awssdk.services.sts.model.GetCallerIdentityResponse;
+
+import java.util.HashMap;
+import java.util.Optional;
+
+@Slf4j
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class StreamARNUtil {
+    private static final HashMap<String, Arn> streamARNCache = new HashMap<>();
+
+    /**
+     * This static method attempts to retrieve the stream ARN using the stream name, region, and accountId returned by STS
+     * It is designed to fail gracefully, returning Optional.empty() if any errors occur.
+     * @param streamName: stream name
+     * @param kinesisRegion: kinesisRegion is a nullable parameter used to construct the stream arn
+     * @return
+     */
+    public static Optional<Arn> getStreamARN(String streamName, Region kinesisRegion) {
+        return getStreamARN(streamName, kinesisRegion, Optional.empty());
+    }
+
+    public static Optional<Arn> getStreamARN(String streamName, Region kinesisRegion, @NonNull Optional<String> accountId) {
+        if (kinesisRegion == null || StringUtils.isEmpty(kinesisRegion.toString())) {
+            return Optional.empty();
+        }
+        // Consult the cache before contacting STS
+        String key = getCacheKey(streamName, kinesisRegion, accountId);
+        if (streamARNCache.containsKey(key)) {
+            return Optional.of(streamARNCache.get(key));
+        }
+
+        Optional<Arn> stsCallerArn = getStsCallerArn();
+        if (!stsCallerArn.isPresent() || !stsCallerArn.get().accountId().isPresent()) {
+            return Optional.empty();
+        }
+        accountId = accountId.isPresent() ? accountId : stsCallerArn.get().accountId();
+        Arn kinesisStreamArn = Arn.builder()
+                .partition(stsCallerArn.get().partition())
+                .service("kinesis")
+                .region(kinesisRegion.toString())
+                .accountId(accountId.get())
+                .resource("stream/" + streamName)
+                .build();
+
+        // Update the cache
+        streamARNCache.put(key, kinesisStreamArn);
+        return Optional.of(kinesisStreamArn);
+    }
+
+    private static Optional<Arn> getStsCallerArn() {
+        GetCallerIdentityResponse response;
+        try {
+            response = getStsClient().getCallerIdentity();
+        } catch (AwsServiceException | SdkClientException e) {
+            log.warn("Unable to get sts caller identity to build stream arn", e);
+            return Optional.empty();
+        }
+        return Optional.of(Arn.fromString(response.arn()));
+    }
+
+    private static StsClient getStsClient() {
+        return StsClient.builder()
+                .httpClient(UrlConnectionHttpClient.builder().build())
+                .build();
+    }
+
+    private static String getCacheKey(
+            String streamName, @NonNull Region kinesisRegion, @NonNull Optional<String> accountId) {
+        return Joiner.on(":").join(streamName, kinesisRegion.toString(), accountId.orElse(""));
+    }
+
+}
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamIdentifier.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamIdentifier.java
@@ -16,38 +16,39 @@
 package software.amazon.kinesis.common;
 
 import com.google.common.base.Joiner;
+import lombok.AccessLevel;
+import lombok.Builder;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
-import lombok.NonNull;
+import lombok.ToString;
 import lombok.experimental.Accessors;
+import software.amazon.awssdk.arns.Arn;
+import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.utils.Validate;
 
 import java.util.Optional;
 import java.util.regex.Pattern;
 
-@EqualsAndHashCode @Getter @Accessors(fluent = true)
+@Builder(access = AccessLevel.PRIVATE)
+@EqualsAndHashCode
+@Getter
+@ToString
+@Accessors(fluent = true)
 public class StreamIdentifier {
-    private final Optional<String> accountIdOptional;
+    @Builder.Default
+    private final Optional<String> accountIdOptional = Optional.empty();
     private final String streamName;
-    private final Optional<Long> streamCreationEpochOptional;
+    @Builder.Default
+    private final Optional<Long> streamCreationEpochOptional = Optional.empty();
+    @Builder.Default
+    private final Optional<Arn> streamARNOptional = Optional.empty();
 
     private static final String DELIMITER = ":";
-    private static final Pattern PATTERN = Pattern.compile(".*" + ":" + ".*" + ":" + "[0-9]*");
-
-    private StreamIdentifier(@NonNull String accountId, @NonNull String streamName, @NonNull Long streamCreationEpoch) {
-        this.accountIdOptional = Optional.of(accountId);
-        this.streamName = streamName;
-        this.streamCreationEpochOptional = Optional.of(streamCreationEpoch);
-    }
-
-    private StreamIdentifier(@NonNull String streamName) {
-        this.accountIdOptional = Optional.empty();
-        this.streamName = streamName;
-        this.streamCreationEpochOptional = Optional.empty();
-    }
+    private static final Pattern PATTERN = Pattern.compile(".*" + ":" + ".*" + ":" + "[0-9]*" + ":?([a-z]{2}(-gov)?-[a-z]+-\\d{1})?");
 
     /**
      * Serialize the current StreamIdentifier instance.
+     * TODO: Consider appending region info for cross-account consumer support
      * @return
      */
     public String serialize() {
@@ -63,14 +64,35 @@ public class StreamIdentifier {
 
     /**
      * Create a multi stream instance for StreamIdentifier from serialized stream identifier.
-     * The serialized stream identifier should be of the format account:stream:creationepoch
+     * See the format of a serialized stream identifier at {@link StreamIdentifier#multiStreamInstance(String, Region)}
      * @param streamIdentifierSer
      * @return StreamIdentifier
      */
     public static StreamIdentifier multiStreamInstance(String streamIdentifierSer) {
+        return multiStreamInstance(streamIdentifierSer, null);
+    }
+
+    /**
+     * Create a multi stream instance for StreamIdentifier from serialized stream identifier.
+     * @param streamIdentifierSer The serialized stream identifier should be of the format
+     *                            account:stream:creationepoch[:region]
+     * @param kinesisRegion This nullable region is used to construct the optional StreamARN
+     * @return StreamIdentifier
+     */
+    public static StreamIdentifier multiStreamInstance(String streamIdentifierSer, Region kinesisRegion) {
         if (PATTERN.matcher(streamIdentifierSer).matches()) {
             final String[] split = streamIdentifierSer.split(DELIMITER);
-            return new StreamIdentifier(split[0], split[1], Long.parseLong(split[2]));
+            final String streamName = split[1];
+            final Optional<String> accountId = Optional.ofNullable(split[0]);
+            StreamIdentifierBuilder builder = StreamIdentifier.builder()
+                    .accountIdOptional(accountId)
+                    .streamName(streamName)
+                    .streamCreationEpochOptional(Optional.of(Long.parseLong(split[2])));
+            final Region region = (split.length == 4) ?
+                    Region.of(split[3]) :   // Use the region extracted from the serialized string, which matches the regex pattern
+                    kinesisRegion;          // Otherwise just use the provided region
+            final Optional<Arn> streamARN = StreamARNUtil.getStreamARN(streamName, region, accountId);
+            return builder.streamARNOptional(streamARN).build();
         } else {
             throw new IllegalArgumentException("Unable to deserialize StreamIdentifier from " + streamIdentifierSer);
         }
@@ -82,7 +104,21 @@ public class StreamIdentifier {
      * @return StreamIdentifier
      */
     public static StreamIdentifier singleStreamInstance(String streamName) {
+        return singleStreamInstance(streamName, null);
+    }
+
+    /**
+     * Create a single stream instance for StreamIdentifier from the provided stream name and kinesisRegion.
+     * This method also constructs the optional StreamARN based on the region info.
+     * @param streamName
+     * @param kinesisRegion
+     * @return StreamIdentifier
+     */
+    public static StreamIdentifier singleStreamInstance(String streamName, Region kinesisRegion) {
         Validate.notEmpty(streamName, "StreamName should not be empty");
-        return new StreamIdentifier(streamName);
+        return StreamIdentifier.builder()
+                .streamName(streamName)
+                .streamARNOptional(StreamARNUtil.getStreamARN(streamName, kinesisRegion))
+                .build();
     }
 }
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/KinesisShardDetector.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/KinesisShardDetector.java
@@ -222,11 +222,12 @@ public class KinesisShardDetector implements ShardDetector {
             final boolean shouldPropagateResourceNotFoundException) {
         ListShardsRequest.Builder builder = KinesisRequestsBuilder.listShardsRequestBuilder();
         if (StringUtils.isEmpty(nextToken)) {
-            builder = builder.streamName(streamIdentifier.streamName()).shardFilter(shardFilter);
-        } else {
-            builder = builder.nextToken(nextToken);
-        }
+            builder.streamName(streamIdentifier.streamName()).shardFilter(shardFilter);
+            streamIdentifier.streamARNOptional().ifPresent(arn -> builder.streamARN(arn.toString()));
 
+        } else {
+            builder.nextToken(nextToken);
+        }
         final ListShardsRequest request = builder.build();
         log.info("Stream {}: listing shards with list shards request {}", streamIdentifier, request);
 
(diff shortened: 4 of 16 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamARNUtil.java
```diff
@@ -0,0 +1,90 @@
...
+                .service("kinesis")
+                .region(kinesisRegion)
+                .accountId(stsCallerArn.get().accountId().get())
+                .resource("stream/" + streamName)
+                .build();
+        return Optional.of(kinesisStreamArn);
+    }
+
+    private static boolean isStreamIdentifierConsistentWithARN(StreamIdentifier streamIdentifier, Optional<Arn> streamArn) {
+        if (streamIdentifier.accountIdOptional().isPresent() && streamArn.isPresent()) {
```
shanmsac, line 83: nit: we don't need to check `streamArn.isPresent()` here. This function is called in line 44 because streamArn is present.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamARNUtil.java
```diff
@@ -0,0 +1,90 @@
...
+            String accountIdProvided = streamIdentifier.accountIdOptional().get();
+            return StringUtils.equals(accountIdProvided, streamArn.get().accountId().get());
+        }
+        return true;
```
shanmsac, line 87: Why return true when we are not able to verify it is consistent as streamIdentifier is not present?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/KinesisShardDetector.java
```diff
@@ -308,11 +308,12 @@ public ListShardsResponse getListShardsResponse(ListShardsRequest request) throw
 
     @Override
     public List<ChildShard> getChildShards(final String shardId) throws InterruptedException, ExecutionException, TimeoutException {
-        final GetShardIteratorRequest getShardIteratorRequest = KinesisRequestsBuilder.getShardIteratorRequestBuilder()
+        GetShardIteratorRequest.Builder requestBuilder = KinesisRequestsBuilder.getShardIteratorRequestBuilder()
```
shanmsac, line 311: nit: this can still be `final` right?

stair-aws (changes requested):

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/ConfigsBuilder.java
```diff
@@ -183,7 +189,7 @@ public void appStreamTracker(Either<MultiStreamTracker, String> appStreamTracker
     }
 
     public void streamTracker(StreamTracker streamTracker) {
-        this.streamTracker = streamTracker;
+        this.streamTracker = StreamARNUtil.trySetStreamARN(streamTracker, this.kinesisRegion);
```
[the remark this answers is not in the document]
zengyu714: Yeah this is what I'm thinking: The StreamTracker can be passed by a customer without a StreamARN set, but even in this situation, we still try to set the StreamARN for the customer internally. So I'm not quite sure how to achieve the Factory mode for this scenario. However, I do understand your point about maintaining the immutability of this object. I think it may only be possible when we expose the StreamARN to the customer

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamARNUtil.java
```diff
@@ -0,0 +1,90 @@
...
+     *                     constructed based on the provided streamName
+     * @param kinesisRegion: kinesisRegion is required to construct the stream arn
+     * @return
+     */
+    public static StreamTracker trySetStreamARN(StreamTracker streamTracker, String kinesisRegion) {
+        StsClient sts = StsClient.builder().httpClient(UrlConnectionHttpClient.builder().build()).build();
+        return trySetStreamARN(sts, streamTracker, kinesisRegion);
+    }
+
+    @VisibleForTesting
```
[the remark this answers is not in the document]
zengyu714: This annotation is only for documentation/annotation purposes, so we can remove it without any impact on the code and achieve the same result with existing package-level access

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamARNUtil.java
zengyu714: The StreamIdentifier contains three fields initially, namely streamName, accountIdOptional, and streamCreationEpochOptional. Out of these fields, only the accountIdOptional can be inconsistent with StreamARN (since we get the accountId from STS). If accountIdOptional is not present, we assume that the accountId is correct.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamIdentifier.java
```diff
@@ -27,22 +28,23 @@
 
 @EqualsAndHashCode @Getter @Accessors(fluent = true)
 public class StreamIdentifier {
-    private final Optional<String> accountIdOptional;
     private final String streamName;
+    private final Optional<String> accountIdOptional;
     private final Optional<Long> streamCreationEpochOptional;
+    @Setter private Optional<String> streamARNOptional = Optional.empty();
```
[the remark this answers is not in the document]
zengyu714: Yes, that makes sense, but fortunately we only invoke the setter during initialization, i.e. ConfigsBuilder and RetrievalConfig.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/RetrievalConfig.java
```diff
@@ -80,6 +81,11 @@ public class RetrievalConfig {
      */
     private StreamTracker streamTracker;
 
+    /**
+     * AWS region that the Kinesis stream is in, e.g., us-west-2
+     */
+    private String kinesisRegion;
```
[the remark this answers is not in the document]
zengyu714: This change just follows the current KCL initialization logic which allows the customer to set the configs separately for ConfigsBuilder and RetrievalConfigs... as you can see from the individual initialization of the streamTracker. I agree that we can eliminate this duplication, but I am afraid that is beyond the scope of this change.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/SchedulerTest.java
```diff
@@ -812,7 +816,7 @@ public void testKinesisStaleDeletedStreamCleanup() throws ProvisionedThroughputE
         assertEquals(1, syncedStreams2.size());
         assertEquals(0, scheduler.deletedStreamListProvider().purgeAllDeletedStream().size());
 
-        verify(multiStreamTracker, times(3)).streamConfigList();
+        verify(multiStreamTracker, times(4)).streamConfigList();
```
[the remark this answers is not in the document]
zengyu714: Only one additional iteration during initialization

amazon-kinesis-client/src/test/java/software/amazon/kinesis/retrieval/RetrievalConfigTest.java
```diff
@@ -30,6 +33,8 @@ public class RetrievalConfigTest {
     @Before
     public void setUp() {
         MockitoAnnotations.initMocks(this);
+        when(mockKinesisClient.serviceClientConfiguration())
```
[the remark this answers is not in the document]
zengyu714: Hmm... I am not sure how I can integrate this mock setup into the code...

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamARNUtilTest.java
```diff
@@ -0,0 +1,86 @@
...
+        // the result obtained from sts
+        when(stsClientMock.getCallerIdentity())
+                .thenReturn(GetCallerIdentityResponse.builder().arn(String.format(STS_RESPONSE_ARN_FORMAT, "aws", "111111111111")).build());
+        streamTracker = StreamARNUtil.trySetStreamARN(stsClientMock,
+                new SingleStreamTracker(StreamIdentifier.multiStreamInstance("222222222222:streamName:1000000000")), "region");
+        Assert.assertFalse(streamARNExtractor(streamTracker).get(0).isPresent());
+        reset(stsClientMock);
+    }
+
+    private List<Optional<String>> streamARNExtractor(StreamTracker streamTracker) {
```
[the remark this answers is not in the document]
zengyu714: For StreamARNUtil unit tests, we only test against a single stream tracker, so this method is essentially a helper function that extracts the StreamIdentifier from the StreamTracker object.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamIdentifier.java
```diff
@@ -16,38 +16,37 @@
...
 public class StreamIdentifier {
-    private final Optional<String> accountIdOptional;
+    @Builder.Default
+    private final Optional<String> accountIdOptional = Optional.empty();
     private final String streamName;
-    private final Optional<Long> streamCreationEpochOptional;
+    @Builder.Default
+    private final Optional<Long> streamCreationEpochOptional = Optional.empty();
+    @Builder.Default
+    private Optional<Arn> streamARNOptional = Optional.empty();
```
stair-aws, line 42: +`final` ?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamARNUtilTest.java
```diff
@@ -0,0 +1,70 @@
...
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({ StreamARNUtil.class })
+public class StreamARNUtilTest {
+    private static final String STS_RESPONSE_ARN_FORMAT = "arn:aws:sts::%s:assumed-role/Admin/alias";
+    private static final String KINESIS_STREAM_ARN_FORMAT = "arn:aws:kinesis:us-east-1:%s:stream/%s";
+
+    @Before
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
+        PowerMockito.spy(StreamARNUtil.class);
```
stair-aws, line 28: I recommend intercepting the `StsClient` constructor so you can return a mock `StsClient` object. This will "simplify" the interactions so your tests can validate expectations on a mock instead of a `private static` method. (see additional comments in this class for why that's important)

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamARNUtilTest.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Test
+    public void testGetStreamARNHappyCase() throws Exception {
+        String streamName = "testStreamName";
+        String accountId = "123456789012";
+        PowerMockito.doReturn(Optional.of(Arn.fromString(String.format(STS_RESPONSE_ARN_FORMAT, accountId))))
```
stair-aws, line 35: Unfortunately, this avoids testing the `getStsCallerArn()` method and, instead, assumes the method is correct. An error in the method will go unnoticed until run-time (or exposed by another test). Considering that said method is a part of `StreamARNUtil`, the class directly under test, these tests are the place to prove said method works.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamARNUtilTest.java
```diff
@@ -0,0 +1,70 @@
...
+                .when(StreamARNUtil.class, "getStsCallerArn");
+
+        Optional<Arn> actualStreamARNOptional = StreamARNUtil.getStreamARN(streamName, Region.US_EAST_1);
+        String expectedStreamARN = String.format(KINESIS_STREAM_ARN_FORMAT, accountId, streamName);
+        assertTrue(actualStreamARNOptional.isPresent());
+        assertEquals(expectedStreamARN, actualStreamARNOptional.get().toString());
```
stair-aws, line 41: If you change to use a mock `StsClient` object, then you should validate expectations on that object (in this test, and all other tests). Example:
```
Mockito.verify(mockStsClient).getCallerIdentity();
```

Validating the mock helps ensure any `when(...)` expectations were triggered, and reduces the likelihood that the test passes due to some other fortunate happenstance.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamARNUtilTest.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Test
+    public void testGetStreamARNReturnsEmptyOnSTSError() throws Exception {
```
stair-aws, line 45: Please include a test where `mockStsClient.getCallerIdentity()` throws one of its possible Exceptions.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamARNUtilTest.java
```diff
@@ -0,0 +1,70 @@
...
+        // Optional.empty() is expected when there is an error with the STS call and STS returns empty Arn
+        PowerMockito.doReturn(Optional.empty()).when(StreamARNUtil.class, "getStsCallerArn");
+        Optional<Arn> actualStreamARNOptional = StreamARNUtil.getStreamARN("testStreamName", Region.US_EAST_1);
+        assertFalse(actualStreamARNOptional.isPresent());
+    }
+
+    @Test
+    public void testGetStreamARNReturnsEmptyOnInvalidKinesisRegion() {
```
stair-aws, line 53: If using `mockStsClient`, the verification here should be:
```
Mockito.verify(mockStsClient, never()).getCallerIdentity();
```
... or:
```
Mockito.verifyZeroInteractions(mockStsClient);
```

That is, the short-circuit was successful and the mock object wasn't invoked.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamARNUtilTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test
+    public void testGetStreamARNWithProvidedAccountIDAndIgnoredSTSResult() throws Exception {
+        // If the account id is provided in the StreamIdentifier, it will override the result (account id) returned by sts
+        PowerMockito.doReturn(Optional.of(Arn.fromString(String.format(STS_RESPONSE_ARN_FORMAT, "111111111111"))))
+                .when(StreamARNUtil.class, "getStsCallerArn");
+        Optional<Arn> actualStreamARNOptional = StreamARNUtil.getStreamARN(
+                "testStreamName", Region.US_EAST_1, Optional.of("222222222222"));
+        assertTrue(actualStreamARNOptional.isPresent());
+        assertTrue(actualStreamARNOptional.get().toString().contains("222222222222"));
```
stair-aws, line 67: Would you kindly restructure this as an `assertEquals(<expected>, <actual>)`? If/When this assertion fails, all the engineer will know is that it wasn't `true`. `assertEquals(...)` provides a comparison which greatly facilitates debugging.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/MockObjectHelper.java
```diff
@@ -0,0 +1,27 @@
+package software.amazon.kinesis.utils;
+
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.kinesis.KinesisAsyncClient;
+import software.amazon.awssdk.services.kinesis.KinesisServiceClientConfiguration;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+final public class MockObjectHelper {
```
stair-aws, line 10: Nit: `s/final public/public final/`

stair-aws (changes requested):

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamIdentifier.java
```diff
@@ -16,38 +16,37 @@
...
-        this.streamName = streamName;
-        this.streamCreationEpochOptional = Optional.of(streamCreationEpoch);
-    }
-
-    private StreamIdentifier(@NonNull String streamName) {
-        this.accountIdOptional = Optional.empty();
-        this.streamName = streamName;
-        this.streamCreationEpochOptional = Optional.empty();
-    }
+    private static final Pattern PATTERN = Pattern.compile(".*" + ":" + ".*" + ":" + "[0-9]*" + "([a-z]{2}(-gov)?-[a-z]+-\\d{1})?");
```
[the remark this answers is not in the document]
zengyu714: > Also, this pattern is most likely incorrect. All * (i.e., zero-or-more) should likely be + (i.e., one-or-more). Not your responsibility to change this, but mentioning it nonetheless.

Agreed, I am quite hesitant to modify the existing patterns

> Should there be a : before the new region token? (This might be a reason to do the String.join.)

Not really, the `*` preceding the region matches the `:` used as a delimiter for the region.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamARNUtilTest.java
zengyu714: Unfortunately, it seems that intercepting StsClient using PowerMock.whenNew is not working for some reason. One possibility is to wrap the StsClient and its getCallerIdentity method in a separate class

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Test
+    public void testSingleStreamInstanceWithNameAndRegion() {
+        Optional<Arn> arn = Optional.of(Arn.builder().partition("aws").service("kinesis")
+                .region(kinesisRegion.toString()).accountId("123").resource("stream/" + streamName).build());
+        mockStatic(StreamARNUtil.class);
+        when(StreamARNUtil.getStreamARN(eq(streamName), eq(kinesisRegion))).thenReturn(arn);
+        StreamIdentifier actualStreamIdentifier = StreamIdentifier.singleStreamInstance(streamName, kinesisRegion);
+        Assert.assertFalse(actualStreamIdentifier.streamCreationEpochOptional().isPresent());
+        Assert.assertFalse(actualStreamIdentifier.accountIdOptional().isPresent());
```
stair-aws, line 45: Is this correct? When a provided region yields an Arn, should we set the account id to match that from the Arn (if it was going to be otherwise empty)?

Anotherwards, can/should this be `assertEquals(Optional.of("123"), aSI.accountIdOptional())` ?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -0,0 +1,94 @@
...
+        Assert.assertTrue(actualStreamIdentifier.streamARNOptional().isPresent());
```
stair-aws, line 46: Replace w/ `assertEquals(arn, actualStreamIdentifier.streamARNOptional())` ?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Test
+    public void testMultiStreamInstanceWithIdentifierSerialization() {
+        String epoch = "1680616058";
+        Optional<Arn> arn = Optional.ofNullable(Arn.builder().partition("aws").service("kinesis")
+                .accountId(accountId).region(kinesisRegion.toString()).resource("stream/" + streamName).build());
+
+        mockStatic(StreamARNUtil.class);
+        when(StreamARNUtil.getStreamARN(eq(streamName), any(), any())).thenReturn(arn);
```
stair-aws, line 57: `any()` sometimes obscures problems and, subjectively, makes it harder for humans to read the code (because there is no hint as to what is expected). Use more explicit `any(Region.class), any(Optional.class)`

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -0,0 +1,94 @@
+package software.amazon.kinesis.common;
+
+import com.google.common.base.Joiner;
```
stair-aws, line 3: Similar callout as earlier w.r.t. Joiner.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+        mockStatic(StreamARNUtil.class);
+        when(StreamARNUtil.getStreamARN(eq(streamName), eq(kinesisRegion), any())).thenReturn(arn);
+        StreamIdentifier actualStreamIdentifier = StreamIdentifier.multiStreamInstance(
+                Joiner.on(":").join(accountId, streamName, epoch), kinesisRegion);
+        assertActualStreamIdentifierExpected(arn, actualStreamIdentifier);
+    }
+
+    private void assertActualStreamIdentifierExpected(Optional<Arn> expectedARN, StreamIdentifier actual) {
+        Assert.assertTrue(actual.streamCreationEpochOptional().isPresent());
```
stair-aws, line 89: How many of these lossy `assertTrue` can you replace w/ stronger expectations? (e.g.,  `assertEquals(Optional.of(EPOCH), actual.sCEO)` ?

Validating _exact_ values yields better tests. These assertions would not flag incorrect results so long as they're non-empty.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/MockObjectHelper.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public static KinesisAsyncClient createKinesisClient() {
+        return createKinesisClient(Region.US_EAST_1);
+    }
+
+    /**
+     * @param isRegionDummy a boolean to determine whether to use a null value for the Kinesis client's region.
+     * @return
+     */
+    public static KinesisAsyncClient createKinesisClient(boolean isRegionDummy) {
```
stair-aws, line 20: DRY+KISS: this method is redundant with the calls in the ternary. This doesn't simplify anything, and introduces more code we must maintain.

stair-aws (changes requested):
Many comments are nits and other small items. However, there is a memory leak, the performance impact of avoidable calls to `STS.getCallerIdentity()`, and some unit tests which need a touch-up.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamARNUtil.java
```diff
@@ -0,0 +1,86 @@
...
+import software.amazon.awssdk.services.sts.StsClient;
+import software.amazon.awssdk.services.sts.model.GetCallerIdentityResponse;
+
+import java.util.HashMap;
+import java.util.Optional;
+
+@Slf4j
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class StreamARNUtil {
+    private static final HashMap<String, Arn> streamARNCache = new HashMap<>();
```
stair-aws, line 23: FYI: I have submitted https://github.com/awslabs/amazon-kinesis-client/pull/1096 to provide a `SupplierCache`. This may be leveraged here to provide a singular `Arn` cache. Example:

```
private static final SupplierCache<Arn> CACHED_ARN = new SupplierCache<>(() -> {
        try {
            final GetCallerIdentityResponse response = getStsClient().getCallerIdentity();
            return Arn.fromString(response.arn());
        } catch (AwsServiceException | SdkClientException e) {
            log.warn("Unable to get sts caller identity to build stream arn", e);
            return null;
        }
    });
```

stair-aws (approved):
Approving as-is w/ awareness that this PR introduces a memory leak and has unaddressed concerns. Will publish a fast-follow PR in a few minutes that addresses the known issues.

PR #1088, opened 2023-04-10
Minimize race in PSSM to avoid unnecessary shard sync calls 

chenylee-aws:
*Description of changes:*
There exists a race condition between `PeriodicShardSyncManager (PSSM)` and `Scheduler` as reads and writes can happen concurrently in the respective threads. This would cause unnecessary shard syncs for example when `PSSM` is initiated at the same time as the `Scheduler` the streamToLease map is filtered based on the previous snapshot of the `currentStreamConfigMap` which tricks the `PSSM` into believing that there is no leases for the newly added streams that already have leases. 

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/PeriodicShardSyncManager.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/PeriodicShardSyncManager.java
@@ -178,12 +178,20 @@ class PeriodicShardSyncManager {
             final long runStartMillis = System.currentTimeMillis();
 
             try {
+                // Create a shallow copy of the map to avoid data race with Scheduler.
+                final Map<StreamIdentifier, StreamConfig> streamConfigMap = currentStreamConfigMap.entrySet().stream()
+                        .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+
                 // Construct the stream to leases map to be used in the lease sync
                 final Map<StreamIdentifier, List<Lease>> streamToLeasesMap = getStreamToLeasesMap(
-                        currentStreamConfigMap.keySet());
+                        streamConfigMap.keySet());
 
                 // For each of the stream, check if shard sync needs to be done based on the leases state.
-                for (Map.Entry<StreamIdentifier, StreamConfig> streamConfigEntry : currentStreamConfigMap.entrySet()) {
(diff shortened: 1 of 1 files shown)


furq-aws (approved):

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/PeriodicShardSyncManager.java
```diff
@@ -178,12 +178,20 @@ private void runShardSync() {
...
                 // Construct the stream to leases map to be used in the lease sync
                 final Map<StreamIdentifier, List<Lease>> streamToLeasesMap = getStreamToLeasesMap(
-                        currentStreamConfigMap.keySet());
+                        streamConfigMap.keySet());
 
                 // For each of the stream, check if shard sync needs to be done based on the leases state.
-                for (Map.Entry<StreamIdentifier, StreamConfig> streamConfigEntry : currentStreamConfigMap.entrySet()) {
+                for (Map.Entry<StreamIdentifier, StreamConfig> streamConfigEntry : streamConfigMap.entrySet()) {
+                    // Skip shardSync if CSCM has purged the stream.
+                    if (!currentStreamConfigMap.containsKey(streamConfigEntry.getKey())) {
```
stair-aws, line 192: +1 for short-circuit in scenario where `currentStreamConfigMap` is no longer aware of the stream.

However, to reduce the wall-time between "check" and "execute", WDYT about moving this short-circuit to approximately line ~203 just-before `shardSyncTaskManagerProvider.apply(...)`. This may yield the highest rate of short-circuits which avoid a network call.

Example:
```
log.info("Periodic shard ...");
final StreamConfig streamConfig = currentStreamConfigMap.get(streamConfigEntry.getKey());
if (streamConfig == null) {
   // Skip shardSync if CSCM has purged the stream.
   continue;
}
final ShardSyncTaskManager sSTM = sSTMP.apply(streamConfig);
// rest of block is unchanged
```

stair-aws (approved):
Bias-for-action approve since this is better than what-is. Please consider the comment for a potentially even-better solution.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/PeriodicShardSyncManager.java
chenylee-aws: That's a good callout. I think we can keep the current one and add this new check to optimize both the CPU time and network bandwidth and to achieve higher consistency.

stair-aws (approved):

PR #1094, opened 2023-04-18
Code cleanup to introduce better testing and simplify future removal of

stair-aws:
*Issue #, if available:*
N/A

*Description of changes:*
Code cleanup to introduce better testing and simplify future removal of deprecated parameters (e.g., `Either<L, R> appStreamTracker`).

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/RetrievalConfig.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/RetrievalConfig.java
@@ -133,6 +133,8 @@ public class RetrievalConfig {
     }
 
     /**
+     * Convenience method to reconfigure the embedded {@link StreamTracker},
+     * but only when <b>not</b> in multi-stream mode.
      *
      * @param initialPositionInStreamExtended
      *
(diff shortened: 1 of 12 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/RetrievalConfig.java
```diff
@@ -142,62 +144,46 @@ public RetrievalConfig(@NonNull KinesisAsyncClient kinesisAsyncClient, @NonNull
...
-        if(isInvalidFanoutConfig) {
-            throw new IllegalArgumentException(
-                    "Invalid config: Either in multi-stream mode with streamName/consumerArn configured or in single-stream mode with no streamName configured");
-        }
+    /**
+     * Convenience method to return the {@link StreamIdentifier} from a
+     * single-stream tracker.
+     */
+    private StreamIdentifier getSingleStreamIdentifier() {
+        return streamTracker.streamConfigList().get(0).streamIdentifier();
```
noahbt, line 186: Can `get(0)` return null in this line?

noahbt (approved):
Refactors look good and the added tests should help strengthen the code base

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/RetrievalConfig.java
stair-aws: Hypothetically, yes: a customer could subclass a `StreamTracker` that is not multi-stream, not a [`SingleStreamTracker`](https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/main/java/software/amazon/kinesis/processor/SingleStreamTracker.java#L67-L70) (which protects against this via `@NonNull`), and is capable of tracking an empty list.

However, at that point they'd be using KCL in single-stream mode to configure record retrieval for zero (0) streams. While hypothetical, that configuration is not productive.

PR #1098, opened 2023-04-19
StreamARN: enhanced `StreamIdentifier` to spawn instances from stream…

stair-aws:
*Issue #, if available:*
N/A

*Description of changes:*
StreamARN: enhanced `StreamIdentifier` to spawn instances from stream ARNs.

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/KinesisClientFacade.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/KinesisClientFacade.java
@@ -1,0 +1,103 @@
+package software.amazon.kinesis.retrieval;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+import software.amazon.awssdk.services.kinesis.KinesisAsyncClient;
+import software.amazon.awssdk.services.kinesis.model.DescribeStreamSummaryRequest;
+import software.amazon.awssdk.services.kinesis.model.DescribeStreamSummaryResponse;
+import software.amazon.awssdk.services.kinesis.model.KinesisException;
+import software.amazon.awssdk.services.kinesis.model.LimitExceededException;
+import software.amazon.awssdk.services.kinesis.model.ResourceInUseException;
+import software.amazon.awssdk.services.kinesis.model.ResourceNotFoundException;
+import software.amazon.kinesis.common.KinesisRequestsBuilder;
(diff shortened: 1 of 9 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/KinesisClientFacade.java
```diff
@@ -0,0 +1,103 @@
...
+            final int maxRetries,
+            final String streamArn,
+            @NonNull final String apiName) {
+        LimitExceededException finalException = null;
+
+        int retries = maxRetries;
+        while (retries > 0) {
+            try {
+                try {
+                    return retriever.get();
```
akidambisrinivasan, line 71: Dont we need to get a new future by calling kinesisClient.describeStreamSummary before retrying?

Also dont understand whats the value of nested try-catch blocks?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/retrieval/KinesisClientFacadeTest.java
```diff
@@ -0,0 +1,67 @@
...
+    @Test
+    public void testDescribeStreamSummaryRetries() throws Exception {
+        final DescribeStreamSummaryResponse expectedResponse = DescribeStreamSummaryResponse.builder().build();
+        final CompletableFuture<DescribeStreamSummaryResponse> mockFuture = mock(CompletableFuture.class);
+        final TimeoutException timeoutException = new TimeoutException();
+
+        when(mockKinesisClient.describeStreamSummary(any(DescribeStreamSummaryRequest.class)))
+                .thenReturn(mockFuture);
+        when(mockFuture.get(anyInt(), any(TimeUnit.class)))
+                .thenThrow(timeoutException)
```
akidambisrinivasan, line 57: Is this valid? A future once completed why will it return different exceptions? I think we need the API call to be made again for retries which would result in different futures.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/KinesisClientFacade.java
stair-aws: Great catch w.r.t. the isolated `future = ...`! This would, indeed, reuse the singular `Future` which may have completed exceptionally.

---
> Also dont understand whats the value of nested try-catch blocks?

The inner `try` allows for Exception substitution (e.g., catch-repackage-throw, unwrap-and-throw). You'll often see an [`AWSExceptionManager`](https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/AWSExceptionManager.java) in similar nested-`try`.

akidambisrinivasan (approved):

PR #1109, opened 2023-05-16
Add support for stream ARNs

furq-aws:
*Issue #, if available:*
N/A

*Description of changes:*
Adds support for referencing streams by `streamARN` in single-stream mode, or by the combination of `streamARN` and `creationEpoch` in multi-stream mode.


amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamIdentifier.java
```diff
@@ -101,129 +84,82 @@ public String toString() {
...
-     *
-     * @see #singleStreamInstance(String, Region)
+     * @param streamARN
+     * @param creationEpoch
+     * @return StreamIdentifier
      */
-    public static StreamIdentifier singleStreamInstance(String streamNameOrArn) {
-        return singleStreamInstance(streamNameOrArn, null);
+    public static StreamIdentifier multiStreamInstance(Arn streamARN, Long creationEpoch) {
+        validateARN(streamARN);
```
stair-aws, line 119: 1. It might benefit to validate `creationEpoch > 0`. Negative epochs make no sense.
1. This is a trade-off that allows customers to self-harm: if they don't provide the correct epoch, and KCL doesn't validate it against the _actual_ stream, customers might recycle the epoch for two editions of the same stream which might present odd/deleterious behavior in lease management. (e.g., create stream `A` at epoch `t1`, create `StreamIdentifer(A, t1)`, delete `A`, create stream `A` at epoch `t2`, reuse old epoch to create `StreamIdentifier(A, t1)`)

... or, potentially, a customer decides the epoch correctness doesn't matter and they provide `0` (or 1, or 42, or ...).

At the least, **this expected behavior must be very clearly detailed in the Javadoc**.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -51,13 +40,6 @@ public void testMultiStreamDeserializationSuccess() {
...
     public void testMultiStreamDeserializationFail() {
         for (final String pattern : Arrays.asList(
-                // arn examples
-                "arn:aws:kinesis::123456789012:stream/stream-name", // missing region
-                "arn:aws:kinesis:region::stream/stream-name", // missing account id
-                "arn:aws:kinesis:region:123456789:stream/stream-name", // account id not 12 digits
-                "arn:aws:kinesis:region:123456789abc:stream/stream-name", // 12char alphanumeric account id
-                "arn:aws:kinesis:region:123456789012:stream/", // missing stream-name
-                // serialization examples
                 ":stream-name:123", // missing account id
```
stair-aws, line 43: Please add a pattern that has a _negative_ epoch. We should expect it to not match the pattern (since `-` is not valid).

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/KinesisClientFacade.java
```diff
@@ -1,114 +0,0 @@
...
-import software.amazon.awssdk.services.kinesis.model.LimitExceededException;
-import software.amazon.awssdk.services.kinesis.model.ResourceInUseException;
-import software.amazon.awssdk.services.kinesis.model.ResourceNotFoundException;
-import software.amazon.kinesis.common.KinesisRequestsBuilder;
-
-/**
- * Facade pattern to simplify interactions with a {@link KinesisAsyncClient}.
- */
-@Slf4j
-public final class KinesisClientFacade {
```
stair-aws, line 35: A facade is an improvement over the scattered ownership, but we can revisit this holistically in a future enhancement.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/ConfigsBuilderTest.java
```diff
@@ -53,12 +54,27 @@ public class ConfigsBuilderTest {
...
+        testTrackerConstruction(streamName, configBySingleTracker, configByName);
+    }
+
+    @Test
+    public void testTrackerConstructionByStreamARN() {
+        final Arn streamARN = Arn.fromString("arn:aws:kinesis:region:123456789012:stream/stream-name");
+        final ConfigsBuilder configByArn = createConfig(streamARN);
+        final ConfigsBuilder configBySingleTracker = createConfig(new SingleStreamTracker(streamARN));
+
+        final String streamName = streamARN.resource().resource();
```
stair-aws, line 71: Suggestion: for test legibility, include a `assertEquals("stream-name", streamName);`

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/ConfigsBuilderTest.java
```diff
@@ -53,12 +54,27 @@ public class ConfigsBuilderTest {
...
+        testTrackerConstruction(streamName, configBySingleTracker, configByArn);
+    }
+
+    private void testTrackerConstruction(String streamName,
+                                         ConfigsBuilder expectedConfig, ConfigsBuilder actualConfig) {
```
stair-aws, line 76: Style: don't do horizontal alignment. This increases blast radius of edits, increases risk of merge conflicts, and provides zero objective benefit.

If your IDE is doing this on its own, a quick workaround is to CRLF before the first param. It should then be properly indented (roughly two indentation levels more than the `private`), and CRLF for every other param will follow that same indentation. Example:
```
private void ...( // observe: first param is on next line
      String streamName,
      ConfigsBuilder ...,
      ConfigsBuilder ...) {
```

---

FYI: you could also KISS and use varargs: `testTrackerConstruction(String streamName, ConfigsBuilder... singleStreamConfig) {`

Recommend refactoring along aspect of single- vs multi-stream, tho. (That seems better to me, and avoids redundancy.)

stair-aws, line 76: `expectedConfig` is a misnomer. `expected` generally signals "this will be a left-hand param during `assertXXX(...)`.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/ConfigsBuilderTest.java
```diff
@@ -53,12 +54,27 @@ public class ConfigsBuilderTest {
...
 
     @Test
-    public void testTrackerConstruction() {
+    public void testTrackerConstructionByStreamName() {
         final String streamName = "single-stream";
         final ConfigsBuilder configByName = createConfig(streamName);
         final ConfigsBuilder configBySingleTracker = createConfig(new SingleStreamTracker(streamName));
 
-        for (final ConfigsBuilder cb : Arrays.asList(configByName, configBySingleTracker)) {
+        testTrackerConstruction(streamName, configBySingleTracker, configByName);
```
stair-aws, line 62: This method tests multi-stream construction twice which is both redundant, and will have a higher blast-radius if multi-stream fails.

Consider splitting this into two test methods: single-stream and multi-stream. Single-stream may look something like:
```
for (final ConfigsBuilder cb : Arrays.asList(
    createConfig(streamName),
    createConfig(new SST(streamName)),
    createConfig(streamArn),
    createConfig(new SST(streamArn))) {
   // do the single-stream tests here
}
```

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -76,47 +58,58 @@ public void testMultiStreamDeserializationFail() {
...
+    public void testMultiStreamArnFail() {
+        for (final String pattern : Arrays.asList(
+                "arn:aws:kinesis::123456789012:stream/stream-name", // missing region
+                "arn:aws:kinesis:region::stream/stream-name", // missing account id
+                "arn:aws:kinesis:region:123456789:stream/stream-name", // account id not 12 digits
+                "arn:aws:kinesis:region:123456789abc:stream/stream-name", // 12char alphanumeric account id
+                "arn:aws:kinesis:region:123456789012:stream/" // missing stream-name
+        )) {
+            try {
+                StreamIdentifier.multiStreamInstance(Arn.fromString(pattern), EPOCH);
```
stair-aws, line 74: Please move `Arn.fromString(...)` before the `try`. We should not allow an IAE thrown from Arn -- if it even does, now or in the future -- to create false assurance for this test.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -76,47 +58,58 @@ public void testMultiStreamDeserializationFail() {
...
-                .thenReturn(Optional.empty());
-
-        try {
-            StreamIdentifier.singleStreamInstance(DEFAULT_ARN.toString());
-        } finally {
-            verifyStatic(StreamARNUtil.class);
-            getStreamARN(STREAM_NAME, KINESIS_REGION, TEST_ACCOUNT_ID);
+    public void testMultiStreamArnFail() {
+        for (final String pattern : Arrays.asList(
+                "arn:aws:kinesis::123456789012:stream/stream-name", // missing region
```
stair-aws, line 67: Include a valid Arn that isn't for a Kinesis stream.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -76,47 +58,58 @@ public void testMultiStreamDeserializationFail() {
...
-        mockStatic(StreamARNUtil.class);
-        when(getStreamARN(STREAM_NAME, KINESIS_REGION, TEST_ACCOUNT_ID))
-                .thenReturn(Optional.empty());
-
-        try {
-            StreamIdentifier.singleStreamInstance(DEFAULT_ARN.toString());
-        } finally {
-            verifyStatic(StreamARNUtil.class);
-            getStreamARN(STREAM_NAME, KINESIS_REGION, TEST_ACCOUNT_ID);
+    public void testMultiStreamArnFail() {
```
stair-aws, line 65: Include a test that attempts to construct multi-stream identifiers w/ invalid epochs (e.g., `null` [if not prevented by primitive], `<0`, `=0`).

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -76,47 +58,58 @@ public void testMultiStreamDeserializationFail() {
...
+        assertEquals(actualStreamIdentifier.streamName(), actualStreamIdentifier.serialize());
+    }
+
+    @Test
+    public void testMultiStreamInstanceFromArn() {
+        final StreamIdentifier actualStreamIdentifier = StreamIdentifier.multiStreamInstance(DEFAULT_ARN, EPOCH);
+
+        assertTrue(actualStreamIdentifier.streamARNOptional().isPresent());
+        assertTrue(actualStreamIdentifier.accountIdOptional().isPresent());
+        assertTrue(actualStreamIdentifier.streamCreationEpochOptional().isPresent());
```
stair-aws, line 100: Replace this epoch `assertTrue` with `assertEquals(EPOCH, ...)`. That is, don't just validate existence; validate the correct value.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -76,47 +58,58 @@ public void testMultiStreamDeserializationFail() {
...
     }
 
     @Test
-    public void testSingleStreamInstanceWithName() {
-        StreamIdentifier actualStreamIdentifier = StreamIdentifier.singleStreamInstance(STREAM_NAME);
+    public void testSingleStreamInstanceFromArn() {
+        final StreamIdentifier actualStreamIdentifier = StreamIdentifier.singleStreamInstance(DEFAULT_ARN);
+
+        assertTrue(actualStreamIdentifier.streamARNOptional().isPresent());
+        assertTrue(actualStreamIdentifier.accountIdOptional().isPresent());
```
stair-aws, line 87: DRY: these are tested -- and better-tested using equality, not existence -- in `assertActualStreamIdentifierExpected`.

edit: ditto for lines 98~99 in `testMultiStreamInstanceFromArn`

stair-aws (changes requested):

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/ConfigsBuilder.java
```diff
@@ -142,7 +143,31 @@ public ConfigsBuilder(@NonNull String streamName, @NonNull String applicationNam
...
+     *
+     * @param streamARN
+     * @param applicationName
+     * @param kinesisClient
+     * @param dynamoDBClient
+     * @param cloudWatchClient
+     * @param workerIdentifier
+     * @param shardRecordProcessorFactory
+     */
+    public ConfigsBuilder(@NonNull Arn streamARN, @NonNull String applicationName,
```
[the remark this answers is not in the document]
furq-aws: Personally prefer `streamARN`, but noticed `consumerArn` variables already exist in the KCL, so converted `ARN` to `Arn` for now.

kcl-2x-caa-partial-solution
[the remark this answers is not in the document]
furq-aws: I believe this may have been created when previously setting up a new repository through GitHub Desktop. Deleted.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/RetrievalConfig.java
```diff
@@ -119,9 +119,7 @@ public class RetrievalConfig {
 
     public RetrievalConfig(@NonNull KinesisAsyncClient kinesisAsyncClient, @NonNull String streamName,
                            @NonNull String applicationName) {
-        this(kinesisAsyncClient,
-                new SingleStreamTracker(streamName, kinesisAsyncClient.serviceClientConfiguration().region()),
-                applicationName);
+        this(kinesisAsyncClient, new SingleStreamTracker(streamName), applicationName);
```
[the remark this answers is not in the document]
furq-aws: Added `RetrievalConfig(KinesisAsyncClient, Arn, String)` constructor. Confirmed with @lucienlu-aws that `creationEpoch` would not be an issue because this constructor can only be used for single-stream mode.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
furq-aws: This was indeed creating false assurance for this test. Updated test to focus on Arns that can be successfully constructed with `Arn.Builder()` but are not valid Kinesis stream Arns.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamIdentifier.java
```diff
@@ -64,34 +63,18 @@ public class StreamIdentifier {
...
-            //      Furthermore, if epoch isn't lazy-loaded here, the problem quickly spirals into
-            //      systemic issues of concurrency and consistency (e.g., PeriodicShardSyncManager,
-            //      Scheduler, DDB leases). We should look at leveraging dependency injection.
-            //      (NOTE: not to inject the Kinesis client here, but to ensure the client is
-            //      accessible elsewhere ASAP.)
-            final DescribeStreamSummaryResponse dss = KinesisClientFacade.describeStreamSummary(
-                    streamARNOptional().get().toString());
-            final long creationEpoch = dss.streamDescriptionSummary().streamCreationTimestamp().getEpochSecond();
-            streamCreationEpochOptional = Optional.of(creationEpoch);
+            return streamName;
```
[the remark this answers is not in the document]
furq-aws: Added. 
Nit clarification: `if` conditional is expected to be true for both single-stream by-name case and single-stream by-Arn case (`creationEpoch` is only needed and available in multi-stream mode).

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/StreamIdentifier.java
```diff
@@ -64,34 +63,18 @@ public class StreamIdentifier {
     /**
      * Serialize the current StreamIdentifier instance.
      *
-     * @return a String of {@code account:stream:creationEpoch[:region]}
-     *      where {@code region} is the id representation of a {@link Region}
-     *      and is optional.
+     * @return a String of {@code account:stream:creationEpoch}.
```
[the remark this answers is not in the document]
furq-aws: Correct, updated javadoc for single-stream case.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -1,39 +1,29 @@
...
+    private static final String RESOURCE = "stream/" + STREAM_NAME;
     private static final long EPOCH = 1680616058L;
-
-    private static final Arn DEFAULT_ARN = toArn(KINESIS_REGION);
-
-    @BeforeClass
-    public static void setUpBeforeClass() throws Exception {
-        StreamARNUtilTest.setUpSupplierCache(new SupplierCache<>(() -> DEFAULT_ARN));
-    }
+    private static final Arn DEFAULT_ARN = Arn.builder()
```
stair-aws, line 21: DRY: can use `streamArn(...)` (which was previously `toArn(...)` but with fewer parameters).

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -51,72 +41,88 @@ public void testMultiStreamDeserializationSuccess() {
...
-    public void testInstanceFromArn() {
-        final Arn arn = toArn(KINESIS_REGION);
-        final StreamIdentifier single = StreamIdentifier.singleStreamInstance(arn.toString());
-        final StreamIdentifier multi = StreamIdentifier.multiStreamInstance(arn.toString());
-
-        assertEquals(single, multi);
-        assertEquals(Optional.empty(), single.streamCreationEpochOptional());
-        assertActualStreamIdentifierExpected(arn, single);
+    public void testMultiStreamByArnWithInvalidStreamArnFail() {
+        final String region = KINESIS_REGION.id();
```
stair-aws, line 68: [Primitive obsession](https://wiki.c2.com/?PrimitiveObsession). Pass a `Region` into `streamArn(...)` and let *it* Stringify (if non-null).

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -51,72 +41,88 @@ public void testMultiStreamDeserializationSuccess() {
...
+        for (final Arn invalidStreamArn : Arrays.asList(
+                streamArn("abc", SERVICE,  region, TEST_ACCOUNT_ID, RESOURCE), // invalid partition
+                streamArn(PARTITION, "dynamodb",  region, TEST_ACCOUNT_ID, RESOURCE), // incorrect service
+                streamArn(PARTITION, SERVICE,  null, TEST_ACCOUNT_ID, RESOURCE), // missing region
+                streamArn(PARTITION, SERVICE,  region, null, RESOURCE), // missing account id
+                streamArn(PARTITION, SERVICE,  region, "123456789", RESOURCE), // account id not 12 digits
+                streamArn(PARTITION, SERVICE,  region, "123456789abc", RESOURCE), // 12char alphanumeric account id
+                streamArn(PARTITION, SERVICE,  region, TEST_ACCOUNT_ID, "table/name"), // incorrect resource type
```
stair-aws, line 76: Would you kindly include the following (if-and-only-if the ARN builder allows):
1. A Kinesis ARN w/ empty resource (i.e., `""`).
1. A Kinesis ARN w/o a stream name (i.e., `stream/`).

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -51,72 +41,88 @@ public void testMultiStreamDeserializationSuccess() {
...
+                Arn.fromString("arn:aws:dynamodb:us-east-2:123456789012:table/myDynamoDBTable") // valid Arn for incorrect resource
```
stair-aws, line 77: FMI: how is this materially different from `streamArn(PARTITION, "dynamodb",  region, TEST_ACCOUNT_ID, RESOURCE)` ?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
```diff
@@ -51,72 +41,88 @@ public void testMultiStreamDeserializationSuccess() {
...
-            StreamIdentifier.singleStreamInstance(DEFAULT_ARN.toString());
-        } finally {
-            verifyStatic(StreamARNUtil.class);
-            getStreamARN(STREAM_NAME, KINESIS_REGION, TEST_ACCOUNT_ID);
+    @Test
+    public void testMultiStreamByArnWithInvalidCreationEpochFail() {
+        final Arn streamArn = streamArn(PARTITION, SERVICE,  KINESIS_REGION.id(), TEST_ACCOUNT_ID, RESOURCE);
+        for (final long invalidCreationEpoch : Arrays.asList(-123, 0)) {
+            try {
+                StreamIdentifier.multiStreamInstance(streamArn, invalidCreationEpoch);
```
stair-aws, line 93: 1. KISS: replace `streamArn` w/ `DEFAULT_ARN`.
2. KISS: eliminate the try+catch+Assert and just have two very simple tests:
```
@Test(expected = IllegalArgumentException.class)
public void testNegativeCreationEpoch() {
   SI.multiSI(DEFAULT_ARN, -123);
}

@Test(expected = IllegalArgumentException.class)
public void testZeroCreationEpoch() {
   SI.multiSI(DEFAULT_ARN, 0);
}
```

FYI: this could also be managed by JUnit parameters (https://junit.org/junit4/javadoc/4.12/org/junit/runners/Parameterized.html).

stair-aws:
LGTM once 1) the Javadoc is, subjectively, made more crisp w.r.t. creation epoch `@param`, and 2) a few more ARN variations in unit test variations are included to help verify correctness. Both should be very quick fixes.

Thanks!

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
lucienlu-aws: This ARN is a valid matching ARN for a different service (a dynamodb table resource for dynamodb). `streamArn(PARTITION, "dynamodb",  region, TEST_ACCOUNT_ID, RESOURCE` is not a valid ARN as there is no `stream` resource in dynamodb

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java
furq-aws: ARN builder does not allow these cases.

lucienlu-aws (approved):

stair-aws (approved):

svroonland:
There is no support for ARN yet in FanoutConfig (enhanced fanout). Was that an oversight or on purpose?

PR #1136, opened 2023-06-09
Adding testing architecture and KCL 2.x basic polling/streaming tests

mmankika-aws:
*Issue #, if available:*

*Description of changes:*
Adding testing architecture to implement automated integration tests. 
Added KCL2XIntegrationTests which implements basic polling and streaming consumer tests against a single producer. 
Integration tests require AWS credentials to be provided. The integration tests can be run with the command 
`mvn -Dit.test=KCLV2IntegrationTest -Dcredentials="<IAM_USER>" verify`


amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/RecordValidatorQueueTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Test
+    public void validationFailedRecordOutOfOrderTest() {
+        recordValidator = new RecordValidatorQueue();
+        recordValidator.add(shardId, "0");
+        recordValidator.add(shardId, "1");
+        recordValidator.add(shardId, "3");
+        recordValidator.add(shardId, "2");
+
+        int error = recordValidator.validateRecords( 4 );
```
ZeyuLi-AWS, line 28: I believe we don't need space between 4 and ( 
Can check with other places for consistency.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLReleaseCanary2XPollingH2TestConfig.java
```diff
@@ -0,0 +1,106 @@
...
+import software.amazon.kinesis.retrieval.polling.PollingConfig;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.util.Date;
+
+public class KCLReleaseCanary2XPollingH2TestConfig implements KCLAppConfig {
```
stair-aws, line 17: DRY. There's a lot of duplication between these subclasses and `KCLAppConfig`. Create a base class that provides default implementations, and only override what needs to be overridden.

`KCLAppConfig` is probably better suited as an abstract class.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/integration_tests/KCL2XIntegrationTest.java
```diff
@@ -0,0 +1,35 @@
...
+@Slf4j
+public class KCL2XIntegrationTest {
+
+    private static final String CONFIG_PACKAGE = "software.amazon.kinesis.config";
+
+    @Test
+    public void KCLReleaseCanary2XPollingH2Test() throws Exception {
+        String[] configName = { "KCLReleaseCanary2XPollingH2TestConfig" };
+        KCLAppConfig consumerConfig = (KCLAppConfig) Class.forName(CONFIG_PACKAGE + "." + configName[0]).newInstance();
+        TestConsumerV2 consumer = new TestConsumerV2( consumerConfig );
```
stair-aws, line 16: KISS. Why not:
```
KCLAppConfig consumerConfig = new KCLRC2XPollingH2TestConfig();
```

amazon-kinesis-client/src/test/java/software/amazon/kinesis/integration_tests/KCL2XIntegrationTest.java
```diff
@@ -0,0 +1,35 @@
+package software.amazon.kinesis.integration_tests;
```
stair-aws, line 1: FYI: convention is to have IT intermixed w/ non-IT tests, and named appropriately. We already have the `pom.xml` configured w/ the expected suffix:
https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/pom.xml#L218-L220

amazon-kinesis-client/src/test/java/software/amazon/kinesis/integration_tests/StreamExistenceManager.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+import static java.lang.Thread.sleep;
+
+@Data
```
stair-aws, line 22: `@Value` ? Value is immutable.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/integration_tests/StreamExistenceManager.java
```diff
@@ -0,0 +1,101 @@
...
+@Slf4j
+public class StreamExistenceManager {
+    private final KinesisAsyncClient client;
+    private final KCLAppConfig testConfig;
+
+    public StreamExistenceManager(KCLAppConfig config) throws URISyntaxException, IOException {
+        this.testConfig = config;
+        this.client = config.buildAsyncKinesisClient(Protocol.HTTP1_1);
```
stair-aws, line 30: This ignores any protocol specified by the `KCLAppConfig`

amazon-kinesis-client/src/test/java/software/amazon/kinesis/integration_tests/StreamExistenceManager.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public static StreamExistenceManager newManager(KCLAppConfig config) throws URISyntaxException, IOException {
```
stair-aws, line 33: DRY. What is the value of this method over using the public constructor?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/integration_tests/StreamExistenceManager.java
```diff
@@ -0,0 +1,101 @@
...
+            }
+        }
+    }
+
+    public void checkStreamAndCreateIfNecessary(String streamName) {
+
+        if (!isStreamActive(streamName)) {
+            createStream(streamName, testConfig.getShardCount());
+        }
+        log.info("Using stream " + streamName + " in endpoint " + testConfig.getEndpoint() + " with region " + testConfig.getRegion());
```
stair-aws, line 98: Use parameterized logging.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/RecordValidatorQueue.java
```diff
@@ -0,0 +1,66 @@
+package software.amazon.kinesis.utils;
+
+import lombok.extern.slf4j.Slf4j;
+import java.util.*;
+
+@Slf4j
+public class RecordValidatorQueue {
+
+    HashMap<String, List<String>> dict = new HashMap<>();
```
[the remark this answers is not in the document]
mmankika-aws: 3. The way that this class is being used is that each thread "puts" records to a different key (shardId) until all the threads are completed. Only after all threads have finished is anything read from the HashMap. Because all the writes are to different keys and no reads are done until all the threads are done, I did not think that any race conditions were possible here. There may be a case that I am missing, however.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/RecordValidatorQueue.java
```diff
@@ -0,0 +1,66 @@
...
+            oldVal.add( data );
+            dict.put( shardId, oldVal );
+        } else {
+            List<String> newVal = new ArrayList<>();
+            newVal.add( data );
+            dict.put( shardId, newVal );
+        }
+    }
+
+    public int validateRecords( int trueTotalShardCount ) {
```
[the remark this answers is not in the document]
mmankika-aws: One thing that validate looks at is that the total shard count expected is equal to the total number of successful putRecord calls made throughout the integration test. These numbers may be out of sync at the time that add is called (in processRecords) but is expected to be correct once all workers have finished processing records. Therefore, I think it makes sense to check this at the end of the test case after ensuring that all the workers have finished.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/RecordValidatorQueue.java
```diff
@@ -0,0 +1,66 @@
...
+        // Validate that each List in the HashMap has data records in increasing order
+
+        boolean incOrder = true;
+        for ( Map.Entry<String, List<String>> entry : dict.entrySet() ) {
+            List<String> recordsPerShard = entry.getValue();
+            int prevVal = -1;
+            boolean shardIncOrder = true;
+            for ( String record : recordsPerShard ) {
+                int nextVal = Integer.parseInt( record );
```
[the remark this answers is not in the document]
mmankika-aws: I make this assumption because I am working on the assumption that all the data I am putting onto streams in these integration tests are incrementing numbers "1", "2", "3", ... as BigIntegers. In TestRecordProcessorV2/processRecords, the data is read and converted from a byte array to a String. I could do the parsing here, but that seems to be to be making the same assumption that you point out here. Is there any way that I can improve this or note the assumption that all the data on these streams are integers?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/RecordValidatorQueueTest.java
```diff
@@ -0,0 +1,55 @@
+package software.amazon.kinesis.common;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.mockito.Mock;
+import software.amazon.kinesis.utils.RecordValidatorQueue;
+
+public class RecordValidatorQueueTest {
```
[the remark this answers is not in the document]
mmankika-aws: I was not sure if I needed unit tests to validate that my implementation of record validation was as expected. Would you suggest removing these tests?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLReleaseCanary2XPollingH1TestConfig.java
```diff
@@ -0,0 +1,105 @@
...
+import java.net.URISyntaxException;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.util.Date;
+
+public class KCLReleaseCanary2XPollingH1TestConfig implements KCLAppConfig {
+    @Override
+    public String getStreamName() {
+        return "KCLReleaseCanary2XPollingH1TestConfig";
```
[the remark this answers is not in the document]
mmankika-aws: After changing the name of the class to remove 2X, I will leave the string here. We should have the "2X" in the name of the stream and the lease table in case we are testing for both KCL 1.x and 2.x using the same AWS account for clarity and to avoid resources with overlapping names between 1.x and 2.x

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLReleaseCanary2XStreamingTestConfig.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public RetrievalConfig getRetrievalConfig() throws IOException, URISyntaxException {
+        LocalDateTime d = LocalDateTime.now();
+        d = d.minusMinutes( 5 );
+        Instant instant = d.atZone( ZoneId.systemDefault() ).toInstant();
+        Date startStreamTime = Date.from( instant );
+
+        InitialPositionInStreamExtended initialPosition = InitialPositionInStreamExtended
```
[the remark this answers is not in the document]
mmankika-aws: We want to start processing records that were put onto the stream in this test case. This is just a small value that estimates approximately when the test case may have started (related to the time that the consumer has finished getting started and is now processing records).

amazon-kinesis-client/pom.xml
```diff
@@ -207,6 +207,10 @@
               <name>sqlite4java.library.path</name>
               <value>${sqlite4java.libpath}</value>
             </property>
+            <property>
+              <name>credentials</name>
+              <value>${credentials}</value>
```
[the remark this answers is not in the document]
mmankika-aws: I will add the expectations of this parameter to the README. These credentials are the name of the IAM User that we want to run integration tests with. I will make it so that this property is optional. In the case that we want to override the default credentials on a system (local PC or within an AWS resource like CodePipeline), this parameter can be provided. Otherwise, the credentials installed in the test environment will be used.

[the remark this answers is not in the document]
mmankika-aws: I added some clarification to the README about how this parameter would be used. You can optionally run ` mvn -Dit.test=TESTNAME -Dcredentials=IAM_USER_NAME verify` to override the default IAM user. If this parameter is not provided in the maven command, it will use the default credentials of the test environment where the code is being run.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLAppConfig.java
```diff
@@ -0,0 +1,217 @@
...
+    boolean isProd();
+
+    default boolean durabilityCheck() {
+        return true;
+    }
+
+    // "default" profile, should match with profiles listed in "cat ~/.aws/config"
+    String getProfile();
+
+    // '-1' means round robin across 0, 5_000, 15_000, 30_000 milliseconds delay.
```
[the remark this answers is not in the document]
mmankika-aws: Deleting processingDelayMillis for now as it is not being used

amazon-kinesis-client/src/test/java/software/amazon/kinesis/integration_tests/StreamExistenceManager.java
mmankika-aws: This client is for the DescribeStreams API call. For this, it does not matter which HTTP protocol is being used so a default is put here. When making PutRecord, GetRecord calls, the Consumer Protocol specified in KCLAppConfig is used. However, I will update this so that they are consistent.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLAppConfig.java
```diff
@@ -0,0 +1,217 @@
...
+
+    default KinesisAsyncClient buildAsyncKinesisClient( Optional< Protocol > protocol ) throws URISyntaxException, IOException {
+
+        // Setup H2 client config.
+        final NettyNioAsyncHttpClient.Builder builder = NettyNioAsyncHttpClient.builder()
+                .maxConcurrency( Integer.MAX_VALUE );
+
+        if ( protocol.isPresent() ) {
+            builder.protocol( protocol.get() );
+        }
```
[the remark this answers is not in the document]
mmankika-aws: If this is not present, the default is HTTP1. I can note that here

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLReleaseCanary2XPollingH1TestConfig.java
```diff
@@ -0,0 +1,105 @@
...
+    public ReshardConfig getReshardConfig() {
+        return null;
+    }
+
+    @Override
+    public RetrievalConfig getRetrievalConfig() throws IOException, URISyntaxException {
+        LocalDateTime d = LocalDateTime.now();
+        d = d.minusMinutes( 5 );
+        Instant instant = d.atZone( ZoneId.systemDefault() ).toInstant();
+        Date startStreamTime = Date.from( instant );
```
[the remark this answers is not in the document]
mmankika-aws: Resolved in person discussion. For testing purposes, this will work as long as the region in getRegion() is the same as the local region where tests are being run.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLAppConfig.java
```diff
@@ -0,0 +1,217 @@
...
+        } else {
+            return null;
+        }
+    }
+
+    default ConfigsBuilder getConfigsBuilder() throws IOException, URISyntaxException {
+        return getConfigsBuilder( "" );
+    }
+
+    default ConfigsBuilder getConfigsBuilder( String workerIdSuffix ) throws IOException, URISyntaxException {
```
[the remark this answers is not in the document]
mmankika-aws: Yes for now we do not need suffixes. This is a parameter that was expected to be inputted by the developer doing testing. the ConfigsBuilder says that it is used to "KCL worker identifier to distinguish between 2 unique workers". For now, in the single consumer case, we do not need this.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLAppConfig.java
```diff
@@ -0,0 +1,197 @@
...
+
+        /**
+         * If not present, defaults to HTTP1_1
+         */
+        if (protocol.isPresent()) {
+            builder.protocol(protocol.get());
+        }
+
+        final SdkAsyncHttpClient sdkAsyncHttpClient =
+                builder.buildWithDefaults(AttributeMap.builder().put(SdkHttpConfigurationOption.TRUST_ALL_CERTIFICATES, true).build());
```
stair-aws, line 94: Can we remove this [`TRUST_ALL_CERTIFICATES`](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/http/SdkHttpConfigurationOption.html#TRUST_ALL_CERTIFICATES) so SSL validation is re-enabled? Do the tests execute/pass without this?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLAppConfig.java
```diff
@@ -0,0 +1,197 @@
...
+        private long callPeriodMills;
+    }
+
+    /**
+     * Description of the method of resharding for a test case
+     * <p>
+     * reshardingFactorCycle: lists the scales by which the number of shards in a stream will be updated
+     * in sequence. e.g {2.0, 0.5} means that the number of shards will first be doubled, then halved
+     * <p>
+     * numReshardCycles: the number of resharding cycles that will be executed in a test]
```
stair-aws, line 185: Nit: trailing `]`

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLAppConfig.java
```diff
@@ -0,0 +1,197 @@
...
+     * <p>
+     * reshardFrequencyMillis: the period of time between reshard cycles (in milliseconds)
```
stair-aws, line 187: FMI: what happens if the stream is still being resharded (by Kinesis) when the next "reshard cycle", controlled by this config, happens? Will this code fail, or retry?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/ReleaseCanaryPollingH1TestConfig.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public class ReleaseCanaryPollingH1TestConfig extends BasicReleaseCanaryConfig {
+    @Override
+    public String getStreamName() {
+        return "KCLReleaseCanary2XPollingH1TestConfig";
+    }
+
+    @Override
+    public int getShardCount() {
+        return 20;
```
stair-aws, line 26: Frugality: 20->4?

edit: ditto in other subclasses

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/ReleaseCanaryPollingH1TestConfig.java
```diff
@@ -0,0 +1,55 @@
...
+import java.net.URISyntaxException;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.util.Date;
+
+/**
+ * Config for a polling consumer with HTTP protocol of HTTP1
+ */
+public class ReleaseCanaryPollingH1TestConfig extends BasicReleaseCanaryConfig {
```
stair-aws, line 18: +100 These subclasses are **much** simpler. Thank you!

amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/BasicStreamingPollingIntegrationTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+import org.junit.Test;
+import software.amazon.kinesis.config.KCLAppConfig;
+import software.amazon.kinesis.config.ReleaseCanaryPollingH1TestConfig;
+import software.amazon.kinesis.config.ReleaseCanaryPollingH2TestConfig;
+import software.amazon.kinesis.config.ReleaseCanaryStreamingTestConfig;
+import software.amazon.kinesis.utils.TestConsumer;
+
+@Slf4j
```
stair-aws, line 11: Unused annotation.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/RecordValidatorQueueTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.Test;
+
+public class RecordValidatorQueueTest {
+
+    private RecordValidatorQueue recordValidator = new RecordValidatorQueue();
+
+    private static final String SHARD_ID = "ABC";
+
+    @Test
+    public void validationFailedRecordOutOfOrderTest() {
```
stair-aws, line 13: Convention: test methods are usually prefixed by `test`.  Example: `testValidationFailedRecord...`

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLAppConfig.java
```diff
@@ -0,0 +1,197 @@
...
+//    default ConfigsBuilder getConfigsBuilder(String workerIdSuffix) throws IOException, URISyntaxException {
+//        final String workerId = getWorkerId() + workerIdSuffix;
+//        if (getStreamArn() == null) {
+//            return new ConfigsBuilder(getStreamName(), getApplicationName(), buildConsumerClient(), buildAsyncDynamoDbClient(),
+//                    buildAsyncCloudWatchClient(), workerId, getShardRecordProcessorFactory());
+//        } else {
+//            return new ConfigsBuilder(Arn.fromString(getStreamArn()), getApplicationName(), buildConsumerClient(), buildAsyncDynamoDbClient(),
+//                    buildAsyncCloudWatchClient(), workerId, getShardRecordProcessorFactory());
+//        }
+//    }
```
mmankika-aws, line 163: Dead code, I have removed. Haven't pushed that change yet

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestConsumer.java
```diff
@@ -0,0 +1,202 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+@Slf4j
+public class TestConsumer {
```
stair-aws, line 37: It looks like this framework does not teardown/delete the Kinesis stream or DDB?

Please ensure the following:
1. Test resources are deleted on teardown -- both on test success, and test failure (e.g., thrown Exception that fails the whole process).
    * We don't want inert test resources to continue stockpiling.
    * We don't want test resources to be reused (e.g., DDB) which may contaminate a test-run leading to either a false success or false failure.
3. Resources created per test are unique. Your [comment](https://github.com/awslabs/amazon-kinesis-client/pull/1136#discussion_r1228550016) highlighted a problem w/ the non-uniqueness.
    * Uniqueness helps avoid test contamination.
    * Uniqueness allows parallelization of multiple tests. (Having one name per test partially addresses this, but what if we need to execute multiple tests of the same "name"?)

stair-aws (changes requested):
Partial review.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLAppConfig.java
```diff
@@ -0,0 +1,197 @@
...
+    default KinesisAsyncClient buildAsyncKinesisClient(Optional<Protocol> protocol) throws URISyntaxException, IOException {
+
+        /**
+         * Setup H2 client config.
+         */
+        final NettyNioAsyncHttpClient.Builder builder = NettyNioAsyncHttpClient.builder()
+                .maxConcurrency(Integer.MAX_VALUE);
+
+        /**
+         * If not present, defaults to HTTP1_1
```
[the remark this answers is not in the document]
mmankika-aws: In [this class](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/http/SdkHttpConfigurationOption.html#PROTOCOL), there are private default settings that set the DEFAULT_PROTOCOL as HTTP1_1

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLAppConfig.java
```diff
@@ -0,0 +1,197 @@
...
+
+    ReshardConfig getReshardConfig();
+
+    default KinesisAsyncClient buildConsumerClient() throws URISyntaxException, IOException {
+        return buildAsyncKinesisClient(getConsumerProtocol());
+    }
+
+    default KinesisAsyncClient buildProducerClient() throws URISyntaxException, IOException {
+        return buildAsyncKinesisClient(getProducerProtocol());
+    }
```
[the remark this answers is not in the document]
mmankika-aws: We specify two different clients because in many cases, they will have different protocols. The producer will always have a protocol of HTTP1. For KCL 2.x, the consumer can have either HTTP1 or HTTP2 (if it is HTTP2, the two clients cannot be the same).

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLAppConfig.java
mmankika-aws: This will cause this individual reshard to fail (not the whole cycle) and it will move on to the next reshard in the cycle (or next cycle, if at the end of the current cycle).

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/ReleaseCanaryPollingH1TestConfig.java
mmankika-aws: :)

README.md
```diff
@@ -34,7 +34,13 @@ Please open an issue if you have any questions.
...
  this command: `mvn clean install -Dgpg.skip=true`. Note: This command runs Integration tests, which in turn creates AWS
   resources (which requires manual cleanup). Integration tests require valid AWS credentials need to be discovered at
-   runtime. To skip running integration tests, add ` -DskipITs` option to the build command.  
+   runtime. To skip running integration tests, add ` -DskipITs` option to the build command.
+
+## Running Integration Tests
+
+To run integration tests to test any changes to KCL, you can use this command: `mvn -Dit.test=*IntegrationTest verify`. 
+This will look for default AWS credentials in your local `.aws/credentials`. If you want to override these
+credentials, you can provide the name of an IAM user as a string using this command: `mvn -Dit.test=*IntegrationTest -Dcredentials="<IAM_USER>" verify`.
```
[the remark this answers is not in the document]
mmankika-aws: Yes, this looks for a profile in the .aws/credentials file (this holds the access key and secret access key of a profile). I will update this wording.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/BasicReleaseCanaryConfig.java
```diff
@@ -0,0 +1,108 @@
...
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.util.Date;
+
+/**
+ * Basic config for a release canary (streaming) with default settings
```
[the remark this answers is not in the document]
mmankika-aws: Since this file was combined with KCLAppConfig, these comments are now addressed in that file. I have updated the description at the top. Streaming (vs polling) refers to the protocol for getting data. In the streaming scenario, the connection between the producer and the consumer is made once and the producer sends data continuously while the connection is maintained. In the polling scenario, the consumer calls the producer each time it wants records to process.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/BasicReleaseCanaryConfig.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * This will get the credentials that are provided in the maven command
+     * when running integration tests if any are provided through -Dcredentials=iamUser
+     * Otherwise, iamUser will be null and the test will search for default credentials
+     * in the test environment.
+     */
+    @Override
+    public String getProfile() {
+        String iamUser = System.getProperty("credentials");
```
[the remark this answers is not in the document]
mmankika-aws: If this argument is provided, a [ProfileCredentialsProvider](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/profile/ProfileCredentialsProvider.html) is created. Otherwise, a default AWSCredentialsProvider is created to the account whos credentials are found at .aws/configuration as mentioned in the README.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLAppConfig.java
mmankika-aws: Yes they do, removing this

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/BasicReleaseCanaryConfig.java
```diff
@@ -0,0 +1,108 @@
...
+@Slf4j
+public class BasicReleaseCanaryConfig implements KCLAppConfig {
+    @Override
+    public String getStreamName() {
+        return "";
+    }
+
+    @Override
+    public int getShardCount() {
+        return 10;
```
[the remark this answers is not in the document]
mmankika-aws: Changed to 4 shards

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestConsumer.java
mmankika-aws: 1. Implemented resource teardown on success or fail of test case. 
2. Resources created (stream and DDB) both have name getStreamName(). As long as these are unique per test case, there will be no issues with overlap. Not really sure that I understand your point about "executing multiple tests of the same name"

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/ReleaseCanaryPollingH1TestConfig.java
```diff
@@ -0,0 +1,37 @@
...
+import software.amazon.kinesis.retrieval.RetrievalConfig;
+import software.amazon.kinesis.retrieval.polling.PollingConfig;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+
+/**
+ * Config for a polling consumer with HTTP protocol of HTTP1
+ */
+public class ReleaseCanaryPollingH1TestConfig extends KCLAppConfig {
```
mmankika-aws, line 14: Are these names still unclear? Not sure what I could change them to to make it more clear.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestConsumer.java
```diff
@@ -0,0 +1,208 @@
...
+        } catch (Exception e) {
+            log.error("Error creating payload data for {}", payloadCounter.toString());
+            throw new RuntimeException("Error converting object to bytes: ", e);
+        }
+        return ByteBuffer.wrap(returnData);
+    }
+
+    private void deleteResources(StreamExistenceManager streamExistenceManager, DynamoDbAsyncClient dynamoDBClient) throws Exception {
+        log.info("-------------Start deleting test resources.----------------");
+        streamExistenceManager.deleteStream(this.streamName);
```
ZeyuLi-AWS, line 193: Can separately log deleting resources, i.e. stream and table.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/RecordValidatorQueue.java
```diff
@@ -0,0 +1,74 @@
...
+        // Validate that no records are missing over all shards
+        int totalShardCount = 0;
+        for (Map.Entry<String, List<String>> entry : dict.entrySet()) {
+            List<String> recordsPerShard = entry.getValue();
+            Set<String> noDupRecords = new HashSet<String>(recordsPerShard);
+            totalShardCount += noDupRecords.size();
+        }
+
+        // If this is true, then there was some record that was missed during processing.
+        if (totalShardCount != trueTotalShardCount) {
```
ZeyuLi-AWS, line 65: Can change to "expectedShardCount" and "actuaShardCounl" for convention

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/StreamExistenceManager.java
```diff
@@ -0,0 +1,125 @@
...
+        try {
+            client.createStream(request).get(30, TimeUnit.SECONDS);
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to create stream with name " + streamName, e);
+        }
+
+        int i = 0;
+        while (true) {
+            i++;
+            if (i > 100) {
```
ZeyuLi-AWS, line 68: Why do we want to choose 100*10 s this magic number?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/StreamExistenceManager.java
```diff
@@ -0,0 +1,125 @@
...
+
+        try {
+            final DescribeStreamSummaryResponse response = describeStreamSummaryResponseCompletableFuture.get(30, TimeUnit.SECONDS);
+            boolean isActive = response.streamDescriptionSummary().streamStatus().equals(StreamStatus.ACTIVE);
+            if (!isActive) {
+                throw new RuntimeException("Stream is not active, instead in status: " + response.streamDescriptionSummary().streamStatus());
+            }
+            return true;
+        } catch (ExecutionException e) {
+            if (e.getCause() instanceof ResourceNotFoundException) {
```
ZeyuLi-AWS, line 47: Why can't we directly check `catch (ResourceNotFoundException e)`

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestRecordProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+import software.amazon.kinesis.exceptions.ShutdownException;
+import software.amazon.kinesis.lifecycle.events.LeaseLostInput;
+import software.amazon.kinesis.lifecycle.events.ShardEndedInput;
+import software.amazon.kinesis.lifecycle.events.ShutdownRequestedInput;
+import software.amazon.kinesis.processor.ShardRecordProcessor;
+import software.amazon.kinesis.lifecycle.events.InitializationInput;
+import software.amazon.kinesis.retrieval.KinesisClientRecord;
+
+import java.nio.ByteBuffer;
+
```
ZeyuLi-AWS, line 15: It'd better tp add the class doc here

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestRecordProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+        try {
+            log.info("Initializing @ Sequence: {}", initializationInput.extendedSequenceNumber());
+        } finally {
+            MDC.remove(SHARD_ID_MDC_KEY);
+        }
+    }
+
+
+    @Override
+    public void processRecords(software.amazon.kinesis.lifecycle.events.ProcessRecordsInput processRecordsInput) {
```
ZeyuLi-AWS, line 42: Why didn't you import this package just like others `mport software.amazon.kinesis.lifecycle.events.ShutdownRequestedInput;`

README.md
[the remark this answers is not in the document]
mmankika-aws: All of this terminology is described in the AWS docs that go along with this code ([example](https://docs.aws.amazon.com/streams/latest/dev/enhanced-consumers.html)). I don't think it is necessary to repeat the same information in the README.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLAppConfig.java
```diff
@@ -87,24 +86,14 @@ public ReshardConfig getReshardConfig() {
...
-    }
-
-    private AwsCredentialsProvider getCredentialsProvider() {
-        return (getProfile() != null) ?
-                ProfileCredentialsProvider.builder().profileName(getProfile()).build() : DefaultCredentialsProvider.create();
+    public final KinesisAsyncClient buildAsyncKinesisClient(Protocol protocol) throws URISyntaxException, IOException {
+        if (kinesisAsyncClient == null) {
+            this.kinesisAsyncClient = buildAsyncKinesisClient(Optional.ofNullable(protocol));
+        }
+        return this.kinesisAsyncClient;
```
abhit17, line 93: This is duplicate method, remove this.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/config/KCLAppConfig.java
```diff
@@ -128,26 +117,33 @@ public KinesisAsyncClient buildAsyncKinesisClient(Optional<Protocol> protocol) t
...
+    public final CloudWatchAsyncClient buildAsyncCloudWatchClient() throws IOException {
+        if (this.cloudWatchAsyncClient == null) {
+            final CloudWatchAsyncClientBuilder builder = CloudWatchAsyncClient.builder().region(getRegion());
+            builder.credentialsProvider(getCredentialsProvider());
+            this.cloudWatchAsyncClient = builder.build();
+        }
+        return this.cloudWatchAsyncClient;
     }
 
     public String getWorkerId() throws UnknownHostException {
```
abhit17, line 138: private method

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/LeaseTableManager.java
```diff
@@ -0,0 +1,74 @@
...
+
+        // Wait till table is deleted to return
+        int i = 0;
+        while (true) {
+            i++;
+            if (i > 100) {
+                throw new RuntimeException("Failed lease table deletion");
+            }
+            try {
+                if (!listAllLeaseTables().contains(tableName)) {
```
abhit17, line 53: describeTable call here instead of list Table call

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/LeaseTableManager.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public LeaseTableManager(DynamoDbAsyncClient dynamoClient) throws URISyntaxException, IOException {
+        this.dynamoClient = dynamoClient;
+    }
+
+    private List<String> listAllLeaseTables() throws Exception {
+        final ListTablesRequest request = ListTablesRequest.builder().build();
+        final ListTablesResponse response;
+        try {
+            response = FutureUtils.resolveOrCancelFuture(dynamoClient.listTables(request), Duration.ofSeconds(60));
```
abhit17, line 30: this is paginated call, u might not get all the tables in one go, perform paginated call

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/LeaseTableManager.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.time.Duration;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+@Slf4j
+public class LeaseTableManager {
```
abhit17, line 18: you can use better abstractions to remove the redundant code.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestConsumer.java
```diff
@@ -50,33 +43,87 @@ public class TestConsumer {
...
     }
 
     public void run() throws Exception {
 
+        final StreamExistenceManager streamExistenceManager = new StreamExistenceManager(this.consumerConfig);
+        final LeaseTableManager leaseTableManager = new LeaseTableManager(this.dynamoClient);
+
+        // Clean up any old streams or lease tables left in test environment
+        cleanTestEnvironment(streamExistenceManager, leaseTableManager);
+        Thread.sleep(TimeUnit.SECONDS.toMillis(30));
```
abhit17, line 68: once we delete and validated the delete is done no need to sleep again.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestConsumer.java
```diff
@@ -98,66 +145,21 @@ public void run() throws Exception {
...
-
-        } catch (Exception e) {
-            // Test Failed. Clean up resources and then throw exception.
-            Thread.sleep(TimeUnit.SECONDS.toMillis(30));
-            deleteResources(streamExistenceManager, dynamoClient);
-            throw e;
-        }
+    private void startConsumer() {
+        // Start record processing of dummy data
+        final Thread schedulerThread = new Thread(scheduler);
```
abhit17, line 152: use executor service

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestConsumer.java
```diff
@@ -50,33 +43,87 @@ public class TestConsumer {
...
+        this.dynamoClient = consumerConfig.buildAsyncDynamoDbClient();
     }
 
     public void run() throws Exception {
 
+        final StreamExistenceManager streamExistenceManager = new StreamExistenceManager(this.consumerConfig);
+        final LeaseTableManager leaseTableManager = new LeaseTableManager(this.dynamoClient);
+
+        // Clean up any old streams or lease tables left in test environment
+        cleanTestEnvironment(streamExistenceManager, leaseTableManager);
```
abhit17, line 67: rename to cleanTestResources

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestConsumer.java
```diff
@@ -50,33 +43,87 @@ public class TestConsumer {
...
     public int successfulPutRecords = 0;
     public BigInteger payloadCounter = new BigInteger("0");
 
-    public TestConsumer(KCLAppConfig consumerConfig) {
+    public TestConsumer(KCLAppConfig consumerConfig) throws Exception {
         this.consumerConfig = consumerConfig;
         this.region = consumerConfig.getRegion();
         this.streamName = consumerConfig.getStreamName();
-        this.kinesisClient = KinesisClientUtil.createKinesisAsyncClient(KinesisAsyncClient.builder().region(this.region));
+        this.kinesisClient = consumerConfig.buildAsyncKinesisClient(consumerConfig.getConsumerProtocol());
```
abhit17, line 57: have this protocal in the config itself which you are creating based on your test config

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestConsumer.java
```diff
@@ -50,33 +43,87 @@ public class TestConsumer {
...
+        log.info("----------Finished deleting all streams-------------");
+    }
+
+    private void startProducer() {
+        // Send dummy data to stream
+        this.producerExecutor = Executors.newSingleThreadScheduledExecutor();
+        this.producerFuture = producerExecutor.scheduleAtFixedRate(this::publishRecord, 60, 1, TimeUnit.SECONDS);
+    }
+
+    private void setUpTestResources() throws Exception {
```
abhit17, line 124: rename setupConsumerResources

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestConsumer.java
```diff
@@ -188,23 +188,42 @@ private ByteBuffer wrapWithCounter(int payloadSize, BigInteger payloadCounter) t
...
+    }
 
+    private void validateRecordProcessor() throws Exception {
+        log.info("The number of expected records is: {}", successfulPutRecords);
+        final RecordValidationStatus errorVal = consumerConfig.getRecordValidator().validateRecords(successfulPutRecords);
+        if (errorVal == RecordValidationStatus.OUT_OF_ORDER) {
+            throw new RuntimeException("There was an error validating the records that were processed. The records were out of order");
+        } else if (errorVal == RecordValidationStatus.MISSING_RECORD) {
+            throw new RuntimeException("There was an error validating the records that were processed. Some records were missing.");
+        }
```
abhit17, line 217: combine into one ?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestRecordProcessorFactory.java
```diff
@@ -0,0 +1,19 @@
+package software.amazon.kinesis.utils;
+
+import software.amazon.kinesis.processor.ShardRecordProcessor;
+import software.amazon.kinesis.processor.ShardRecordProcessorFactory;
+
+public class TestRecordProcessorFactory implements ShardRecordProcessorFactory {
+
+    RecordValidatorQueue recordValidator;
```
abhit17, line 8: private final

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestRecordProcessor.java
```diff
@@ -0,0 +1,108 @@
...
+ * Implement initialization and deletion of shards and shard record processing
+ */
+@Slf4j
+public class TestRecordProcessor implements ShardRecordProcessor {
+
+    private static final String SHARD_ID_MDC_KEY = "ShardId";
+
+    private String shardId;
+
+    RecordValidatorQueue recordValidator;
```
abhit17, line 27: private final

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestRecordProcessor.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+
+    @Override
+    public void processRecords(ProcessRecordsInput processRecordsInput) {
+        MDC.put(SHARD_ID_MDC_KEY, shardId);
+        try {
+            log.info("Processing {} record(s)", processRecordsInput.records().size());
+
+            for (KinesisClientRecord r : processRecordsInput.records()) {
```
abhit17, line 51: np r -> kinesisRecord or similar. Dont use single alphabet words

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestRecordProcessor.java
```diff
@@ -0,0 +1,108 @@
...
+                String data = new String(asByteArray(r.data()));
```
abhit17, line 52: final

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestConsumer.java
```diff
@@ -0,0 +1,229 @@
...
+            validateRecordProcessor();
+
+            // Clean up resources created
+            Thread.sleep(TimeUnit.SECONDS.toMillis(30));
+            deleteResources(streamExistenceManager, leaseTableManager);
+
+        } catch (Exception e) {
+            // Test Failed. Clean up resources and then throw exception.
+            log.info("----------Test Failed: Cleaning up resources------------");
+            Thread.sleep(TimeUnit.SECONDS.toMillis(30));
```
abhit17, line 101: Remove this delay, we dont need any specific delay on error

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/RecordValidatorQueue.java
```diff
@@ -0,0 +1,74 @@
...
+        boolean incOrder = true;
+        for (Map.Entry<String, List<String>> entry : dict.entrySet()) {
+            List<String> recordsPerShard = entry.getValue();
+            int prevVal = -1;
+            boolean shardIncOrder = true;
+            for (String record : recordsPerShard) {
+                int nextVal = Integer.parseInt(record);
+                if (prevVal > nextVal) {
+                    log.error("The records are not in increasing order. Saw record data {} before {}.", prevVal, nextVal);
+                    shardIncOrder = false;
```
abhit17, line 41: after this point, no need to check the next records of this shard or even other shards, test is going to fail anyways, Fail here directly and go back

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestConsumer.java
mmankika-aws: I think it is helpful to have a message telling which one of these two errors occurred because both are checked in the validation function. If you do not think so, I can change that function to just return true (success) or false (fail).

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestRecordProcessorFactory.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public TestRecordProcessorFactory(RecordValidatorQueue recordValidator) {
```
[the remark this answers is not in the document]
mmankika-aws: This gives a compile error

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/LeaseTableManager.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+@Slf4j
+public class LeaseTableManager extends AWSResourceManager {
+
+    private final DynamoDbAsyncClient dynamoClient;
+
+    public LeaseTableManager(DynamoDbAsyncClient dynamoClient) throws URISyntaxException, IOException {
+        this.dynamoClient = dynamoClient;
```
abhit17, line 29: use lombok's @AllArgConstructor

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/TestConsumer.java
```diff
@@ -0,0 +1,226 @@
...
+            Thread.sleep(TimeUnit.SECONDS.toMillis(10));
+
+            // Finishes processing current batch of data already received from Kinesis before shutting down.
+            awaitConsumerFinish();
+
+            // Validate processed data
+            validateRecordProcessor();
+
+            // Clean up resources created
+            Thread.sleep(TimeUnit.SECONDS.toMillis(30));
```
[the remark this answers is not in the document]
mmankika-aws: There is sometimes a delay in the workers getting shutdown and deleted. This delay ensures that that process is complete before we try to delete any resources.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/BasicStreamingPollingIntegrationTest.java
```diff
@@ -0,0 +1,32 @@
...
+import software.amazon.kinesis.config.KCLAppConfig;
+import software.amazon.kinesis.config.ReleaseCanaryPollingH1TestConfig;
+import software.amazon.kinesis.config.ReleaseCanaryPollingH2TestConfig;
+import software.amazon.kinesis.config.ReleaseCanaryStreamingTestConfig;
+import software.amazon.kinesis.utils.TestConsumer;
+
+public class BasicStreamingPollingIntegrationTest {
+
+    @Test
+    public void KCLReleaseCanaryPollingH2Test() throws Exception {
```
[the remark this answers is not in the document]
stair-aws: > also better naming convention for tests is "MethodName_StateUnderTest_ExpectedBehavior" ref https://medium.com/@stefanovskyi/unit-test-naming-conventions-dd9208eadbea

The above statement is subjective.

Fact: Java [method naming convention](https://en.wikipedia.org/wiki/Naming_convention_(programming)#Java) is lowerCamelCase.

From [Javadoc wiki](https://en.wikipedia.org/wiki/Javadoc):
> Writing comments and Javadoc is for better understanding the code and thus better maintaining it. 

---
Opinion 1: Natural language (e.g., English) is much better at communicating intent than anAbridgedCamelCaseMashup, regardless of whether underscores are used as delimiters.
Opinion 2: The proposed method naming suggestion feels like a manifestation of [Law of the instrument](https://en.wikipedia.org/wiki/Law_of_the_instrument). (e.g., "If the only tool you have is a hammer, it is tempting to treat everything as if it were a nail")

Suggestion: if the test and intent cannot be made evident w/ a conventional method name, **add Javadoc**. Document the intent, preconditions, postconditions, and anything else that a future maintainer might find useful or informative. Natural language does not have compile-time constraints, and provides a much broader palette on which the author may convey details.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/utils/LeaseTableManager.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public List<String> _getAllResourceNames() throws Exception {
+        List<String> tableNames = new ArrayList<>();
+        final ListTablesRequest request = ListTablesRequest.builder().build();
+        ListTablesResponse response = null;
+        while(response == null || response.lastEvaluatedTableName() != null) {
+            try {
+                response = FutureUtils.resolveOrCancelFuture(dynamoClient.listTables(request), Duration.ofSeconds(60));
+            } catch (ExecutionException | InterruptedException e) {
```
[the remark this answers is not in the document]
mmankika-aws: Fixing the request that is made. With this new change there are many places in this method that are resource dependent (the initialization at the top, while condition, try block, addAll call...) that would need to be implemented in the resource specific managers that I think that abstracting this would harm readability more than it would help.

abhit17 (approved):

abhit17 (approved):

PR #1140, opened 2023-06-13
Multilang streamarn update

pelaezryan:
*Description of changes:*
1. Updated multilang daemon to support 'streamArn' being passed in via Properties
2. streamName and regionName is parsed out if 'streamArn' is passed in. 
    -  **Meaning streamArn takes precedence if it is passed in**
3. Added unit test

===========================================================================================

### DIFF
--- a/amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/config/KinesisClientLibConfigurator.java
+++ b/amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/config/KinesisClientLibConfigurator.java
@@ -26,6 +26,7 @@ import org.apache.commons.beanutils.ConvertUtilsBean;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.Validate;
+import software.amazon.awssdk.regions.Region;
 
 /**
  * KinesisClientLibConfigurator constructs a KinesisClientLibConfiguration from java properties file. The following
@@ -76,12 +77,19 @@ public class KinesisClientLibConfigurator {
         Validate.notBlank(configuration.getApplicationName(), "Application name is required");
 
         try {
-            Validate.notBlank(configuration.getStreamName(), "");
-        }catch (Exception e) {
-            Validate.notBlank(configuration.getStreamArn(), "Stream name or Stream Arn is required. (Stream Name takes precedence if both are passed in)");
+            Validate.notBlank(configuration.getStreamArn(), "");
             Arn streamArnObj = Arn.fromString(configuration.getStreamArn());
+
+            //Parse out the stream Name from the Arn (and/or override existing value for Stream Name)
             String streamNameFromArn = streamArnObj.getResource().getResource();
             configuration.setStreamName(streamNameFromArn);
+
+            //Parse out the region from the Arn and set (and/or override existing value for region)
+            String regionName = streamArnObj.getRegion();
+            Region regionObj = Region.of(regionName);
+            configuration.setRegionName(regionObj);
+        }catch (Exception e) {
+            Validate.notBlank(configuration.getStreamName(), "Stream name or Stream Arn is required. (Stream Arn takes precedence if both are passed in)");
 
         }
         Validate.isTrue(configuration.getKinesisCredentialsProvider().isDirty(), "A basic set of AWS credentials must be provided");
--- a/amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/config/KinesisClientLibConfigurator.java
+++ b/amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/config/KinesisClientLibConfigurator.java
@@ -74,15 +74,15 @@ public class KinesisClientLibConfigurator {
         });
 
         Validate.notBlank(configuration.getApplicationName(), "Application name is required");
+
         try {
+            Validate.notBlank(configuration.getStreamName(), "");
+        }catch (Exception e) {
+            Validate.notBlank(configuration.getStreamArn(), "Stream name or Stream Arn is required. (Stream Name takes precedence if both are passed in)");
             Arn streamArnObj = Arn.fromString(configuration.getStreamArn());
             String streamNameFromArn = streamArnObj.getResource().getResource();
-            Validate.notBlank(streamNameFromArn, "");
-
-            //Stream Arn takes precedence. Override existing configuration with the stream name found in the Arn
             configuration.setStreamName(streamNameFromArn);
-        }catch (Exception e) {
-            Validate.notBlank(configuration.getStreamName(), "Stream name or Stream Arn is required. (Stream Arn takes precedence if both are passed in)");
+
         }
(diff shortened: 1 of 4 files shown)


amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/config/KinesisClientLibConfigurator.java
```diff
@@ -62,14 +64,34 @@ public KinesisClientLibConfigurator() {
     public MultiLangDaemonConfiguration getConfiguration(Properties properties) {
         properties.entrySet().forEach(e -> {
             try {
-                utilsBean.setProperty(configuration, (String) e.getKey(), e.getValue());
+                Object value = e.getValue();
+                if(value instanceof String){
+                    value = ((String) value).trim();
```
noahbt, line 69: Is there any case where the property value could have leading/trailing spaces?

amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/config/KinesisClientLibConfigurator.java
```diff
@@ -62,14 +64,34 @@ public KinesisClientLibConfigurator() {
...
+
+            //Parse out the stream Name from the Arn (and/or override existing value for Stream Name)
+            String streamNameFromArn = streamArnObj.getResource().getResource();
+            configuration.setStreamName(streamNameFromArn);
+
+            //Parse out the region from the Arn and set (and/or override existing value for region)
+            String regionName = streamArnObj.getRegion();
+            Region regionObj = Region.of(regionName);
+            configuration.setRegionName(regionObj);
+        }catch (Exception e) {
```
noahbt, line 91: nit: space before catch `} catch (Exc..`

amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/config/KinesisClientLibConfigurator.java
```diff
@@ -62,14 +64,34 @@ public KinesisClientLibConfigurator() {
...
+            Validate.notBlank(configuration.getStreamName(), "Stream name or Stream Arn is required. (Stream Arn takes precedence if both are passed in)");
```
noahbt, line 92: Is it correct that if Stream Arn is passed in, then Stream Name is skipped? If Stream Arn throws an exception, then Stream Name is checked?

amazon-kinesis-client-multilang/src/test/java/software/amazon/kinesis/multilang/MultiLangDaemonConfigTest.java
```diff
@@ -48,39 +56,131 @@ public class MultiLangDaemonConfigTest {
...
 
+        AssertConfigurationsMatch(deamonConfig, TestExe, TestApplicationName, TestStreamNameInArn, TestRegionInArn);
+    }
+
+    /**
+     * Verify the daemonConfig properties are what we expect them to be.
+     * @param deamonConfig
+     * @param expectedStreamName
+     */
+    private void AssertConfigurationsMatch(MultiLangDaemonConfig deamonConfig,
```
lucienlu-aws, line 165: Nit: Java convention should make the first letter of a method lower case.

amazon-kinesis-client-multilang/src/test/java/software/amazon/kinesis/multilang/config/KinesisClientLibConfiguratorTest.java
```diff
@@ -306,12 +306,33 @@ public void testWithMissingWorkerId() {
...
+        thrown.expect(IllegalArgumentException.class);
+        thrown.expectMessage("Stream name or Stream Arn is required. (Stream Arn takes precedence if both are passed in)");
+
+        String test = StringUtils.join(new String[] {
+                        "applicationName = b",
+                        "AWSCredentialsProvider = " + credentialName1,
+                        "workerId = 123",
+                        "failoverTimeMillis = 100",
+                        "streamName = ",
+                        "streamArn = "},
```
lucienlu-aws, line 334: Same comment in other test class, add some tests where it passes in an invalid string for an Arn (invalid stream Arn) with various cases

amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/config/KinesisClientLibConfigurator.java
lucienlu-aws, line 91: This will catch and swallow exceptions thrown from sdk when calling `Arn.fromString()` with an invalid Arn. IS this the error message we want to give to the user if, for example, they pass in an invalid stream Arn and no streamName?

Suggest only catching the `Validate.notBlank` exceptions and letting the SDK throw an error if an invalid Arn is passed.

amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/config/KinesisClientLibConfigurator.java
lucienlu-aws: +1, does this break any backwards compatibility where customers could have accidentally included spaces in their property values previously that won't work now?

amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/config/KinesisClientLibConfigurator.java
pelaezryan: Correct, if both are passed in Stream Arn will take precedence.

amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/config/MultiLangDaemonConfiguration.java
```diff
@@ -73,6 +73,8 @@ public class MultiLangDaemonConfiguration {
     private String applicationName;
 
     private String streamName;
+    private String streamArn;
```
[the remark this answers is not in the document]
pelaezryan: Yeah we can do that

amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/config/KinesisClientLibConfigurator.java
pelaezryan: good callout, i don't believe so but ill remove that to be safe.

amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/multilang/config/MultiLangDaemonConfiguration.java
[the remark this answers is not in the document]
pelaezryan: I don't believe so. 

Reason being...
When we go through the properties to be set in the configuration it checks the object type and attempts to have it as that type. Because of that, streamArn will always be seen as type 'String'. An explicit cast from string to Arn does not work.

PR #1152, opened 2023-06-26
Adding resharding integration tests and changing ITs to not run by default

mmankika-aws:
*Issue #, if available:*

*Description of changes:*

Adding KCL 2.x reshard integration test. Updating settings in pom.xml so that integration tests don't run by default. The instructions to build source and run tests in README has been updated to reflect this. 

### DIFF
--- a/amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
+++ b/amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
@@ -151,7 +151,9 @@ public class TestConsumer {
         checkpointConfig = configsBuilder.checkpointConfig();
         coordinatorConfig = configsBuilder.coordinatorConfig();
         leaseManagementConfig = configsBuilder.leaseManagementConfig()
-                .initialPositionInStream(InitialPositionInStreamExtended.newInitialPosition(consumerConfig.getInitialPosition()))
+                .initialPositionInStream(
+                    InitialPositionInStreamExtended.newInitialPosition(consumerConfig.getInitialPosition())
+                )
                 .initialLeaseTableReadCapacity(50).initialLeaseTableWriteCapacity(50);
         lifecycleConfig = configsBuilder.lifecycleConfig();
         processorConfig = configsBuilder.processorConfig();
--- a/amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
+++ b/amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
@@ -55,7 +55,7 @@ public class TestConsumer {
     private LifecycleConfig lifecycleConfig;
     private ProcessorConfig processorConfig;
     private Scheduler scheduler;
-    private static final ScheduledExecutorService PRODUCER_EXECUTOR = Executors.newSingleThreadScheduledExecutor();
+    private ScheduledExecutorService producerExecutor;
     private ScheduledFuture<?> producerFuture;
     private ScheduledExecutorService consumerExecutor;
     private ScheduledFuture<?> consumerFuture;
@@ -127,7 +127,8 @@ public class TestConsumer {
     }
 
     private void startProducer() {
-        producerFuture = PRODUCER_EXECUTOR.scheduleAtFixedRate(this::publishRecord, 10, 1, TimeUnit.SECONDS);
+        this.producerExecutor = Executors.newSingleThreadScheduledExecutor();
+        this.producerFuture = producerExecutor.scheduleAtFixedRate(this::publishRecord, 10, 1, TimeUnit.SECONDS);
 
         // Reshard logic if required for the test
         if (consumerConfig.getReshardFactorList() != null) {
@@ -137,7 +138,7 @@ public class TestConsumer {
 
             // Schedule the stream scales 4 minutes apart with 2 minute starting delay
             for (int i = 0; i < consumerConfig.getReshardFactorList().size(); i++) {
-                PRODUCER_EXECUTOR.schedule(s, (4 * i) + 2, TimeUnit.MINUTES);
+                producerExecutor.schedule(s, (4 * i) + 2, TimeUnit.MINUTES);
             }
         }
     }
@@ -179,8 +180,8 @@ public class TestConsumer {
         if (producerFuture != null) {
             producerFuture.cancel(false);
         }
-        if (PRODUCER_EXECUTOR != null) {
-            PRODUCER_EXECUTOR.shutdown();
+        if (producerExecutor != null) {
+            producerExecutor.shutdown();
         }
     }
 
--- a/amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
+++ b/amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
@@ -89,7 +89,8 @@ public class TestConsumer {
         try {
             startConsumer();
 
-            // Sleep to allow the producer/consumer to run and then end the test case. If non-reshard sleep 3 minutes, else sleep 4 minutes per scale.
+            // Sleep to allow the producer/consumer to run and then end the test case.
+            // If non-reshard sleep 3 minutes, else sleep 4 minutes per scale.
             final int sleepMinutes = (consumerConfig.getReshardFactorList() == null) ? 3 : (4 * consumerConfig.getReshardFactorList().size());
             Thread.sleep(TimeUnit.MINUTES.toMillis(sleepMinutes));
 
@@ -134,7 +135,12 @@ public class TestConsumer {
         if (consumerConfig.getReshardFactorList() != null) {
             log.info("----Reshard Config found: {}", consumerConfig.getReshardFactorList());
 
-            final StreamScaler s = new StreamScaler(kinesisClient, consumerConfig.getStreamName(), consumerConfig.getReshardFactorList(), consumerConfig);
+            final StreamScaler s = new StreamScaler(
+                    kinesisClient,
+                    consumerConfig.getStreamName(),
+                    consumerConfig.getReshardFactorList(),
+                    consumerConfig
+            );
 
             // Schedule the stream scales 4 minutes apart with 2 minute starting delay
             for (int i = 0; i < consumerConfig.getReshardFactorList().size(); i++) {
--- a/amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
+++ b/amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
@@ -183,7 +183,7 @@ public class TestConsumer {
         this.consumerFuture = consumerExecutor.schedule(scheduler, 0, TimeUnit.SECONDS);
     }
 
-    public void stopProducer() {
+    private void stopProducer() {
         log.info("Cancelling producer and shutting down executor.");
         if (producerFuture != null) {
             producerFuture.cancel(false);
--- a/amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
+++ b/amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
@@ -1,13 +1,21 @@
-package software.amazon.kinesis.utils;
+package software.amazon.kinesis.application;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.RandomStringUtils;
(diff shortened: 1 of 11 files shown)


amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
```diff
@@ -115,9 +134,28 @@ private void cleanTestResources(StreamExistenceManager streamExistenceManager, L
...
     private void startProducer() {
-        // Send dummy data to stream
-        this.producerExecutor = Executors.newSingleThreadScheduledExecutor();
-        this.producerFuture = producerExecutor.scheduleAtFixedRate(this::publishRecord, 60, 1, TimeUnit.SECONDS);
+        producerExecutor = Executors.newSingleThreadScheduledExecutor();
+        producerFuture = producerExecutor.scheduleAtFixedRate(this::publishRecord, 10, 1, TimeUnit.SECONDS);
+
+        // Reshard logic if required for the test
+        if (consumerConfig.getReshardConfig() != null) {
+            log.info("------------------------- Reshard Config found -----------------------------");
```
stair-aws, line 142: Nit: `---` is unnecessary.

This log statement **lacks context**. Log the reshard config.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
```diff
@@ -115,9 +134,28 @@ private void cleanTestResources(StreamExistenceManager streamExistenceManager, L
     }
 
     private void startProducer() {
-        // Send dummy data to stream
-        this.producerExecutor = Executors.newSingleThreadScheduledExecutor();
-        this.producerFuture = producerExecutor.scheduleAtFixedRate(this::publishRecord, 60, 1, TimeUnit.SECONDS);
+        producerExecutor = Executors.newSingleThreadScheduledExecutor();
```
stair-aws, line 137: This instantiation should occur outside this block, otherwise there's a non-zero risk that an Exception, or future refactor, might raise NPEs.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
```diff
@@ -115,9 +134,28 @@ private void cleanTestResources(StreamExistenceManager streamExistenceManager, L
...
+        producerFuture = producerExecutor.scheduleAtFixedRate(this::publishRecord, 10, 1, TimeUnit.SECONDS);
+
+        // Reshard logic if required for the test
+        if (consumerConfig.getReshardConfig() != null) {
+            log.info("------------------------- Reshard Config found -----------------------------");
+            final Queue<ReshardOptions> reshardQueue = new LinkedList<>(Arrays.asList(consumerConfig.getReshardConfig().getReshardingFactorCycle()));
```
stair-aws, line 143: FMI: why convert to a `LinkedList|Queue`? What is the expected gain?

Why is the resharding factor cycle defined as a primitive array instead of a `List`? That decision appears to introduce unnecessary complexity.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
```diff
@@ -220,4 +258,53 @@ private void deleteResources(StreamExistenceManager streamExistenceManager, Leas
         log.info("-------------Finished deleting resources.----------------");
     }
 
+    @Data
+    @ToString
```
stair-aws, line 262: `@Data` provides `@ToString`  This is redundant.

ref: https://projectlombok.org/features/Data

amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
```diff
@@ -220,4 +258,53 @@ private void deleteResources(StreamExistenceManager streamExistenceManager, Leas
...
+    @AllArgsConstructor
+    public static class StreamScaler implements Runnable {
```
stair-aws, line 264: Must this be `public` or can it be `private`?

Consider reading https://thefinestartist.com/effective-java/13

amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
```diff
@@ -220,4 +258,53 @@ private void deleteResources(StreamExistenceManager streamExistenceManager, Leas
...
+        private final KinesisAsyncClient client;
+        private final String streamName;
+        private final Queue<ReshardOptions> scalingFactor;
+        private int totalRotations;
+        private KCLAppConfig testConfig;
```
stair-aws, line 269: +`final`?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
```diff
@@ -220,4 +258,53 @@ private void deleteResources(StreamExistenceManager streamExistenceManager, Leas
...
+
+        @Override
+        public void run() {
```
stair-aws, line 272: This is not synchronized, it's a mutable class, and it's scheduled for execution. The only saving grace is the Executor is _currently_ single-threaded, which is not guaranteed to remain that way (e.g., future refactor).

amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
```diff
@@ -115,9 +134,28 @@ private void cleanTestResources(StreamExistenceManager streamExistenceManager, L
...
+            final Queue<ReshardOptions> reshardQueue = new LinkedList<>(Arrays.asList(consumerConfig.getReshardConfig().getReshardingFactorCycle()));
+            int totalRotations = reshardQueue.size() * (consumerConfig.getReshardConfig().getNumReshardCycles() - 1);
+
+            final StreamScaler s = new StreamScaler(kinesisClient, consumerConfig.getStreamName(), reshardQueue, totalRotations, consumerConfig);
+
+            Runnable task1 = () -> {
+                log.info("----------------------Starting new reshard------------------------------");
+                s.run();
+            };
+
```
stair-aws, line 152: Nit: redundant w/ `log.info` in `StreamScaler#run()`

This whole block of code could be removed.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
```diff
@@ -115,9 +134,28 @@ private void cleanTestResources(StreamExistenceManager streamExistenceManager, L
...
+            // Split shard
+            producerExecutor.schedule(task1, 2, TimeUnit.MINUTES);
+
+            // Merge shard
+            producerExecutor.schedule(task1, 6, TimeUnit.MINUTES);
```
stair-aws, line 157: Does this assume the contents of the reshard config? What if a future config is merge-split, or split-merge-split?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
```diff
@@ -220,4 +258,53 @@ private void deleteResources(StreamExistenceManager streamExistenceManager, Leas
...
+        private int totalRotations;
+        private KCLAppConfig testConfig;
+
+        @Override
+        public void run() {
+            try {
+                log.info("----------------------------Starting stream scale----------------------");
+                if (!scalingFactor.isEmpty()) {
+                    log.info("Starting stream scaling with params : {}", this);
+                    final DescribeStreamSummaryRequest describeStreamSummaryRequest = DescribeStreamSummaryRequest.builder().streamName(streamName).build();
```
stair-aws, line 277: 1. This does not support stream ARNs. Is stream ARN support coming in a future revision?
2. `streamName` isn't changing. If this is executed multiple times, this DSSR could be created once and reused.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
```diff
@@ -220,4 +258,53 @@ private void deleteResources(StreamExistenceManager streamExistenceManager, Leas
...
+        private final String streamName;
+        private final Queue<ReshardOptions> scalingFactor;
+        private int totalRotations;
+        private KCLAppConfig testConfig;
+
+        @Override
+        public void run() {
+            try {
+                log.info("----------------------------Starting stream scale----------------------");
+                if (!scalingFactor.isEmpty()) {
```
stair-aws, line 275: The `else` condition is a one-liner. The `if` forces an indentation of ~23 LoC.

Suggestion: when pragmatic to do so, avoid deep-nesting. Example:
```
if (scalingFactor.isEmpty()) {
   log.info(...);
   return;
}

// rest of code here, unindented one level
```

---
Also, do as much as safely possible before/outside a `try` to avoid the performance impact. Refactoring the `if` (as above) could be done outside the `try` to return faster.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
```diff
@@ -220,4 +258,53 @@ private void deleteResources(StreamExistenceManager streamExistenceManager, Leas
...
+
+                    int openShardCount = response.streamDescriptionSummary().openShardCount();
+                    int targetShardCount;
+                    if (scalingFactor.peek() == ReshardOptions.SPLIT) {
+                        // Split case: double the number of shards
+                        targetShardCount = (int) (openShardCount * 2.0);
+                    } else {
+                        // Merge case: half the number of shards
+                        targetShardCount = (int) (openShardCount * 0.5);
+                    }
```
stair-aws, line 288: Use a smart enum. This avoids distributing, and potentially duplicating, logic all over the codebase. Example:
```
enum ReshardOptions {
   SPLIT {
      int calculateShardCount(int currentShards) {
         return 2 * currentShards;
      }
   },
   ...

   abstract int calculateShardCount(int currentShards);
}
```

This code then simplifies to:
```
final int targetShardCount = scalingFactor.peek().calculateShardCount(openShardCount);
```

amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
```diff
@@ -220,4 +258,53 @@ private void deleteResources(StreamExistenceManager streamExistenceManager, Leas
...
+                    log.info("Scaling stream {} from {} shards to {} shards w/ scaling factor {}", streamName, openShardCount, targetShardCount, scalingFactor.peek());
+
+                    final UpdateShardCountRequest updateShardCountRequest = UpdateShardCountRequest.builder().streamName(streamName).targetShardCount(targetShardCount).scalingType(ScalingType.UNIFORM_SCALING).build();
+                    final UpdateShardCountResponse shardCountResponse = client.updateShardCount(updateShardCountRequest).get();
+                    log.info("Executed shard scaling request. Response Details : {}", shardCountResponse.toString());
+
+                    if (--totalRotations >= 0) {
+                        scalingFactor.offer(scalingFactor.poll());
```
stair-aws, line 296: FYI: it'd be more performant, and less destructive, to keep track of a pointer/index. Once you hit the end, wraparound to the head if-and-only-if another rotation is required.

This would simplify `Queue->List`, and avoid side-effects to an object **shared** with, but not controlled/encapsulated by, this class.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
mmankika-aws: totalRotations gets updated, but I can change testConfig to be final

amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
```diff
@@ -220,4 +258,53 @@ private void deleteResources(StreamExistenceManager streamExistenceManager, Leas
...
+                    } else {
+                        scalingFactor.remove();
+                    }
+                } else {
+                    log.info("No scaling factor found in queue");
+                }
+            } catch (Exception e) {
```
stair-aws, line 303: Whenever possible, please be more explicit than `Exception`. Catching `Exception|RuntimeException|Throwable` is generally regarded as bad smell.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/application/TestConsumer.java
```diff
@@ -152,6 +183,16 @@ private void startConsumer() {
         this.consumerFuture = consumerExecutor.schedule(scheduler, 0, TimeUnit.SECONDS);
     }
 
+    public void stopProducer() {
```
ZeyuLi-AWS, line 186: Change back to private

ZeyuLi-AWS (changes requested):

ZeyuLi-AWS (approved):

PR #1157, opened 2023-06-27
Fix NPE on graceful shutdown before DDB `LeaseCoordinator` starts.

stair-aws:
*Issue #, if available:*
* https://github.com/awslabs/amazon-kinesis-client/issues/745
* https://github.com/awslabs/amazon-kinesis-client/issues/900

*Description of changes:*
Fix NPE on graceful shutdown before DDB `LeaseCoordinator` has fully started.

`DynamoDBLeaseCoordinator` still has issues since it doesn't use a FSM. For example, initialization/startup could happen in parallel, or immediately following, a graceful shutdown. However, this PR provides a bias-for-action fix of the NPE.

This PR supplants https://github.com/awslabs/amazon-kinesis-client/pull/901

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseCoordinator.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseCoordinator.java
@@ -329,8 +329,9 @@ public class DynamoDBLeaseCoordinator implements LeaseCoordinator {
 
     @Override
     public void stopLeaseTaker() {
-        takerFuture.cancel(false);
-
+        if (takerFuture != null) {
+            takerFuture.cancel(false);
+        }
     }
 
     @Override
--- a/amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseCoordinatorTest.java
+++ b/amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseCoordinatorTest.java
@@ -1,6 +1,5 @@
 package software.amazon.kinesis.leases.dynamodb;
 
-import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -12,7 +11,7 @@ import software.amazon.kinesis.metrics.MetricsFactory;
 
 import java.util.UUID;
 
-import static org.mockito.Mockito.times;
+import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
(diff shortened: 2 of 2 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseCoordinator.java
```diff
@@ -329,8 +329,9 @@ public void stop() {
 
     @Override
     public void stopLeaseTaker() {
-        takerFuture.cancel(false);
-
+        if (takerFuture != null) {
+            takerFuture.cancel(false);
```
brendan-p-lynch, line 333: In handling this null case is there any chance that takerFuture is assigned to Null. Do we need to mark this somewhere?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseCoordinatorTest.java
```diff
@@ -51,17 +50,34 @@ public void testInitialize_tableCreationSucceeds() throws Exception {
...
+     * Validates a {@link NullPointerException} is not thrown when the lease taker
+     * is stopped before it starts/exists.
+     *
+     * @see <a href="https://github.com/awslabs/amazon-kinesis-client/issues/745">issue #745</a>
+     * @see <a href="https://github.com/awslabs/amazon-kinesis-client/issues/900">issue #900</a>
+     */
+    @Test
+    public void testStopLeaseTakerBeforeStart() {
+        leaseCoordinator.stopLeaseTaker();
+        assertTrue(leaseCoordinator.getAssignments().isEmpty());
```
brendan-p-lynch, line 80: I believe this test is checking that assignments are empty to show that it is before the LeaseTaker Starts. If this is the case, shouldn't we verify that it is empty before calling stopLeaseTaker?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseCoordinatorTest.java
```diff
@@ -51,17 +50,34 @@ public void testInitialize_tableCreationSucceeds() throws Exception {
...
+    }
+
```
brendan-p-lynch, line 82: style: remove the extra line

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseCoordinator.java
stair-aws: > is there any chance that takerFuture is assigned to Null.

Yes, `takerFuture=null` at declaration and remains so until reassigned @ https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseCoordinator.java#L236-L239  This is currently the only assignment, so it will not revert to `null`.

> Do we need to mark this somewhere?

Would you kindly clarify what is being asked?

aravindakidambi: Trying to understand the RCA for the bugs, first one says if scheduler is shutdown too fast after started, I dont think thats possible because start and shutdown are synchronized on the scheduler object. But its possible that if shutdown happened first before start could even run, but then we dont have the right behavior either, ideally if shutdown ran first and then start runs, we should disallow the start, so im thinking maybe we want to keep track of a flag for started and shutdown and only do shutdown processing if it was started and dont allow a start if it was already shutdown?

I dint quite understand the second linked issue from a quick look.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseCoordinatorTest.java
stair-aws: > this test is checking that assignments are empty to show that it is before the LeaseTaker Starts

Not exactly. For one referenced issue, `getAssignments()` [happens immediately-after](https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java#L765-L767) `stopLeaseTaker()`. This check ensures that we don't solve one NPE only to expose/throw another.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseCoordinator.java
stair-aws: > [aravinda] I think maybe thats what you called out in your description but a bias-for-action fix is to address NPE?

Yeah, replication paths weren't provided in the issues but the stacktraces consistently implicated this line.

I agree that we _should_ "disallow the start", and other more defensive/correct behaviors. However, time is a luxury and this bias-for-action PR is better than letting this linger another ~2.5yrs.

aravindakidambi: I agree with bias-for-action if it fixes the issues referenced, but maybe they dont see the NPE with this fix, but still their application wont behave as they want which I assume is, cancel the start and shutdown gracefully? So is there value in doing this, i.e. if we have correct guarantees about only shutdown if started and only start if not already shutdown then this null check wont even be needed? Just a thought, please let me know what you think.

brendan-p-lynch: I was just asking if for some reason the function returned null do we need to log this for future reference?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseCoordinatorTest.java
brendan-p-lynch: I see. Perfect.

brendan-p-lynch (approved):

PR #1218, opened 2023-10-24
Added doc for leases and the lease lifecycle.

stair-aws:
*Issue #, if available:*
N/A

*Description of changes:*
Added doc for leases and the lease lifecycle.

Rendered @ https://github.com/stair-aws/amazon-kinesis-client/blob/lease-doc/docs/lease-lifecycle.md

edit: fixed rendered link since the file was renamed.


docs/leases-and-lease-lifecycle.md
```diff
@@ -0,0 +1,154 @@
...
+](images/lease-shard-sync.png)
+
+For convenience, links to code:
+* `Scheduler`: [implementation][scheduler]
+* `LeaseCoordinator`: [interface][lease-coordinator], [implementation][lease-coordinator-impl]
+* `PeriodicShardSyncManager`: [implementation][periodic-shard-sync-manager]
+* `ShardSyncTask`: [interface][consumer-task], [implementation][consumer-task-impl]
+* `LeaseRefresher`: [interface][lease-refresher], [implementation][lease-refresher-impl]
+* `LeaseSynchronizer`: [implementation][non-empty-lease-table-synchronizer]
+* `HierarchicalShardSyner`: [implementation][hierarchical-shard-syncer]
```
stair-aws, line 81: Typo: `s/Syner/Syncer/`

Will fix when addressing first round of comments.

docs/leases-and-lease-lifecycle.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Lease Table
+
+To persist metadata about lease state (e.g., last read checkpoint, current assigned worker), KCL creates a lease table in [DynamoDB][dynamodb].
+Each KCL application will have its own distinct lease table that transcludes the application name.
+More information, including schema, is provided at [KCL LeaseTable][kcl-leasetable].
+
+## Lease Assignment
+
+The "life" of a lease closely mirrors the duration that a shard is being processed.
```
brendan-p-lynch, line 22: Are there any scenarious where the lease isn't directly tied to a single shard (ie a shard without a lease or a lease with more or less than one shard) if no can we clear this up and say for each shard there is one and only one lease. Closely mirrors implies that there may be some difference to me.

brendan-p-lynch: This is answered later in document

docs/leases-and-lease-lifecycle.md
```diff
@@ -0,0 +1,154 @@
...
+Per the diagram (above), this can occur when there are stream mutations propagating through KCL.
+For example, a 10-shard stream that is split on every shard may temporarily have up-to 30 leases: 10 original + 20 children.
+
+N.B. Leases are uniquely identified by their `leaseKey` which looks vastly different than `lease_X`.
+For details on the `leaseKey` format, please see [KCL LeaseTable][kcl-leasetable].
+
+## Lease Lifecycle
+
+Leases follow a relatively simple, progressive state machine:
+`DISCOVERY -> CREATION -> PROCESSING -> SHARD_END -> DELETION`
```
brendan-p-lynch, line 43: A quick side not that this is a very effective way of thinking, and I do think we should eventually codify it into the system.

docs/leases-and-lease-lifecycle.md
```diff
@@ -0,0 +1,154 @@
...
+1. `DELETION`: Since there are no more records to process, KCL will delete the lease from the lease table.
+   * Deletion will only be triggered after all parents of a child shard have converged to `SHARD_END`.
+   Convergence is required to preserve ordering of records between parent-child relationships.
+   * Deletion is configurable yet recommended to minimize I/O of lease table scans.
+
+### Lease Syncing
+
+Lease syncing is a complex responsibility owned by the "leader" host in a KCL application.
+By invoking the [ListShards API][list-shards], KCL will identify the shards for the configured stream(s).
+This process is scheduled at a configurable interval so KCL can self-identify new shards introduced via stream mutations.
```
brendan-p-lynch, line 67: Can we name the configuration here to piont to the right place for customers

docs/plantuml/lease-shard-sync.puml
```diff
@@ -0,0 +1,63 @@
...
+  PSS->PSS: schedule self every\n  leasesRecoveryAuditorExecutionFrequencyMillis
+end
+
+loop every leasesRecoveryAuditorExecutionFrequencyMillis
+  opt if worker is not leader
+    PSS->PSS: go back to sleep
+  end
+
+  PSS->PSS: runShardSync()
+  opt if not required to sync shards
```
brendan-p-lynch, line 34: Why would a leader not be required to sync shards? Is this if there are no shards, and if so can we say that explicitly

docs/leases-and-lease-lifecycle.md
```diff
@@ -0,0 +1,154 @@
...
+
+![Abridged sequence diagram of the Shard Sync process.
+Listed participants are the Scheduler, LeaseCoordinator, PeriodicShardSyncManager, ShardSyncTask,
+Lease Table (DDB), LeaseRefresher, LeaseSynchronizer, HierarchicalShardSyncer, and ShardDetector.
+](images/lease-shard-sync.png)
```
brendan-p-lynch, line 72: This image is hard to read and small. I don't know if we can increase text size. Also it introduces a lot of new terms that the customer maynot be familiar with

docs/leases-and-lease-lifecycle.md
```diff
@@ -0,0 +1,154 @@
...
+Leases are created in, and leveraged by, KCL to manage a stream's shards.
+This document should help provide insights into the lease lifecycle.
+
+**Nota bene (N.B.):** while actual shard ids are formatted like `shardId-000000000042`, this document uses `shardId[_-]42` for concision.
+
+## Leases
+
+In KCL, a lease provides a temporal assignment between one Kinesis shard and an assigned worker.
+Leases are persistent for the duration of shard processing (detailed later).
+However, lease assignment is transient -- leases may be "stolen" by other workers in the same KCL application.
```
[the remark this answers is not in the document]
stair-aws: > because the assignment has a consistent logic

Pedantically, yes and no. "Yes", the programmed logic doesn't change, yet "no" in that X leases divvied up between Y workers will not produce consistent assignments due to intentional shuffling by KCL.

Per offline consensus, will reword to:
> However, the worker that is processing a lease may change since leases may be "stolen" by other workers in the same KCL application.

docs/leases-and-lease-lifecycle.md
stair-aws: No objection from me! Resolving this conversation as that's outside the scope of this PR. :smile:

docs/plantuml/lease-shard-sync.puml
stair-aws: Some examples:
1. KCL is a multi-threaded app, and the stream could be purged at some time before, or during, the `runShardSync()`
2. the known shard list is complete (i.e., no holes in the hash range)
3. the known shard list is incomplete, but the holes don't cross the threshold to trigger a sync
4. some future edit that may add, or remove, another condition that affects sync logic

> if so can we say that explicitly

The more specific/nuanced documentation is, the more likely it will suffer bitrot with an implementation change. We should balance documentation to convey ideas vs. repacking the implementation as English.

Update pending to provide some additional details, yet not a full enumeration.

docs/leases-and-lease-lifecycle.md
stair-aws: > This image is hard to read and small.

Agreed, and this is after creating the abridged version that omits several classes/calls. It's likely a safe assumption that KCL users know how to click-into an image, or zoom-in via their browsers.

> it introduces a lot of new terms that the customer maynot be familiar with

Anything specific to call out? Code links are provided immediately after so any reader can dive deeper.

docs/leases-and-lease-lifecycle.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Lease Table
+
+To persist metadata about lease state (e.g., last read checkpoint, current assigned worker), KCL creates a lease table in [DynamoDB][dynamodb].
```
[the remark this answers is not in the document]
stair-aws: This iteration doesn't go into checkpoint behavior. Please do not hesitate to contribute!

docs/leases-and-lease-lifecycle.md
```diff
@@ -0,0 +1,166 @@
...
+   * Stream mutations (i.e., split, merge) that create child shards.
+   * In multi-stream mode, dynamic discovery of a new stream.
+1. `CREATION`: Leases are created 1:1 for each discovered shard, and initialized at the configured initial position.
+   * Leases are only created if they are eligible for processing.
+   For example, child shards will not have leases created until its parent(s) have reached `SHARD_END`.
+1. `PROCESSING`: Leases are processed, and continually updated with new checkpoints.
+   * In general, leases spend the majority of their life in this state.
+1. `SHARD_END`: The associated shard is `SHARD_END` and all records have been processed by KCL for the shard.
+1. `DELETION`: Since there are no more records to process, KCL will delete the lease from the lease table.
+   * Deletion will only be triggered after all parents of a child shard have converged to `SHARD_END`.
```
akidambisrinivasan, line 59: Its not clear which is getting deleted here from reading the sentence. Maybe we should say a lease is deleted only after the child shard has checkpointed atleast once. So its not just converging to SHARD_END along with other co-parent, but also that the child lease must have started processing and checkpoined atleast once.

docs/leases-and-lease-lifecycle.md
```diff
@@ -0,0 +1,166 @@
...
+`DISCOVERY -> CREATION -> PROCESSING -> SHARD_END -> DELETION`
+
+Excluding `SHARD_END`, these phases are illustrative of KCL logic and are not explicitly codified.
+
+1. `DISCOVERY`: KCL [shard-syncing](#lease-syncing) identifies new shards.
+Discovered shards may result from:
+   * First time starting KCL with an empty lease table.
+   * Stream mutations (i.e., split, merge) that create child shards.
+   * In multi-stream mode, dynamic discovery of a new stream.
+1. `CREATION`: Leases are created 1:1 for each discovered shard, and initialized at the configured initial position.
```
akidambisrinivasan, line 52: Only for root leases they are created with configured initial position, all other leases for mutations of shards have TRIM_HORIZON ?

docs/leases-and-lease-lifecycle.md
```diff
@@ -0,0 +1,166 @@
...
+1. `DELETION`: Since there are no more records to process, KCL will delete the lease from the lease table.
+   * Deletion will only be triggered after all parents of a child shard have converged to `SHARD_END`.
+   Convergence is required to preserve ordering of records between parent-child relationships.
+   * Deletion is configurable yet recommended to minimize I/O of lease table scans.
+
+### Lease Syncing
+
+Lease syncing is a complex responsibility owned by the "leader" host in a KCL application.
+By invoking the [ListShards API][list-shards], KCL will identify the shards for the configured stream(s).
+This process is scheduled at a
```
akidambisrinivasan, line 67: You have a section on this below, but I would recommend to add it here so it doesnt seem like ListShards happens periodically but only when the auditor determines leases are missing, othterwise childshard leases are created when parent reaches SHARD_END, or something along those lines ?

docs/leases-and-lease-lifecycle.md
```diff
@@ -0,0 +1,166 @@
...
+
+Leases are stolen if-and-only-if there were zero expired leases and the looking-to-steal-worker desires more leases.
+Stolen leases are randomly selected from whichever worker has the most leases.
+The number of leases to steal on each loop is configured via `maxLeasesToStealAtOneTime`.
+
+Customers should consider the following trade-offs when configuring the lease-taking cadence:
+1. `LeaseRefresher` invokes a DDB `scan` against the lease table which has a cost proportional to the number of leases.
+1. Frequent balancing may cause high lease turn-over which incurs DDB `write` costs, and potentially redundant work for stolen leases.
+1. High `maxLeasesToStealAtOneTime` may cause churn.
+    * For example, worker `B` steals multiple leases from worker `A` creating a numerical imbalance.
```
akidambisrinivasan, line 138: It always only steals to balance, but wont steal max if max is not needed to balance ?

docs/leases-and-lease-lifecycle.md
stair-aws: To avoid readers incorrectly assuming that `ListShards` is invoked _every_ interval, rewording (per offline consensus):
> This process is scheduled at a configurable interval so KCL can self-identify new shards introduced via stream mutations.

... to:
> This process is scheduled at a configurable interval so KCL can decide whether it should query for potentially-new shards.

docs/leases-and-lease-lifecycle.md
stair-aws: Removed. If applicable, we can reintroduce this language later w/ a better example.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/exceptions/LeasePendingDeletion.java
```diff
@@ -32,17 +32,18 @@
...
+@EqualsAndHashCode
 @Value
-@EqualsAndHashCode(exclude = {"queueEntryTime"})
 public class LeasePendingDeletion {
-    private final StreamIdentifier streamIdentifier;
-    private final Lease lease;
-    private final ShardInfo shardInfo;
-    private final ShardDetector shardDetector;
+
+    StreamIdentifier streamIdentifier;
```
akidambisrinivasan, line 39: Are these changes intended ?

stair-aws: Yes. Per Lombok [@Value](https://projectlombok.org/features/Value):
> all fields are made private and final by default, and setters are not generated

Explicit `private final` are unnecessary/redundant, and get flagged by IntelliJ as such.

akidambisrinivasan (dismissed):

spaethp (approved):

spaethp (approved):

spaethp (approved):

akidambisrinivasan (approved):

Issue #999, opened 2022-09-19
Confusing way of setting idleTimeBetweenReadsInMillis in PollingConfig


harodani:
I was trying to set idleTimeBetweenReadsInMillis to increase the time between polls. PollingConfig has fluent setters via Lombok, meaning there is a method

`PollingConfig idleTimeBetweenReadsInMillis(long value)`

However this value is ignored because there is another method in the class 

`public void setIdleTimeBetweenReadsInMillis(long idleTimeBetweenReadsInMillis)`

This one does the same but also changes the value of a class boolean field called `usePollingConfigIdleTimeValue`. That it used afterwards in `retrievalFactory`

I think is very confusing having this two methods and one of them does not provide the desire functionality. 

Thanks!

akidambisrinivasan:
Merged fix should address the issue. The `setIdleTimeBetweenReadsInMillis` API has been deprecated in favor of the fluent API which ensures the value is correctly configured

Issue #950, opened 2022-05-12
shardConsumerDispatchPollIntervalMillis for MultiLangDaemon


richardwu:
It doesn't seem like `CoordinatorConfig#shardConsumerDispatchPollIntervalMillis` is an option for MultiLangDaemon in the properties file, since `idleTimeBetweenReadsInMillis` was removed in v2.x: https://github.com/awslabs/amazon-kinesis-client/issues/515.

akidambisrinivasan:
With the merged fix, for MultiLangDaemon in 2.x (once the latest release is out), please use the following in properties file to set the value for idleTimeBetweenReadsInMillis:

idleTimeBetweenReadsInMillis = 10000 # 10 seconds

Issue #515, opened 2019-03-13
idleTimeBetweenReadsInMillis does not seem to be respected


elrob:
- With default configuration below except for `idleTimeBetweenReadsInMillis`, it appears to not be respected.
- Even with idle time set to 10 seconds below, I see that the `ShardRecordProcessor` is being called multiple times within a 10 second period:
    ```
    2019.03.13 05:51:31 [ShardRecordProcessor-0001] INFO  StreamRecordProcessor ...
    2019.03.13 05:51:36 [ShardRecordProcessor-0001] INFO  StreamRecordProcessor ...
    2019.03.13 05:51:37 [ShardRecordProcessor-0001] INFO  StreamRecordProcessor ...
    ```
- When I attempt to follow the source code through, it appears as though the value is never read from `PollingConfig`?

```java
        KinesisAsyncClient kinesisClient = KinesisAsyncClient.builder().build();
        DynamoDbAsyncClient dynamoClient = DynamoDbAsyncClient.builder().build();
        CloudWatchAsyncClient cloudWatchClient = CloudWatchAsyncClient.builder().build();
        ConfigsBuilder configsBuilder = new ConfigsBuilder(streamName, applicationName, kinesisClient, dynamoClient, cloudWatchClient, UUID.randomUUID().toString(), streamRecordProcessorFactory);
        Scheduler scheduler = new Scheduler(
                configsBuilder.checkpointConfig(),
                configsBuilder.coordinatorConfig(),
                configsBuilder.leaseManagementConfig(),
                configsBuilder.lifecycleConfig(),
                configsBuilder.metricsConfig(),
                configsBuilder.processorConfig(),
                configsBuilder.retrievalConfig().retrievalSpecificConfig(
                        new PollingConfig(streamName, kinesisClient)
                                .idleTimeBetweenReadsInMillis(Duration.ofSeconds(10).toMillis())
                )
        );
```

Version:
```xml
        <dependency>
            <groupId>software.amazon.kinesis</groupId>
            <artifactId>amazon-kinesis-client</artifactId>
            <version>2.1.2</version>
        </dependency>
```

vtlkvl:
We are observing the same issue.

vtlkvl:
It seems like `pollingConfig.recordsFetcherFactory().idleMillisBetweenCalls(Duration.ofSeconds(10).toMillis())` should work.

asanthan-amazon:
It seems the below code should work ideally.

[...]

PR #1230, opened 2023-11-22
Fix an issue in configuring idleTimeBetweenReadsInMillis

akidambisrinivasan:
*Description of changes:*

Fix an issue where the idleTimeBetweenReadInMillis configured via MultiLangDaemon was not taking effect because it used the auto-generated setter from Lombok to set the configured value, while there is a custom setter that must be invoked to set the value correctly.

There is also a general confusion between using Lombok's fluent setter vs custom setter in java.

Unifying the approach to use the custom Lombok-fluent-style setter and deprecating the previously added setIdleTimeBetweenReadsInMillis

Correct way to configure idleTimeBetweenReadsInMillis for MultiLang is to add this in the properties file:
idleTimeBetweenReadsInMillis = 10000 # 10 seconds

Correct way to configure for java:
configsBuilder.retrievalConfig().retrievalSpecificConfig(
    new PollingConfig(streamName, kinesisClient)
        .idleTimeBetweenReadsInMillis(Duration.ofSeconds(10).toMillis())

*Issue # #999, #950, #515*

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
@@ -18,6 +18,8 @@ package software.amazon.kinesis.retrieval.polling;
 import java.time.Duration;
 import java.util.Optional;
 import java.util.function.Function;
+
+import lombok.AccessLevel;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.NonNull;
@@ -86,11 +88,15 @@ public class PollingConfig implements RetrievalSpecificConfig {
      * The value for how long the ShardConsumer should sleep in between calls to
      * {@link KinesisAsyncClient#getRecords(GetRecordsRequest)}.
      *
+     * If this is not set using {@link PollingConfig#idleTimeBetweenReadsInMillis},
+     * it defaults to 1500 ms.
+     *
      * <p>
-     * Default value: 1000L
+     * Default value: 1500L
      * </p>
      */
-    private long idleTimeBetweenReadsInMillis = 1000L;
+    @Setter(AccessLevel.NONE)
+    private long idleTimeBetweenReadsInMillis = 1500L;
 
     /**
      * Time to wait in seconds before the worker retries to get a record.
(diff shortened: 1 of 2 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
```diff
@@ -86,11 +88,15 @@ public PollingConfig(String streamName, KinesisAsyncClient kinesisClient) {
...
+     * it defaults to 1500 ms.
+     *
      * <p>
-     * Default value: 1000L
+     * Default value: 1500L
      * </p>
      */
-    private long idleTimeBetweenReadsInMillis = 1000L;
+    @Setter(AccessLevel.NONE)
+    private long idleTimeBetweenReadsInMillis = 1500L;
```
brendan-p-lynch, line 99: Why is it we are chaniging this default?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
```diff
@@ -119,14 +125,23 @@ public PollingConfig(String streamName, KinesisAsyncClient kinesisClient) {
      */
     private RecordsFetcherFactory recordsFetcherFactory = new SimpleRecordsFetcherFactory();
 
+    /**
+     * @Deprecated Use {@link PollingConfig#idleTimeBetweenReadsInMillis} instead
+     */
+    @Deprecated
+    public void setIdleTimeBetweenReadsInMillis(long idleTimeBetweenReadsInMillis) {
```
brendan-p-lynch, line 132: What is going to be the cost of deprecating this. I know that we like everything to be backward compadible. Do we know how many people use this. What will happen to peoples code as they update. Will this require a major release?

It looks to me that it will just call the value below

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
```diff
@@ -119,14 +125,23 @@ public PollingConfig(String streamName, KinesisAsyncClient kinesisClient) {
...
     /**
      * Set the value for how long the ShardConsumer should sleep in between calls to
      * {@link KinesisAsyncClient#getRecords(GetRecordsRequest)}. If this is not specified here the value provided in
      * {@link RecordsFetcherFactory} will be used.
      */
-    public void setIdleTimeBetweenReadsInMillis(long idleTimeBetweenReadsInMillis) {
+    public PollingConfig idleTimeBetweenReadsInMillis(long idleTimeBetweenReadsInMillis) {
         usePollingConfigIdleTimeValue = true;
         this.idleTimeBetweenReadsInMillis = idleTimeBetweenReadsInMillis;
+        return this;
```
brendan-p-lynch, line 144: Why is it we are now returning the value

brendan-p-lynch (approved):
Pending answers to the below questions it is approved

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
akidambisrinivasan: Because the current default of 1000L is never used, if it is not explicitly configured it used a default defined in an internal API which is 1500L, so I am keeping them consistent without behavior change.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
akidambisrinivasan: Deprecating will print a warning during compilation for customer still using this. The behavior will still be correct because they both do the same thing.

We want to avoid confusion with multiple ways to do the same thing, so this says explicitly that we want to support the lombok-fluent-like API that customers are already used to using over the other setter.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
akidambisrinivasan: This is just the explicit implementation of lombok fluent api implementation. It follows the fluent interface - https://en.wikipedia.org/wiki/Fluent_interface

PR #1260, opened 2024-02-20
Add Deletion protection config

lucienlu-aws:
*Issue #, if available:*

*Description of changes:*
The KCL lease table does not have an option to be created with deletion protection enabled. This commit adds the option to be configured via the LeaseManagementConfig

Sample usage
```
leaseManagementConfig = configsBuilder.leaseManagementConfig()
                    .initialPositionInStream(InitialPositionInStreamExtended.newInitialPosition(config.getKclInitialPosition()))
                    .initialLeaseTableReadCapacity(50).initialLeaseTableWriteCapacity(50).deletionProtectionEnabled(true)
```

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
@@ -195,6 +195,13 @@ public class LeaseManagementConfig {
     private BillingMode billingMode = BillingMode.PAY_PER_REQUEST;
 
     /**
+     * Whether to enabled deletion protection on the DyanmoDB lease table created by KCL.
+     *
+     * <p>Default value: false
+     */
+    private boolean deletionProtectionEnabled = false;
+
+    /**
      * The list of tags to be applied to the DynamoDB table created for lease management.
      *
      * <p>Default value: {@link DefaultSdkAutoConstructList}
@@ -382,6 +389,7 @@ public class LeaseManagementConfig {
                     tableCreatorCallback(),
                     dynamoDbRequestTimeout(),
                     billingMode(),
+                    deletionProtectionEnabled(),
                     tags(),
                     leaseSerializer,
                     customShardDetectorProvider(),
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
@@ -195,11 +195,11 @@ public class LeaseManagementConfig {
     private BillingMode billingMode = BillingMode.PAY_PER_REQUEST;
 
     /**
-     * Whether to enabled deletion protection on the DyanmoDB lease table created by KCL.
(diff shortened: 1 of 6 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseManagementFactory.java
```diff
@@ -549,7 +554,8 @@ public DynamoDBLeaseManagementFactory(final KinesisAsyncClient kinesisClient,
             final long listShardsCacheAllowedAgeInSeconds, final int cacheMissWarningModulus,
             final long initialLeaseTableReadCapacity, final long initialLeaseTableWriteCapacity,
             final HierarchicalShardSyncer deprecatedHierarchicalShardSyncer, final TableCreatorCallback tableCreatorCallback,
-            Duration dynamoDbRequestTimeout, BillingMode billingMode, Collection<Tag> tags, LeaseSerializer leaseSerializer,
+            Duration dynamoDbRequestTimeout, BillingMode billingMode, final boolean deletionProtectionEnabled,
```
brendan-p-lynch, line 557: Is there any chance that someone could be using this to crate directly? Would this be code breaking for us to get rid of the builder with just the current ones, and if so I would recommend having another method that includes deletionProtectionEnabled, and then have this one call that with it being false.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
```diff
@@ -194,6 +194,13 @@ public class LeaseManagementConfig {
 
     private BillingMode billingMode = BillingMode.PAY_PER_REQUEST;
 
+    /**
+     * Whether to enabled deletion protection on the DyanmoDB lease table created by KCL.
+     *
+     * <p>Default value: false
+     */
+    private boolean deletionProtectionEnabled = false;
```
brendan-p-lynch, line 202: Can we make this variable explicitly related to DynamoDB so as not to confuse customers, then we can track it easier in case there were ever more dependencies.

brendan-p-lynch (commented):
Added in two comments on naming conventions, otherwise it looks good.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
lucienlu-aws: Renamed variable

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseManagementFactory.java
lucienlu-aws: This class is annotated with `@KinesisClientInternalApi` which means `Any class/method/variable marked with this annotation is subject to breaking changes between minor releases.`

brendan-p-lynch (approved):

akidambisrinivasan (approved):

PR #1270, opened 2024-03-07
add functionality to retry an InvalidArgumentException

vincentvilo-aws:
*Issue #, if available:* N/A

*Description of changes:* Adding functionality to retry an InvalidArgumentException with a new iterator as opposed to retrying with the same iterator. 

### DIFF
--- a/amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/coordinator/KinesisClientLibConfiguration.java
+++ b/amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/coordinator/KinesisClientLibConfiguration.java
@@ -978,7 +978,8 @@ public class KinesisClientLibConfiguration {
     public KinesisClientLibConfiguration withMaxRecords(int maxRecords) {
         checkIsValuePositive("MaxRecords", (long) maxRecords);
         if (maxRecords > DEFAULT_MAX_RECORDS) {
-            throw new IllegalArgumentException("maxRecords must be less than or equal to 10000, but current value is " + maxRecords);
+            throw new IllegalArgumentException(
+                    "maxRecords must be less than or equal to " + DEFAULT_MAX_RECORDS + " but current value is " + maxRecords);
         }
         this.maxRecords = maxRecords;
         return this;
--- a/amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/coordinator/KinesisClientLibConfiguration.java
+++ b/amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/coordinator/KinesisClientLibConfiguration.java
@@ -977,6 +977,9 @@ public class KinesisClientLibConfiguration {
      */
     public KinesisClientLibConfiguration withMaxRecords(int maxRecords) {
         checkIsValuePositive("MaxRecords", (long) maxRecords);
+        if (maxRecords > DEFAULT_MAX_RECORDS) {
+            throw new IllegalArgumentException("maxRecords must be less than or equal to 10000, but current value is " + maxRecords);
+        }
         this.maxRecords = maxRecords;
         return this;
     }
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
@@ -140,6 +140,7 @@ public class PollingConfig implements RetrievalSpecificConfig {
         if (usePollingConfigIdleTimeValue) {
             recordsFetcherFactory.idleMillisBetweenCalls(idleTimeBetweenReadsInMillis);
         }
+        validateMaxRecords();
         return new SynchronousBlockingRetrievalFactory(streamName(), kinesisClient(), recordsFetcherFactory,
                 maxRecords(), kinesisRequestTimeout, dataFetcherProvider);
     }
@@ -151,6 +152,12 @@ public class PollingConfig implements RetrievalSpecificConfig {
                 throw new IllegalArgumentException(
                         "PollingConfig must not have streamName configured in multi-stream mode");
             }
+        }
+    }
+
+    private void validateMaxRecords() {
+        if (maxRecords() > 10000) {
+            throw new IllegalArgumentException("maxRecords must be less than or equal to 10000, but current value is " + maxRecords());
         }
     }
 }
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
@@ -40,6 +40,8 @@ public class PollingConfig implements RetrievalSpecificConfig {
 
     public static final Duration DEFAULT_REQUEST_TIMEOUT = Duration.ofSeconds(30);
 
+    public static final int DEFAULT_MAX_RECORDS = 10000;
+
     /**
      * Configurable functional interface to override the existing DataFetcher.
      */
@@ -71,7 +73,7 @@ public class PollingConfig implements RetrievalSpecificConfig {
      * Default value: 10000
      * </p>
      */
-    private int maxRecords = 10000;
+    private int maxRecords = DEFAULT_MAX_RECORDS;
 
     /**
      * @param streamName    Name of Kinesis stream.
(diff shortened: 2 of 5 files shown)


amazon-kinesis-client-multilang/src/main/java/software/amazon/kinesis/coordinator/KinesisClientLibConfiguration.java
```diff
@@ -977,6 +977,9 @@ public KinesisClientLibConfiguration withShardSyncIntervalMillis(long shardSyncI
      */
     public KinesisClientLibConfiguration withMaxRecords(int maxRecords) {
         checkIsValuePositive("MaxRecords", (long) maxRecords);
+        if (maxRecords > DEFAULT_MAX_RECORDS) {
+            throw new IllegalArgumentException("maxRecords must be less than or equal to 10000, but current value is " + maxRecords);
```
pelaezryan, line 981: We should output DEFAULT_MAX_RECORDS instead of 10000. Reason being is in that value were to ever change, this log statement would be incorrect

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
```diff
@@ -153,4 +154,10 @@ public void validateState(final boolean isMultiStream) {
             }
         }
     }
+
+    private void validateMaxRecords() {
+        if (maxRecords() > 10000) {
```
pelaezryan, line 159: We should use DEFAULT_MAX_RECORDS once again here so it applies everywhere. Otherwise we will have to manually update all these places

pelaezryan: Also DRY (don't repeat yourself). This function already exists. Move this to a helper function if necessary.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
```diff
@@ -140,6 +140,7 @@ public RetrievalFactory retrievalFactory() {
         if (usePollingConfigIdleTimeValue) {
             recordsFetcherFactory.idleMillisBetweenCalls(idleTimeBetweenReadsInMillis);
         }
+        validateMaxRecords();
```
akidambisrinivasan, line 143: You can override the setter for the maxRecords and add validation there.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
```diff
@@ -511,6 +511,9 @@ private void makeRetrievalAttempt() {
                 } catch (InterruptedException e) {
                     Thread.currentThread().interrupt();
                     log.info("{} :  Thread was interrupted, indicating shutdown was called on the cache.", streamAndShardId);
+                } catch (InvalidArgumentException e) {
```
akidambisrinivasan, line 514: This needs to be before RetryableRetrievalException or within that check for IAE, here it wont be executed. You should add a test to make sure this works.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
```diff
@@ -511,6 +511,9 @@ private void makeRetrievalAttempt() {
...
+                    log.info("{} :  records threw InvalidArgumentException - retrying with new iterator", streamAndShardId, e);
```
akidambisrinivasan, line 515: nit: reword to say iterator will be refreshed before retrying

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
akidambisrinivasan: Also add tests for these validations. and also stream arn validation if it doesnt exist so it wont break in future.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
akidambisrinivasan: Im surprised the test passed I think there are 2 different InvalidArgumentException, can we make sure what is thrown by SDK and use that, maybe your code is correct, but lets make sure.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
vincentvilo-aws: Adding a test to validate maxRecords in the next commit. As for the streamArn, they are validated inS: https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/test/java/software/amazon/kinesis/common/StreamIdentifierTest.java#L63

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
```diff
@@ -71,7 +73,7 @@ public PollingConfig(KinesisAsyncClient kinesisClient) {
      * Default value: 10000
      * </p>
      */
-    private int maxRecords = 10000;
+    private int maxRecords = DEFAULT_MAX_RECORDS;
```
pelaezryan, line 76: This value is now redundant and no longer applicable.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
```diff
@@ -40,6 +40,8 @@ public class PollingConfig implements RetrievalSpecificConfig {
 
     public static final Duration DEFAULT_REQUEST_TIMEOUT = Duration.ofSeconds(30);
 
+    public static final int DEFAULT_MAX_RECORDS = 10000;
```
pelaezryan, line 43: If this is going to be public and static, we can share this value in the multilang as well so that there is one central source of truth.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
```diff
@@ -129,6 +131,14 @@ public void setIdleTimeBetweenReadsInMillis(long idleTimeBetweenReadsInMillis) {
         this.idleTimeBetweenReadsInMillis = idleTimeBetweenReadsInMillis;
     }
 
+    public void maxRecords(int maxRecords) {
```
pelaezryan, line 134: nit: setMaxRecords(...);

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
pelaezryan: as a follow up, it may be worth checking to see how many of these types of values we have "copy pasted" in multiple locations and see if its worth centralizing them.

pelaezryan (approved):

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
vincentvilo-aws: The class has an annotation `@Accessors(fluent = true)`, so the getters and setters do not have the "get" and "set" prefix ([ref](https://projectlombok.org/features/experimental/Accessors)).

akidambisrinivasan (approved):

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PollingConfig.java
etspaceman: @vincentvilo-aws this broke backwards compatibility. The method needed to return PollingConfig. See https://github.com/awslabs/amazon-kinesis-client/issues/1274 and https://github.com/awslabs/amazon-kinesis-client/pull/1275

vincentvilo-aws: Thanks for catching this. I missed that the `chain` boolean by default is set to true when using the `@Accessors` annotation. I'll approve #1275 once the checks pass.

PR #1276, opened 2024-03-16
fix: does not request shutdown for already shutdown consumer

### LINKED ISSUE #1251: graceful shutdown of MultiLangDaemon worker that is assigned for completed shards is always timeout
I found that graceful shutdown of MultiLangDaemon worker that is assigned for completed shards is always timeout. 

```java
2024-01-31 00:43:51,722 [Thread-1] ERROR s.a.k.multilang.MultiLangDaemon [NONE] - Encountered an error during shutdown.
java.util.concurrent.TimeoutException: null
 	at java.base/java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1892)
 	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2027)
 	at software.amazon.kinesis.multilang.MultiLangDaemon.lambda$setupShutdownHook$0(MultiLangDaemon.java:183)
 	at java.base/java.lang.Thread.run(Thread.java:829)
```

I read codes and found that two CountDownLatch variables are used to wait all consumers are shutdown completed in shutdown process and both value is the number of worker lease.
https://github.com/awslabs/amazon-kinesis-client/blob/v2.5.3/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java#L780-L781
https://github.com/awslabs/amazon-kinesis-client/blob/v2.5.2/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/GracefulShutdownCoordinator.java#L65-L126

If the worker has a lease of completed shard (its checkpoint is `SHARD_END`), the number of worker lease is greater than the number of actual consumer subprocess so two latches never becomes to 0 and always timeout.

Is this expected behavior ?

## Environment
- Python 3.11
- amazon-kclpy 2.1.3 ( kinesis client library: 2.5.2)

## Step to reproduce
To demonstrate this  issue, I created a demo application with localstack: 
https://github.com/nao23/kcl-shutdown-timeout-demo

1. Start application
```shell
$ docker compose build
$ docker compose up app
```

2. Shutdown MultiLangDaemon process
```shell
$ docker compose exec app kill 1
```

nao23:
This patch fixes #1251 .

In graceful shutdown phase, two CountDownLatch variables that value is  the number of worker lease are used to wait all consumers are shutdown completed.

If the worker has a lease of completed shard, the worker requests shutdown for its consumer but the consumer is already shutdown and cannot decrement latches so two latches never becomes to 0 and always timeout.

This patch modifies worker not to request shutdown for already shutdown consumer and only decrement latch.

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
@@ -784,7 +784,7 @@ public class Scheduler implements Runnable {
                         lease, notificationCompleteLatch, shutdownCompleteLatch);
                 ShardInfo shardInfo = DynamoDBLeaseCoordinator.convertLeaseToAssignment(lease);
                 ShardConsumer consumer = shardInfoShardConsumerMap.get(shardInfo);
-                if (consumer != null) {
+                if (consumer != null && !consumer.isShutdown()) {
                     consumer.gracefulShutdown(shutdownNotification);
                 } else {
                     //


vincentvilo-aws:
I'll be closing this PR since this was addressed in the original issue. Please see my comment left on #1251.

PR #1304, opened 2024-04-04
Embed StreamConfig within ShardInfo

furq-aws:
*Issue #, if available:*
N/A.

*Description of changes:*
Add `StreamConfig` as a field within `ShardInfo`.

Previously, `StreamIdentifier` instances were reconstructed using the optional `streamIdentifierSerOpt` field within `ShardInfo` (which has since been removed). 
Now, `ShardInfo` maintains a reference to the `StreamIdentifier` through `StreamConfig`, eliminating the need for reconstruction.

This change encompasses any additional refactoring necessitated by embedding `StreamConfig` within `ShardInfo`. 
For example, there's no longer a need to explicitly provide `initialPositionInStreamExtended` (a field of `StreamConfig`) when `ShardInfo` is already available.


amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/SchedulerTest.java
```diff
@@ -234,33 +249,29 @@ public final void testCreateOrGetShardConsumer() {
...
-
         final Checkpoint firstCheckpoint = new Checkpoint(firstSequenceNumber, null, null);
+        when(checkpoint.getCheckpointObject(eq(TEST_SHARD_ID))).thenReturn(firstCheckpoint);
 
-        when(leaseCoordinator.getCurrentAssignments()).thenReturn(initialShardInfo, firstShardInfo, secondShardInfo);
-        when(checkpoint.getCheckpointObject(eq(shardId))).thenReturn(firstCheckpoint);
+        when(leaseCoordinator.getAssignments()).thenReturn(
+                Stream.of(firstSequenceNumber, secondSequenceNumber, finalSequenceNumber)
+                        .map(SchedulerTest::constructLease)
+                        .collect(Collectors.toList()));
```
furq-aws, line 261: This actually changes the behavior - returns all three leases in each call to `getAssignments`, rather than returning one new lease for each of the three subsequent calls (as intended by the original code); will revert this portion in next push.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/ShardInfo.java
```diff
@@ -18,31 +18,37 @@
...
     private final String concurrencyToken;
     // Sorted list of parent shardIds.
     private final List<String> parentShardIds;
     private final ExtendedSequenceNumber checkpoint;
+    private final StreamConfig streamConfig;
+
+    @Getter(AccessLevel.NONE)
+    private final boolean isMultiStreamMode;
+    @Getter(AccessLevel.NONE)
+    private final String streamIdentifierStr;
```
lucienlu-aws, line 51: Do we need these two member variables if they are just extracted out from streamConfig? I see you are using them in some functions in this class for comparisons but would it be more clean to extract it from streamConfig individually instead?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -952,26 +953,20 @@ protected ShardConsumer buildConsumer(@NonNull final ShardInfo shardInfo,
...
+        final StreamConfig streamConfig = shardInfo.streamConfig();
+        if (!currentStreamConfigMap.containsKey(streamConfig.streamIdentifier())) {
             log.info("Created orphan {}", streamConfig);
         }
-        Validate.notNull(streamConfig, "StreamConfig should not be null");
+        /*
+         * NOTE: RecordsPublisher#createGetRecordsCache(ShardInfo, StreamConfig, MetricsFactory) is deprecated.
+         *  RecordsPublisher#createGetRecordsCache(ShardInfo, MetricsFactory) will be called directly in the future.
+         */
         RecordsPublisher cache = retrievalConfig.retrievalFactory().createGetRecordsCache(shardInfo, streamConfig, metricsFactory);
```
lucienlu-aws, line 964: Is the reason we are not changing this now is that there could be some clients which pass in their own interface of RetrievalFactory and may not implement the new function yet? Wouldn't that still mean its backwards incompatible if we ever change its, so it would have to be a major version upgrade?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/RetrievalFactory.java
```diff
@@ -23,11 +23,31 @@
...
+    }
+
+    /**
+     * Creates a {@link RecordsPublisher} instance to retrieve records for the specified shard.
+     *
+     * @param shardInfo The {@link ShardInfo} representing the shard for which records are to be retrieved.
+     * @param metricsFactory The {@link MetricsFactory} for recording metrics.
+     * @return A {@link RecordsPublisher} instance for retrieving records from the shard.
+     */
     RecordsPublisher createGetRecordsCache(ShardInfo shardInfo, MetricsFactory metricsFactory);
```
lucienlu-aws, line 43: Interesting this was deprecated before and we are undeprecating it. Was there a reason this was deprecated before?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/ShardInfo.java
```diff
@@ -18,31 +18,37 @@
...
 
-    private final Optional<String> streamIdentifierSerOpt;
     private final String shardId;
     private final String concurrencyToken;
     // Sorted list of parent shardIds.
     private final List<String> parentShardIds;
     private final ExtendedSequenceNumber checkpoint;
+    private final StreamConfig streamConfig;
+
+    @Getter(AccessLevel.NONE)
```
akidambisrinivasan, line 48: yes thank you, this is necessary because this is class exposes a public contract which cannot be violated and we should not expose more than what is necessary. But I do agree with lucien that we probably dont need them and get it from streamConfig when needed.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ProcessTask.java
```diff
@@ -120,8 +119,7 @@ public TaskResult call() {
          */
         final MetricsScope appScope = MetricsUtil.createMetricsWithOperation(metricsFactory, APPLICATION_TRACKER_OPERATION);
         final MetricsScope shardScope = MetricsUtil.createMetricsWithOperation(metricsFactory, PROCESS_TASK_OPERATION);
-        shardInfo.streamIdentifierSerOpt()
-                .ifPresent(streamId -> MetricsUtil.addStreamId(shardScope, StreamIdentifier.multiStreamInstance(streamId)));
+        MetricsUtil.addStreamId(shardScope, shardInfo.streamConfig().streamIdentifier());
```
akidambisrinivasan, line 122: Okay we dont want to do this by default, for single stream application, this can change dimension and may break public contract on the metrics exposed by KCL.

Seems like then we need some method from streamIdentifier, you had isMultiStreamInstance. We could use it, but I dint want to expose that method based on the comment I left there. Lets think if there is another way.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ProcessTask.java
```diff
@@ -218,8 +216,7 @@ private void callProcessRecords(ProcessRecordsInput input, List<KinesisClientRec
                 .millisBehindLatest(input.millisBehindLatest()).build();
 
         final MetricsScope scope = MetricsUtil.createMetricsWithOperation(metricsFactory, PROCESS_TASK_OPERATION);
-        shardInfo.streamIdentifierSerOpt()
-                .ifPresent(streamId -> MetricsUtil.addStreamId(scope, StreamIdentifier.multiStreamInstance(streamId)));
+        MetricsUtil.addStreamId(scope, shardInfo.streamConfig().streamIdentifier());
```
akidambisrinivasan, line 219: same here

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/SynchronousBlockingRetrievalFactory.java
```diff
@@ -79,42 +63,27 @@ public SynchronousBlockingRetrievalFactory(String streamName,
...
     private static Function<DataFetcherProviderConfig, DataFetcher> defaultDataFetcherProvider(
             KinesisAsyncClient kinesisClient) {
         return dataFetcherProviderConfig -> new KinesisDataFetcher(kinesisClient, dataFetcherProviderConfig);
     }
 
+    /**
+     * {@inheritDoc}
+     */
     @Override
-    public GetRecordsRetrievalStrategy createGetRecordsRetrievalStrategy(@NonNull final ShardInfo shardInfo,
```
akidambisrinivasan, line 96: :( dont know why this was part of the public interface, thanks for marking it deprecated, its never called from outside and its only called from within the class so it dint need to be part of the public interface, could just have been a private method.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRetrievalFactory.java
```diff
@@ -41,36 +38,25 @@ public class FanOutRetrievalFactory implements RetrievalFactory {
...
+     */
     @Override
     public RecordsPublisher createGetRecordsCache(@NonNull final ShardInfo shardInfo,
-            final StreamConfig streamConfig,
             final MetricsFactory metricsFactory) {
-        final Optional<String> streamIdentifierStr = shardInfo.streamIdentifierSerOpt();
-        if (streamIdentifierStr.isPresent()) {
-            final StreamIdentifier streamIdentifier = StreamIdentifier.multiStreamInstance(streamIdentifierStr.get());
+        final StreamIdentifier streamIdentifier = shardInfo.streamConfig().streamIdentifier();
+        if (streamIdentifier.isMultiStreamInstance()) {
```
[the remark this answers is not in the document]
akidambisrinivasan: Got confirmed its ok to change logs

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -1090,6 +1072,31 @@ private void resetInfoLogging() {
...
+            Validate.isTrue(!isMultiStreamMode, "Should not be in MultiStream Mode");
+            final StreamConfig streamConfig = currentStreamConfigMap.values().iterator().next();
+            Validate.notNull(streamConfig, "StreamConfig should not be null");
+            return streamConfig;
+        }
+
+        final StreamIdentifier streamIdentifier =
+                StreamIdentifier.multiStreamInstance(streamIdentifierSerialization.get());
+        return currentStreamConfigMap.getOrDefault(streamIdentifier, streamTracker.createStreamConfig(streamIdentifier));
+    }
```
nakulj, line 1098:
```suggestion
    private ShardInfo constructShardInfoFromLease(final Lease lease) {
        Optional<MultiStreamLease> msl = Optional.of(lease).filter(l->l instanceof MultiStreamLease).map(l->(MultiStreamLease) l);
        String shardId = msl.map(MultiStreamLease::shardId).orElse(lease.leaseKey());
        StreamConfig streamConfig = msl.map(MultiStreamLease::streamIdentifier).map(streamIdentifierSerialization -> {
            final StreamIdentifier streamIdentifier = StreamIdentifier.multiStreamInstance(streamIdentifierSerialization);
            return currentStreamConfigMap.getOrDefault(streamIdentifier, streamTracker.createStreamConfig(streamIdentifier));
        }).orElseGet(()->{
            Validate.isTrue(!isMultiStreamMode, "Should not be in MultiStream Mode");
            return Optional.ofNullable(currentStreamConfigMap.values().iterator().next()).get();
        });
        return new ShardInfo(
                shardId, lease.concurrencyToken().toString(), lease.parentShardIds(), lease.checkpoint(), streamConfig);
    }
```
This makes it a little more clear that the empty optional represents an uncastable Lease.

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/SchedulerTest.java
nakulj, line 261:
```suggestion
        when(leaseCoordinator.getAssignments()).thenReturn(
            ImmutableList.of(
                constructLease(firstSequenceNumber),
                constructLease(secondSequenceNumber),
                constructLease(finalSequenceNumber)));
```
Avoid control flow in unit tests.

PR #1316, opened 2024-04-17
Create script to check for backwards compatibility

vincentvilo-aws:
*Issue #, if available:* N/A

*Description of changes:* Creating a script to check for backwards compatibility.

The script compares the JAR of the latest release and the JAR of the current version (i.e. the version with any implemented changes). Any packages or functions that were removed from the latest release are outputted and causes the Github action to fail. 


.github/scripts/backwards_compatibility_check.sh
```diff
@@ -0,0 +1,102 @@
...
+
+# Get the JAR with the changes that need to be verified.
+get_current_jar() {
+  mvn -B install -DskipTests
+  CURRENT_VERSION=$(mvn -q  -Dexec.executable=echo -Dexec.args='${project.version}' --non-recursive exec:exec)
+  CURRENT_JAR=$KCL_MAVEN_DIR/$CURRENT_VERSION/amazon-kinesis-client-$CURRENT_VERSION.jar
+}
+
+# Skip classes with the KinesisClientInternalApi annotation. These classes are subject to breaking backwards compatibility.
+is_kinesis_client_internal_api() {
```
lucienlu-aws, line 30: Classes with this annotation should not be breaking backwards compatibility between patches, but can break between minor versions. Is there a way to update the script to reflect that? Ideally what we want to do in classes with this annotation is deprecate a method in a minor release and then in the next minor release we can remove it. Not a big blocker for this PR though

.github/scripts/backwards_compatibility_check.sh
```diff
@@ -0,0 +1,102 @@
...
+LATEST_JAR=""
+CURRENT_VERSION=""
+CURRENT_JAR=""
+
+# Get the JAR from the latest version release on Maven.
+get_latest_jar() {
+  #  clear the directory so that the latest release will be the only version in the Maven directory after running mvn dependency:get
+  rm -rf "$KCL_MAVEN_DIR"
+  mvn -B dependency:get -Dartifact=software.amazon.kinesis:amazon-kinesis-client:LATEST
+  LATEST_VERSION=$(ls "$KCL_MAVEN_DIR" | grep '[0-9].[0-9].[0-9]')
```
lucienlu-aws, line 18: This won't work for double digit versions right? i.e. 2.10.x or 2.5.10

vincentvilo-aws: Changing the expression to `grep -E '[0-9]+.[0-9]+.[0-9]+'` address this.

.github/scripts/backwards_compatibility_check.sh
vincentvilo-aws: I created a function `is_new_minor_release` to check if a minor release is being performed. The check will fail if there are changes in classes with the `@KinesisClientInternalApi` and a minor release is *not* being performed.

lucienlu-aws (approved):

PR #1318, opened 2024-04-25
Internally construct and use stream ARNs for all streams in multi-stream mode

furq-aws:
*Issue #, if available:*
N/A.

*Description of changes:*
Construct stream ARNs (using the Kinesis client's configured region) for all stream configs upon writing to the `currentStreamConfigMap` in multi-stream mode. 
This should ensure that all Kinesis API requests from `KinesisDataFetcher` and `KinesisShardDetector` are provided with the stream ARN in multi-stream applications.

Note: An implication of stream ARNs being used for Kinesis API calls is that the `accountId` provided through the `StreamTracker` must be correct, whereas previously applications may have consumed the streams as long as the provided `streamName` existed in the account that the Kinesis client was configured for.

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/ArnUtil.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/ArnUtil.java
@@ -1,0 +1,32 @@
+package software.amazon.kinesis.common;
+
+import lombok.NonNull;
+import software.amazon.awssdk.arns.Arn;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.kinesis.annotations.KinesisClientInternalApi;
+
+import static software.amazon.awssdk.services.kinesis.KinesisAsyncClient.SERVICE_NAME;
+
+@KinesisClientInternalApi
+public final class ArnUtil {
+    private static final String STREAM_RESOURCE_PREFIX = "stream/";
+
+    /**
+     * Construct a Kinesis stream ARN.
+     *
+     * @param region The region the stream exists in.
+     * @param accountId The account the stream belongs to.
+     * @param streamName The name of the stream.
+     * @return The {@link Arn} of the Kinesis stream.
+     */
+    public static Arn constructStreamArn(
+            @NonNull final Region region, @NonNull final String accountId, @NonNull final String streamName) {
+        return Arn.builder()
+                .partition(region.metadata().partition().id())
+                .service(SERVICE_NAME)
+                .region(region.id())
+                .accountId(accountId)
+                .resource(STREAM_RESOURCE_PREFIX + streamName)
+                .build();
+    }
+}
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
@@ -49,8 +49,11 @@ import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
 import lombok.experimental.Accessors;
 import lombok.extern.slf4j.Slf4j;
+import software.amazon.awssdk.arns.Arn;
+import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.utils.Validate;
 import software.amazon.kinesis.checkpoint.CheckpointConfig;
 import software.amazon.kinesis.checkpoint.ShardRecordProcessorCheckpointer;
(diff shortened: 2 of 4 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -1051,6 +1055,73 @@ private StreamIdentifier getStreamIdentifier(Optional<String> streamIdentifierSt
...
+                        "The provided streamARN " + streamArnOptional.get()
+                                + " does not match the Kinesis client's configured region - " + kinesisRegion);
+                return super.put(streamIdentifier, streamConfig);
+            }
+
+            if (isMultiStreamMode) {
+                return super.put(streamIdentifier, withStreamArn(streamConfig, kinesisRegion));
+            }
+
+            return super.put(streamIdentifier, streamConfig);
```
nakulj, line 1121: `isPresent()` followed by `get` is an anti-pattern.
Instead use `map` or `ifPresent` as appropriate. In this case, something like
```suggestion
            StreamConfig cfg = streamConfig.streamIdentifier().streamArnOptional().map(streamArn -> {
                Validate.isTrue(kinesisRegion.id().equals(streamArn.region().get()),
                        "The provided streamARN " + streamArn +
                                + " does not match the Kinesis client's configured region - " + kinesisRegion);
                return streamConfig;
            }).orElse(isMultiStreamMode ? withStreamArn(streamConfig, kinesisRegion):streamConfig);
            return super.put(streamIdentifier, cfg);

```

amazon-kinesis-client/src/main/java/software/amazon/kinesis/common/ArnUtil.java
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * Construct a Kinesis stream ARN.
+     *
+     * @param region The region the stream exists in.
+     * @param accountId The account the stream belongs to.
+     * @param streamName The name of the stream.
+     * @return The {@link Arn} of the Kinesis stream.
+     */
+    public static Arn constructStreamArn(
+            @NonNull final Region region, final long accountId, @NonNull final String streamName) {
```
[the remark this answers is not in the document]
furq-aws: Using a `long` for the `accountId` can lead to truncation issues for accounts with leading zeros, 
so this is actually a bug I think. 

Thanks for the callout!

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
furq-aws: Updated. Thanks!

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/SchedulerTest.java
```diff
@@ -1080,6 +1097,87 @@ private void mockListLeases(List<StreamConfig> configs) throws ProvisionedThroug
...
+                .contains(streamIdentifierSerializationForOrphan));
+
+        when(leaseCoordinator.getCurrentAssignments()).thenReturn(Collections.singletonList(
+                new ShardInfo(TEST_SHARD_ID, null, null, null, streamIdentifierSerializationForOrphan)));
+        retrievalConfig = new RetrievalConfig(kinesisClient, multiStreamTracker, applicationName)
+                .retrievalFactory(retrievalFactory);
+        scheduler = new Scheduler(checkpointConfig, coordinatorConfig, leaseManagementConfig, lifecycleConfig,
+                metricsConfig, processorConfig, retrievalConfig);
+
+        final Scheduler schedulerSpy = spy(scheduler);
```
akidambisrinivasan, line 1142: Why do you need a spy? can you not invoke runProcessLoop on scheduler?

lucienlu-aws (dismissed):

akidambisrinivasan (approved):

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/SchedulerTest.java
furq-aws: Good callout! Spy is not needed here; must've copied from a previous unit test and forgotten to clean it up.

lucienlu-aws (approved):

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -1051,6 +1055,67 @@ private StreamIdentifier getStreamIdentifier(Optional<String> streamIdentifierSt
...
+         * {@inheritDoc}
+         */
+        @Override
+        public StreamConfig put(
+                @NonNull final StreamIdentifier streamIdentifier, @NonNull final StreamConfig streamConfig) {
+            final Region kinesisRegion = getKinesisRegion();
+
+            return super.put(streamIdentifier, streamConfig.streamIdentifier().streamArnOptional()
+                    .map(streamArn -> {
+                        Validate.isTrue(kinesisRegion.id().equals(streamArn.region().get()),
```
akidambisrinivasan, line 1111: Just realized that this is invoked in checkAndSyncStreamShardsAndLeases when new streams are discovered, when a shard sync task is generated and updates the currentStreamConfigMap. In that case if the validation fails, it can short-circuit the processing because the exception is only caught in runProcessLoop. This can prevent other streams from being processed. Ideally if there is an issue with a stream, its reasonable for Leader to log an error and skip the stream and maybe even emit a metric. But not let processing of other streams be affected. Can you fix it as a follow up ?

PR #1329, opened 2024-05-09
Remove unnecessary lambda

nakulj:
*Description of changes:*

> Returning a lambda from a helper method or saving it in a constant is unnecessary; prefer to implement the functional interface method directly and use a method reference instead.

 https://errorprone.info/bugpattern/UnnecessaryLambda


lucienlu-aws (approved):
LGTM, recommend updating the commit message and PR message to "Remove unnecessary lambda"

PR #1330, opened 2024-05-10
Modify backwards compatibility script to ignore synthetic access methods

vincentvilo-aws:
*Issue #, if available:* N/A

*Description of changes:* Modify backwards compatibility script to ignore synthetic access methods. These access methods are not public and should not affect backwards compatibility. 

### DIFF
--- a/.github/scripts/backwards_compatibility_check.sh
+++ b/.github/scripts/backwards_compatibility_check.sh
@@ -82,6 +82,9 @@ find_removed_methods() {
 
     local removed_methods=$(diff <(echo "$LATEST_METHODS") <(echo "$CURRENT_METHODS") | grep '^<')
 
+    # ignore synthetic access methods - these are not available to users and will not break backwards compatibility
+    removed_methods=$(echo "$removed_methods" | grep -v "access\$[1-4]00")
+
     if [[ "$removed_methods" != "" ]]
     then
       REMOVED_METHODS_FLAG=$TRUE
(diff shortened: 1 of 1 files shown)


.github/scripts/backwards_compatibility_check.sh
```diff
@@ -82,6 +82,9 @@ find_removed_methods() {
 
     local removed_methods=$(diff <(echo "$LATEST_METHODS") <(echo "$CURRENT_METHODS") | grep '^<')
 
+    # ignore synthetic access methods - these are not available to users and will not break backwards compatibility
+    removed_methods=$(echo "$removed_methods" | grep -v "access\$[1-4]00")
```
brendan-p-lynch: IS there a reason we set this and then immdediately change it again. Could we do this in one line?

Also if you can provide why that grep works well that would be great. I don't have an intuition for this line: grep -v "access\$[1-4]00"

vincentvilo-aws: Main reason as to why it's on its own line is just to make it explicit that we are ignoring these types of errors. We can include it in the same line but imo I think it's better to keep it separate and make it clear we're ignoring synthetic methods with the following line. 

What `grep -v <...>` does is filter out any lines that contain the specified pattern. If we find that access$100, access$200, access$300, or access$400 is in the original diff output, we remove the line from the output (i.e. we ignore any changes related to access$[1-4]00). 

FYI the newest commit changes this to ignore access$000 to access$900.

brendan-p-lynch: Ok, it's not super intuitive to me but if it is standard and doesn't need a comment that is ok

brendan-p-lynch (approved):

PR #1340, opened 2024-06-01
Generate wrappers from proto files instead of shipping them directly

nakulj:
*Description of changes:*
- proto wrapper classes are now generated from a `.proto` file instead of being version controlled


brendan-p-lynch (dismissed):

amazon-kinesis-client/src/main/proto/messages.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto2";
```
brendan-p-lynch, line 1: Do we have a mechanism for keeping this in line with the kpl version? I just want to make sure it isn't duplicated somewhere else.

brendan-p-lynch (commented):
Overall I like this approach but just want to make sure it is consistent with what KPL is expecting

amazon-kinesis-client/src/main/proto/messages.proto
nakulj: Ideally, yes- proto messages would be defined in a separate repo that would then be used as a submodule here and elsewhere. But from what I've seen these definitions are duplicated across multiple repos so consolidating them should be a separate effort.

amazon-kinesis-client/src/main/proto/messages.proto
```diff
@@ -0,0 +1,20 @@
...
+package software.amazon.kinesis.retrieval.kpl;
+
+message Tag {
```
akidambisrinivasan, line 4: Is this a subset of messages kcl cares about copied from here - https://github.com/awslabs/amazon-kinesis-producer/blob/master/aws/kinesis/protobuf/messages.proto ?

nakulj: Yes.

nakulj: As I mentioned in another comment, ideally these would be in a shared repo... but this is better than the previous setup I think.

akidambisrinivasan: Can we please add a comment "Copied from amazon-kinesis-producer/aws/kinesis/protobuf/messages.proto with subset of messages that KCL needs" or something along those lines

nakulj: Done

akidambisrinivasan (approved):

PR #1354, opened 2024-06-18
bugfix: only update in memory lease if DDB lease renews without error

lucienlu-aws:
*Issue #, if available:*

*Description of changes:*
There is a bug in DDBLeaseRenewer such that if DDB fails to update, the in memory lease would still be updated. This commit fixes this by reverting the update if the DDB update throws any error

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRenewer.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRenewer.java
@@ -322,6 +322,7 @@ public class DynamoDBLeaseRenewer implements LeaseRenewer {
 
         long startTime = System.currentTimeMillis();
         boolean success = false;
+        Lease authoritativeLeaseCopy = authoritativeLease.copy();
         try {
             log.info("Updating lease from {} to {}", authoritativeLease, lease);
             synchronized (authoritativeLease) {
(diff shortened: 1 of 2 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRenewer.java
```diff
@@ -358,6 +359,10 @@ public boolean updateLease(
                 success = true;
                 return updatedLease;
             }
+        } catch (Exception e) {
```
brendan-p-lynch, line 362: I am a little bit concerned that we catch this on every exception. Is it not possible that there are exceptions that we would not want to revert the lease on? Shouldn't we only catch this if the DDB update fails?

brendan-p-lynch: Also even if this is the case now wouldn't it be better to define them in case another case comes along where the exception woudn't stop the lease from updating.

brendan-p-lynch (commented):
I think this might be too broad and not future proof to exceptions where we wouldn't want to roll back. It seems like there could be a problem where we catch some exception but it did update the table and we roll back in memory anyway.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRenewer.java
lucienlu-aws: I agree with your concern, updated to be only exceptions thrown when failing to update DDB

brendan-p-lynch (approved):
I see that the backward compatibility tests are failing. Approving but just want to make sure we are confident it is backward compadible before merging.

lucienlu-aws:
Backwards compatibility tests are failing due to previous PR https://github.com/awslabs/amazon-kinesis-client/pull/1340. Will double check with owners

lucienlu-aws:
Double checked with owner of https://github.com/awslabs/amazon-kinesis-client/pull/1340, the failures is due to internally generated code and we should be okay.

PR #1459, opened 2025-03-31
Dependabot auto merge addition in github workflows

skyero-aws:
*Issue #, if available:*

*Description of changes:*
Add auto-merge functionality to github workflows for dependabot. Auto-merge requires build job to pass, dependabot to be the pull requester, and the requested dependency must be both not a major version and have a cvss level above 0.


amazon-kinesis-client/pom.xml
```diff
@@ -369,7 +369,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-failsafe-plugin</artifactId>
-        <version>3.2.5</version>
```
vincentvilo-aws, line 372: Why are we downgrading the version here?

.github/workflows/maven.yml
```diff
@@ -43,4 +45,21 @@ jobs:
...
+    if: github.event.pull_request.user.login == 'dependabot[bot]'
+    steps:
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v2
+        with:
+          alert-lookup: true
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Enable auto-merge for Dependabot PRs
+        if: steps.metadata.outputs.update-type != 'version-update:semver-major' && steps.metadata.outputs.cvss > 0
```
vincentvilo-aws, line 61: `steps.metadata.outputs.update-type != 'version-update:semver-major'` 

From what I understand, this line means that any patch or minor version upgrades will be automerged. Can we modify this so that we only automerge the patch dependencies instead of both the patch and minor versions?

.github/dependabot.yml
```diff
@@ -1,32 +1,31 @@
...
-# branch - master
-- package-ecosystem: "maven"
-  directory: "/"
-  labels:
-    - "dependencies"
-    - "v3.x"
-  target-branch: "master"
-  schedule:
-    interval: "weekly"
+  # branch - master
```
vincentvilo-aws, line 3: nit: For future PRs, we can ignore the whitespace changes

.github/workflows/maven.yml
pelaezryan: Is there a reason why we don't want to do minor as well? Minor and patch should be OK.

vincentvilo-aws: For some dependencies, minor version upgrades might include dependencies that are incompatible with KCL. For example, we currently use [logback v 1.3.x](https://logback.qos.ch/news.html) since it's compatible with JDK 8, but v1.4.x and above requires JDK 11.

amazon-kinesis-client/pom.xml
skyero-aws: That was the version of the maven failsafe plugin that was live last thursday when I synced the pom files, I will edit the merge so it doesn't revert needlessly.

.github/workflows/maven.yml
skyero-aws: We could *hypothetically* do that, but Zeynep wants both patch and minor versions to be merged. If we change our minds later we can edit the line to just say

skyero-aws: Changed the logic to only trigger auto-merge on patch versions.

PR #1464, opened 2025-04-09
Make LAM run at configurable interval

ehasah-aws:
*Issue #, if available:*

*Description of changes:*

- Make Lease Assignment manager (LAM) run at configurable interval
- Tie LeaseDiscoverer with LAM frequency


amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseAssignmentManager.java
```diff
@@ -129,10 +130,7 @@ public synchronized void start() {
             // so reset the flag to refresh the state before processing during a restart of LAM.
             tookOverLeadershipInThisRun = false;
             managerFuture = executorService.scheduleWithFixedDelay(
-                    this::performAssignment,
-                    0L,
-                    leaseDurationMillis * DEFAULT_LEASE_ASSIGNMENT_MANAGER_FREQ_MULTIPLIER,
-                    TimeUnit.MILLISECONDS);
+                    this::performAssignment, 0L, (int) (leaseAssignmentIntervalMillis), TimeUnit.MILLISECONDS);
```
pelaezryan, line 133: nit: lets try to avoid formatting changes.

pelaezryan (dismissed):

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseAssignmentManager.java
```diff
@@ -129,10 +130,7 @@ public synchronized void start() {
             // so reset the flag to refresh the state before processing during a restart of LAM.
             tookOverLeadershipInThisRun = false;
             managerFuture = executorService.scheduleWithFixedDelay(
-                    this::performAssignment,
-                    0L,
-                    leaseDurationMillis * DEFAULT_LEASE_ASSIGNMENT_MANAGER_FREQ_MULTIPLIER,
```
lucienlu-aws, line 134: We can delete this multiplier variable right? Its unused now

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
```diff
@@ -114,6 +114,16 @@ public class LeaseManagementConfig {
...
+    /**
+     * Lease assignment interval in milliseconds - e.g. wait for this long between Lease assignment run.
+     *
+     * <p>Default value: 2 * {@link LeaseManagementConfig#failoverTimeMillis}</p>
+     */
+    private Long leaseAssignmentIntervalMillis;
+
+    public long leaseAssignmentIntervalMillis() {
+        return leaseAssignmentIntervalMillis != null ? leaseAssignmentIntervalMillis : 2 * failoverTimeMillis;
+    }
```
lucienlu-aws, line 126: I don't think we need this function. We can just set

```
private Long leaseAssignmentIntervalMillis = 2 * failoverTimeMillis;
```

lucienlu-aws:
Also, please make sure to squash the commits and update your commit message so its shorter when merging

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
pelaezryan: If i understand correctly, leaseAssignment is intended to be configurable with the default being 2x the failover. Hence the justification for this function.

ehasah-aws: @lucienlu-aws i used `private Long leaseAssignmentIntervalMillis = 2 * failoverTimeMillis;` before but leaseAssignmentIntervalMillis always gets default failoverTimeMillis value (which is 2 * 10000) even when i set failoverTimeMillis to a different value. unless i set leaseAssignmentIntervalMillis explicitly. That's why i added/override  this method.

lucienlu-aws: You're right, since if the failoverTime is updated it won't recompute everytime

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseAssignmentManager.java
ehasah-aws: @pelaezryan  i tried to remove formatting and pushed the code but it failed the build saying i need to fix format violations

pelaezryan (approved):

lucienlu-aws (approved):

PR #1482, opened 2025-06-05
Add options to set leader lifetime and heartbeat

andrew-nowak:
Hi!

We're using KCL in a possibly odd way - running with a single worker. After recently attempting an upgrade to KCL v3, we found that when the worker was replaced we'd experience a much longer delay before the new worker took over the leases and started processing. After some digging we've figured out that this is due to the change of the leader decider mechanism, which in v3 waits up to 2 minutes before invalidating the old leader.

https://github.com/awslabs/amazon-dynamodb-lock-client offers options for configuring the lease duration and heartbeat period, but KCL currently doesn't expose those options to consumers to allow them to choose values that better suit their usecases.

*Description of changes:*

Exposes two new properties on the `LeaseManagementConfig`, which allow changing the lease duration and heartbeat period from their defaults, which remain at 120 and 30 seconds respectively, and pass those values down into the dynamodb-lock-client.

If this doesn't sound right, I'd very happily take suggestions on other approaches.

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
@@ -282,6 +282,9 @@ public class LeaseManagementConfig {
     private long listShardsCacheAllowedAgeInSeconds = 30;
     private int cacheMissWarningModulus = 250;
 
+    private long dynamoLockBasedLeaderLeaseDurationInSeconds = 120;
+    private long dynamoLockBasedLeaderHeartbeatPeriodInSeconds = 30;
+
     private MetricsFactory metricsFactory = new NullMetricsFactory();
 
     @Deprecated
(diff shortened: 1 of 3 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java
```diff
@@ -282,6 +282,9 @@ public long leaseAssignmentIntervalMillis() {
     private long listShardsCacheAllowedAgeInSeconds = 30;
     private int cacheMissWarningModulus = 250;
 
+    private long dynamoLockBasedLeaderLeaseDurationInSeconds = 120;
+    private long dynamoLockBasedLeaderHeartbeatPeriodInSeconds = 30;
```
lucienlu-aws: Could you update this to use milliseconds as the unit and rename this to include the full name `dynamoDb`? 
- `dynamoDbLockBasedLeaderLeaseDurationInMillis` 
- `dynamoDbLockBasedLeaderHeartbeatPeriodInMillis`

Also, add javadoc for these variables, particularly that the heartbeat period is recommended to be set at least 3x smaller than the lease duration. Here is the DynamoDB lock based client javadoc https://github.com/awslabs/amazon-dynamodb-lock-client/blob/0b71da96c942e1d76d0bdcfbd5c59aabb63b0265/src/main/java/com/amazonaws/services/dynamodbv2/AmazonDynamoDBLockClientOptions.java#L138-L162

andrew-nowak: Thanks for looking, I've pushed a commit that does both of these. Happy to take any suggestions you might have for improving the javadoc wordings

lucienlu-aws: Javadocs look good, thanks

lucienlu-aws:
Build is failing due to a `var` type in unrelated unit test which is only in JDK 10+. I'll fix in another PR

PR #1491, opened 2025-06-11
Fixed plugin related to skipping tests

pelaezryan:
### Background
Currently, the actual skip tests functionality was broken as ``<skipITs>${skipITs}</skipITs>`` is part of ``maven-failsafe-plugin`` and not ``maven-surefire-plugin``

Because of this, running ``mvn install`` leads to 
```
[WARNING] Parameter 'skipITs' is unknown for plugin 'maven-surefire-plugin:3.2.5:test (default-test)'
```

### DIFF
--- a/amazon-kinesis-client/pom.xml
+++ b/amazon-kinesis-client/pom.xml
@@ -54,7 +54,8 @@
     <sqlite4java.libpath>${project.build.directory}/test-lib</sqlite4java.libpath>
     <slf4j.version>2.0.13</slf4j.version>
     <gsr.version>1.1.19</gsr.version>
-    <skipTests>true</skipTests>
+    <skipUT>false</skipUT>
+    <skipIT>true</skipIT>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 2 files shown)


amazon-kinesis-client/pom.xml
```diff
@@ -349,8 +349,7 @@
         <artifactId>maven-surefire-plugin</artifactId>
         <version>3.2.5</version>
         <configuration>
-          <skipTests>${skip.ut}</skipTests>
```
ethkatnic, line 352: We reference this property in the README [here](https://github.com/awslabs/amazon-kinesis-client/blob/master/README.md#building-from-source). 
> To disable running unit tests in the build, add the property -Dskip.ut=true.



We either need to update the documentation like "To disable running unit tests in the build, add the property `-DskipTests=true`." or keep this property as {skip.ut}

ethkatnic: These vars should remain separate, as one disables unit tests while the other disables integration tests. These should remain independent.

PR #1494, opened 2025-06-12
Sonatype/Maven Central Publishing Migration

skyero-aws:
*Issue #, if available:*

*Description of changes:*
Following migration steps listed here: https://central.sonatype.org/publish/publish-portal-maven/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -157,7 +157,7 @@
             <extensions>true</extensions>
             <configuration>
               <publishingServerId>central</publishingServerId>
-              <autoReleaseAfterClose>false</autoReleaseAfterClose>
+              <autoPublish>false</autoPublish>
             </configuration>
           </plugin>
         </plugins>
--- a/pom.xml
+++ b/pom.xml
@@ -63,12 +63,12 @@
 
   <distributionManagement>
     <snapshotRepository>
-      <id>ossrh</id>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -151,13 +151,12 @@
...
-            <version>1.6.13</version>
+            <groupId>org.sonatype.central</groupId>
+            <artifactId>central-publishing-maven-plugin</artifactId>
+            <version>0.7.0</version>
             <extensions>true</extensions>
             <configuration>
-              <serverId>sonatype-nexus-staging</serverId>
-              <nexusUrl>https://aws.oss.sonatype.org</nexusUrl>
+              <publishingServerId>central</publishingServerId>
               <autoReleaseAfterClose>false</autoReleaseAfterClose>
```
lucienlu-aws: I don't think `autoReleaseAfterClose` is a flag anymore in the new plugin. Looks likes there is `<autoPublish>true</autoPublish>`

lucienlu-aws: We should probably not autopublish though

skyero-aws: I'll change the flag to `<autoPublish>false</autoPublish>` if thats the case, or should I remove the flag altogether

lucienlu-aws: `false` is fine to make it explicit for future readers

PR #1495, opened 2025-06-17
Add millisBehindLatestThresholdForReducedTps property to polling config

ethkatnic:
*Description of changes:*
Adds configurable property `millisBehindLatestThresholdForReducedTps` to `PollingConfig`.

Sets the threshold for millisBehindLatest that will trigger reduced throughput when close to tip. This prevents excessive polling calls to the Kinesis service when caught up to latest data. When most recent record has millisBehindLatest less than this threshold, additional sleep time will be added. Sleep time will be the difference between time of last successful record retrieval and this threshold. Default value of 0 effectively disables the reduced TPS functionality unless intentionally set.


amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
```diff
@@ -239,6 +246,60 @@ public PrefetchRecordsPublisher(
             final ThrottlingReporter throttlingReporter,
             final long awaitTerminationTimeoutMillis,
             final SleepTimeController sleepTimeController) {
+        this(
```
lucienlu-aws: Why are we adding more constructors to this class? Do they just call each other? Can we reduce the number of constructors?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/RecordsFetcherFactory.java
```diff
@@ -21,6 +21,12 @@
  * This factory is used to create the records fetcher to retrieve data from Kinesis for a given shard.
  */
 public interface RecordsFetcherFactory {
+    /**
+     * Default value for millisBehindLatestThresholdForReducedTps.
+     * A value of 0 effectively disables the reduced TPS functionality.
+     */
+    long DEFAULT_MILLIS_BEHIND_LATEST_THRESHOLD_FOR_REDUCED_TPS = 0L;
```
lucienlu-aws: I think having the default config variable in PollingConfig makes more sense

ethkatnic: Agreed, moved to polling config

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/polling/PrefetchRecordsPublisher.java
ethkatnic: I had originally added constructors and chained them to work with or without the new field. However, since this is an internal API, we can just add the arg to the constructor and change it wherever called. Keeps the constructors minimal.

lucienlu-aws (approved):

PR #1519, opened 2025-07-15
Fix flaky LeaseAssignmentManager retry behavior test

lucienlu-aws:
*Issue #, if available:*

*Description of changes:*
Previously, the testRetryBehavior test would mock an exception on both the `leaseRefresher` and `workerMetricStatsDao`. In the LAM, however, these calls are asynchronous ([ref](https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseAssignmentManager.java#L522-L525)), causing some runs to run into the mocked exception on the `workerMetricStatsDao` twice and exit the `loadInMemoryStorageView` method before `leaseRefresher` could throw the mocked exception since we only retry once. Technically, the race occurs if either one of the async calls throw twice before the other async call can throw once, since that will throw an exception in the method and exit the method.

On [this failed run](https://github.com/awslabs/amazon-kinesis-client/actions/runs/16176026775/job/45661264985#step:4:9940) on line 9940, the first exception caught is from `workerMetricStatsDao` and the method exits to [here](https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseAssignmentManager.java#L263) without any exception from `leaseRefresher`

[This successful run](https://github.com/awslabs/amazon-kinesis-client/actions/runs/16177376262/job/45665652148#step:4:9940) on line 9940 catches the `leaseRefresher` exception first and then the `workerMetricStatsDao` exception.

This commit splits up the test retry behavior so this race in the unit tests does not occur. One test mocks an exception on the `leaseRefresher` and verifies that it is called twice. The second test mocks the exception on the `workerMetricStatsDao` and verifies that it is called twice.

### DIFF
--- a/amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/assignment/LeaseAssignmentManagerTest.java
+++ b/amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/assignment/LeaseAssignmentManagerTest.java
@@ -729,7 +729,6 @@ class LeaseAssignmentManagerTest {
     @Test
     void performAssignment_testRetryBehaviorForLeaseRefresher()
             throws ProvisionedThroughputException, InvalidStateException, DependencyException {
-        final WorkerMetricStatsDAO mockedWorkerMetricsDAO = Mockito.mock(WorkerMetricStatsDAO.class);
         final LeaseRefresher mockedLeaseRefresher = Mockito.mock(LeaseRefresher.class);
 
         when(mockedLeaseRefresher.listLeasesParallely(any(), anyInt())).thenThrow(new RuntimeException());
--- a/amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/assignment/LeaseAssignmentManagerTest.java
+++ b/amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/assignment/LeaseAssignmentManagerTest.java
@@ -727,36 +727,43 @@ class LeaseAssignmentManagerTest {
     }
 
     @Test
-    void performAssignment_testRetryBehavior()
+    void performAssignment_testRetryBehaviorForLeaseRefresher()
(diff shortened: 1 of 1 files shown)


amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/assignment/LeaseAssignmentManagerTest.java
```diff
@@ -727,36 +727,42 @@ void performAssignment_staleWorkerMetricsEntries_assertCleaning() {
...
 
     @Test
-    void performAssignment_testRetryBehavior()
+    void performAssignment_testRetryBehaviorForLeaseRefresher()
             throws ProvisionedThroughputException, InvalidStateException, DependencyException {
-
-        final WorkerMetricStatsDAO mockedWorkerMetricsDAO = Mockito.mock(WorkerMetricStatsDAO.class);
         final LeaseRefresher mockedLeaseRefresher = Mockito.mock(LeaseRefresher.class);
 
         when(mockedLeaseRefresher.listLeasesParallely(any(), anyInt())).thenThrow(new RuntimeException());
```
chenylee-aws, line 734: If we want to throw once i think we can chain this with a .thenReturn(null) so we don't throw twice for either of the mock

chenylee-aws (approved):

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/assignment/LeaseAssignmentManagerTest.java
lucienlu-aws: Yea I think that would also solve the flakiness. I think its better to still individually test the retries on the components to make it test one thing at a time

PR #1523, opened 2025-07-18
use processRecordsInput.toBuilder() and not drop childShards

yitianwang18:
*Issue #, if available:*

*Description of changes:* 

Avoid copying over fields and derive from original input.

Per [javadoc](https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/events/ProcessRecordsInput.java) of ProcessRecordsInput, also passing childShards when shard has ended. 

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ProcessTask.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ProcessTask.java
@@ -245,13 +245,9 @@ public class ProcessTask implements ConsumerTask {
     private void callProcessRecords(ProcessRecordsInput input, List<KinesisClientRecord> records) {
         log.debug("Calling application processRecords() with {} records from {}", records.size(), shardInfoId);
 
-        final ProcessRecordsInput processRecordsInput = ProcessRecordsInput.builder()
+        final ProcessRecordsInput processRecordsInput = input.toBuilder()
                 .records(records)
-                .cacheExitTime(input.cacheExitTime())
-                .cacheEntryTime(input.cacheEntryTime())
-                .isAtShardEnd(input.isAtShardEnd())
                 .checkpointer(recordProcessorCheckpointer)
-                .millisBehindLatest(input.millisBehindLatest())
                 .build();
 
         final MetricsScope scope = MetricsUtil.createMetricsWithOperation(metricsFactory, PROCESS_TASK_OPERATION);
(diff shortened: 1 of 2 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ProcessTask.java
```diff
@@ -245,13 +245,9 @@ private void backoff() {
     private void callProcessRecords(ProcessRecordsInput input, List<KinesisClientRecord> records) {
         log.debug("Calling application processRecords() with {} records from {}", records.size(), shardInfoId);
 
-        final ProcessRecordsInput processRecordsInput = ProcessRecordsInput.builder()
+        final ProcessRecordsInput processRecordsInput = input.toBuilder()
                 .records(records)
```
akidambisrinivasan, line 249: Why does this need to be set again ?

amazon-kinesis-client/src/test/java/software/amazon/kinesis/lifecycle/ProcessTaskTest.java
```diff
@@ -486,7 +489,8 @@ public void testAllInShardKplData() throws Exception {
                                 .build())
                         .build());
 
-        when(processRecordsInput.records()).thenReturn(rawRecords);
+        processRecordsInput =
```
akidambisrinivasan, line 555: inconsistent convention either do 

```
processRecordsInput
    = processRecordsInput.toBuilder().....
```

or
```
processRecordsInput = processRecordsInput
    .toBuilder()...
```

```
```

yitianwang18: Noted. I will see if I can do that and satisfy spotless check at the same time. 

Given that other code are following this convention of
```
        return KinesisClientRecord.builder()
                .partitionKey(partitionKey)
                .sequenceNumber(sequenceNumber)
                .approximateArrivalTimestamp(arrival)
                .data(ByteBuffer.wrap(TEST_DATA))
                .build();
```
I will follow this convention:
```
processRecordsInput = processRecordsInput.toBuilder()
        ...
        .build();

```

yitianwang18: I tried changing the code to use this convention, but that won't satisfy `mvn spotless:check`. Spotless really prefer this style of line break after the assignment. Either I need to find a way to override the rule or the code has to be this way.

Also, code like
```
        List<KinesisClientRecord> actualRecords =
                outcome.getProcessRecordsCall().records();
```
Already exists in the codebase.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ProcessTask.java
yitianwang18: The records are set because it may be modified by deaggregation before the function call.

akidambisrinivasan (approved):

chenylee-aws (approved):

PR #1670, opened 2025-12-30
Add lease count-based assignment strategy support

chenylee-aws:
*Description of changes:*

Support lease based lease count assignment strategy  

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseCountBasedLeaseAssignmentDecider.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseCountBasedLeaseAssignmentDecider.java
@@ -1,0 +1,245 @@
+/*
+ * Copyright 2024 Amazon.com, Inc. or its affiliates.
+ * Licensed under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package software.amazon.kinesis.coordinator.assignment;
+
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import software.amazon.kinesis.annotations.KinesisClientInternalApi;
+import software.amazon.kinesis.leases.Lease;
+
+import static java.util.Objects.nonNull;
+
+/**
+ * A lease assignment decider that distributes leases based on lease count balancing.
+ * This implementation ensures fair distribution of leases across workers by:
+ * <ul>
+ *   <li>Assigning expired/unassigned leases to workers with the fewest current leases</li>
+ *   <li>Balancing lease distribution to minimize variance between workers</li>
+ *   <li>Respecting the maximum lease limit per worker</li>
+ * </ul>
+ */
+@Slf4j
+@KinesisClientInternalApi
+@RequiredArgsConstructor
+public final class LeaseCountBasedLeaseAssignmentDecider implements LeaseAssignmentDecider {
+
+    private final LeaseAssignmentManager.InMemoryStorageView inMemoryStorageView;
+    private final int maxLeasesForWorker;
+
+    /**
+     * Assigns expired or unassigned leases to workers with the fewest current leases.
+     * Leases are sorted by lastCounterIncrementNanos to ensure consistent assignment order.
+     *
+     * @param expiredOrUnAssignedLeases list of leases that need to be assigned to workers
+     */
+    @Override
+    public void assignExpiredOrUnassignedLeases(final List<Lease> expiredOrUnAssignedLeases) {
+        if (expiredOrUnAssignedLeases.isEmpty()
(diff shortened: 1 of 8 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseCountBasedLeaseAssignmentDecider.java
```diff
@@ -0,0 +1,245 @@
...
+ * <ul>
+ *   <li>Assigning expired/unassigned leases to workers with the fewest current leases</li>
+ *   <li>Balancing lease distribution to minimize variance between workers</li>
+ *   <li>Respecting the maximum lease limit per worker</li>
+ * </ul>
+ */
+@Slf4j
+@KinesisClientInternalApi
+@RequiredArgsConstructor
+public final class LeaseCountBasedLeaseAssignmentDecider implements LeaseAssignmentDecider {
```
lucienlu-aws, line 45: What should we do with the [`varianceBalancingFrequency`](https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/LeaseManagementConfig.java#L605) now? I think all the other configs in `WorkerUtilizationAwareAssignmentConfig` are only relevant in the variance based lease assignment decider but `varianceBalancingFrequency` is something for LAM and can still be used with this lease count based assignment. Should we deprecate that config and move it and create a a new config class for LAM specific based configs? We might need a specific config class for each LeaseAssignmentDecider implementation

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseCountBasedLeaseAssignmentDecider.java
```diff
@@ -0,0 +1,245 @@
...
+
+        final List<Map.Entry<String, Integer>> sortedWorkers = estimatedLeaseCounts.entrySet().stream()
+                .sorted(Map.Entry.comparingByValue())
+                .collect(Collectors.toList());
+
+        // Two pointers: left (underloaded) and right (overloaded)
+        int left = 0;
+        int right = sortedWorkers.size() - 1;
+
+        while (left < right) {
```
lucienlu-aws, line 123: This moves all possible leases from overloaded workers to underloaded workers. I recall KCL 2.x had a maxLeasesToSteal config which limits this. Do you think we need this?

I guess in worst case if one worker increases to two workers, then half the leases will get moved in one LAM iteration

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/assignment/LeaseAssignmentManagerTest.java
```diff
@@ -1352,4 +1392,38 @@ private void writeToWorkerMetricsTables(final WorkerMetricStats workerMetrics) {
...
+        workerMetricsDAO.updateMetrics(createDummyTakeWorkerMetrics(TEST_TAKE_WORKER_ID));
+        leaseRefresher.createLeaseIfNotExists(createDummyUnAssignedLease("lease1"));
+
+        leaseAssignmentManagerRunnable.run();
+
+        assertEquals(TEST_TAKE_WORKER_ID, leaseRefresher.listLeases().get(0).leaseOwner());
+    }
+
+    @Test
+    void performAssignment_withLeaseCountStrategy_usesLeaseCountDecider() throws Exception {
```
lucienlu-aws, line 1446: These tests don't actually check which leaseDecider is being used, just that the leases are assigned. Is there a more robust way we can write this test?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseCountBasedLeaseAssignmentDecider.java
xyulinaws: Yes, we should respect maxLeasesToSteal config, which is important to make sure lease shuffling don't happen too aggressively during fleet deployment. In particular, when 10% of the fleet is being deployed to, we want ideally the rest of the 90% of the fleet to each take up a few leases to balance the load, but when maxLeasesToSteal is too high, we ended up with a few hosts stealing all the leases at once and taking up too much load.

chenylee-aws: The concern about aggressive lease shuffling during deployments is valid for the old KCL 2.x `DynamoDBLeaseTaker` where each worker independently steals leases. However, in LAM the leader is a single centralized decision-maker — there's no race between workers. The two-pointer algorithm already distributes steals across multiple underloaded workers one at a time, and expired lease assignment (the main path during deployments) already spreads load evenly via the target calculation (`ceil(totalLeases/totalWorkers)`).

Adding a global steal cap to `balanceWorkerVariance` could be useful as a dampening mechanism to limit churn per cycle, but it's a different concern than the fleet deployment scenario described. We can introduce this config if needed

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseCountBasedLeaseAssignmentDecider.java
chenylee-aws: Naming aside, think of `WorkerUtilizationAwareAssignmentConfig` as the LAM config. It controls LeaseAssignmentManager behavior regardless of which strategy is selected. Some configs like `dampeningPercentage` are variance specific and won't apply to lease count, but the lease count strategy has very few configs of its own, so a separate class isn't warranted yet. We can always introduce new config if needed

amazon-kinesis-client/src/test/java/software/amazon/kinesis/coordinator/assignment/LeaseAssignmentManagerTest.java
chenylee-aws: Added more tests

vincentvilo-aws (approved):

PR #1728, opened 2026-04-06
feat(metrics): Add CloudWatch OTEL endpoint support

wisemanIV:
Add alternative metrics publishing path that sends KCL metrics to the CloudWatch OTEL endpoint in OTLP format. Users can choose between the existing CloudWatch PutMetricData API and the new OTEL endpoint via MetricsConfig.metricsBackend().

New classes:
- MetricsBackend enum (CLOUDWATCH, CLOUDWATCH_OTEL)
- OtelMetricsFactory, OtelMetricsScope, OtelPublisherRunnable
- OtelMetricsPublisher with OTLP Histogram conversion
- OtelMetricKey for metric accumulation

Key behaviors:
- OTEL path defaults to 30s publish interval (vs 10s for CloudWatch)
- Maps KCL dimensions to OTEL semantic conventions
- Auto-detects resource attributes from KCL configuration
- Custom resource attributes override auto-detected ones
- Backward compatible: default remains CloudWatch endpoint

Tests:
- OtelMetricsPublisherTest (batching, conversion, dimension mapping)
- OtelPublisherRunnableTest (flush, timeout, drain, shutdown)
- OtelMetricsConfigTest (backend selection, buffer time defaults)

*Issue #, if available:*

*Description of changes:*


wisemanIV:
@chenylee-aws  @lucienlu-aws @ehasah-aws would appreciate your review.

wisemanIV:
please can you review / approve? @chenylee-aws @lucienlu-aws @ehasah-aws

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/MetricsConfig.java
```diff
@@ -106,19 +109,90 @@ public class MetricsConfig {
...
         }
         return metricsFactory;
     }
+
+    /**
+     * Builds auto-detected OTEL resource attributes from KCL configuration.
+     */
+    private Map<String, String> buildAutoDetectedResourceAttributes() {
+        Map<String, String> attributes = new HashMap<>();
+        attributes.put("service.name", namespace());
```
john-karp: If the KCL library is to be treated as a service in itself, then we should define the full service triad service.namespace / service.name / service.instance.id and follow the requirements for the same at https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/

However at least in my application, I would prefer that KCL not set its own service attributes, and just take on the service attributes of the application.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsPublisher.java
```diff
@@ -0,0 +1,263 @@
...
+    /**
+     * Mapping of KCL dimension names to OTEL semantic convention attribute names.
+     */
+    static final Map<String, String> DIMENSION_TO_ATTRIBUTE_MAP;
+
+    static {
+        Map<String, String> dimMap = new HashMap<>();
+        dimMap.put(MetricsUtil.OPERATION_DIMENSION_NAME, "aws.kinesis.operation");
+        dimMap.put(MetricsUtil.SHARD_ID_DIMENSION_NAME, "aws.kinesis.shard.id");
+        dimMap.put(MetricsUtil.STREAM_IDENTIFIER, "messaging.destination.name");
```
john-karp: https://opentelemetry.io/docs/specs/semconv/registry/attributes/aws/#amazon-kinesis-attributes would suggest `aws.kinesis.stream_name`?

amazon-kinesis-client/pom.xml
```diff
@@ -275,6 +310,12 @@
       <version>1.3.14</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>net.jqwik</groupId>
+      <artifactId>jqwik</artifactId>
```
john-karp: Where is this used?

amazon-kinesis-client/pom.xml
```diff
@@ -203,6 +217,27 @@
...
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk-metrics</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-exporter-otlp</artifactId>
+    </dependency>
+
```
john-karp, line 223: Consider making these optional?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/MetricsConfig.java
```diff
@@ -106,19 +109,90 @@ public class MetricsConfig {
...
+        attributes.put("cloud.provider", "aws");
```
john-karp: In practice I think this could collide with projects that use this resource attribute to describe where the client is running, which could be on-prem or on a different cloud provider.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsPublisher.java
```diff
@@ -0,0 +1,263 @@
...
+public class OtelMetricsPublisher {
+
+    private static final int BATCH_SIZE = 20;
+    private static final long EXPORT_TIMEOUT_MILLIS = 5000;
+
+    private static final Map<StandardUnit, String> UNIT_MAP;
+
+    static {
+        Map<StandardUnit, String> map = new HashMap<>();
+        map.put(StandardUnit.SECONDS, "s");
```
john-karp: First, seems simpler to do:
```suggestion
        UNIT_MAP = Collections.unmodifiableMap(
            Stream.of(StandardUnit.values())
                .collect(Collectors.toMap(k -> k, StandardUnit::toString)));
```

Second, some of these unit abbrevations seem idiosyncratic. I don't think I've seen 'By' for bytes before.

john-karp (commented):
As a user of both Otel and Kinesis, this looks pretty interesting. Leaving some comments.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/MetricsConfig.java
wisemanIV: Agreed. I will push an update.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsPublisher.java
wisemanIV: Agreed. I will push an update.

amazon-kinesis-client/pom.xml
wisemanIV: Agreed. I will push an update.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsPublisher.java
wisemanIV: StandardUnit.toString() returns values like "Bytes", "Milliseconds", "Count/Second" — those are CloudWatch unit names, not UCUM strings. OTel requires UCUM notation where bytes is By, kilobytes is kBy, etc. So StandardUnit::toString would produce the wrong values.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/MetricsConfig.java
wisemanIV: Agreed. I will push an update.

amazon-kinesis-client/pom.xml
wisemanIV: removed in latest commit.

wisemanIV:
> As a user of both Otel and Kinesis, this looks pretty interesting. Leaving some comments.

Thanks for the feedback @john-karp. I've implemented most of the suggestions with the exception of the unit types. Please see explanation and let me know if you disagree.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsPublisher.java
john-karp: I see, you are right. I was looking at the wrong part of the Otel docs.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsPublisher.java
```diff
@@ -0,0 +1,263 @@
...
+ * Publisher that converts KCL MetricDatum objects into OTLP format and sends them
+ * to the CloudWatch OTEL endpoint via the OTLP HTTP exporter.
+ */
+@Slf4j
+public class OtelMetricsPublisher {
+
+    private static final int BATCH_SIZE = 20;
+    private static final long EXPORT_TIMEOUT_MILLIS = 5000;
+
+    private static final Map<StandardUnit, String> UNIT_MAP;
```
john-karp: Since this map concerns the relationship between CloudWatch `StandardUnit` and OpenTelemetry, and has no direct relationship to Kinesis, ideally this mapping would be some place non-Kinesis-specific.

* Is there some OTEL-specific utility package in CloudWatch?
* Would make sense to have it in CloudWatch SDK itself?
* https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main ?
 
In the short term, perhaps having this map here is the best way forward.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricKey.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * A representation of a key of a MetricDatum for the OTEL metrics path. This class is useful when
+ * wanting to compare whether 2 keys have the same MetricDatum. This feature will be used in
+ * MetricAccumulatingQueue where we aggregate metrics across multiple MetricScopes.
+ */
+public class OtelMetricKey {
+
+    private List<Dimension> dimensions;
+    private String metricName;
```
john-karp: `private final` for both? Ideally classes used for keys are immutable.

amazon-kinesis-client/.jqwik-database
john-karp, line 1: Stray file?

john-karp (commented):
I'm not sure this design is interacting with otel at the right level of abstraction. It is hard-coding the OtlpHttpMetricExporter, and as far as I can tell it is not pulling in standard [environment configuration](https://opentelemetry.io/docs/languages/sdk-configuration/general/) of attributes and exporters.

Looking at:

* https://opentelemetry.io/docs/concepts/instrumentation/libraries/#opentelemetry-api 
* https://opentelemetry.io/docs/languages/java/instrumentation/#native-instrumentation
* https://opentelemetry.io/docs/languages/java/api/#globalopentelemetry
* https://opentelemetry.io/docs/concepts/instrumentation-scope/

It would seem the standard way for a library to natively instrument itself with OTEL is:
 
 * "Libraries should only use the OpenTelemetry API" (that is, not the SDK packages)
 * "Native instrumentation should use `GlobalOpenTelemetry.getOrNoop()` as the default {@link OpenTelemetry} instance, and expose APIs for setting a custom instance."
 * "The [Instrumentation scope](https://opentelemetry.io/docs/specs/otel/common/instrumentation-scope/) represents a logical unit within the application code with which the emitted telemetry can be associated. [...] In the case of a library or framework, a common approach is to use an identifier as scope that is unique to the library or framework, such as a fully qualified name and version of the library or framework."
 
So then I think the pattern would be to use `meterBuilder`/`getMeter` instantiated by the `OpenTelemetry` instance that was returned from `GlobalOpenTelemetry.getOrNoop()`, with the appropriate instrumentation scope name set.

That is difficult though if you are trying to process pre-aggregated MetricDatum objects. Based on the pattern that's used in my application, you could change OtelMetricsScope to override `addData`, which handles individual data points. And then each OtelMetricsScope could have an internal cache for looking up OTEL meter instances by name, instantiated with the proper meter type.

Then you'd get auto-configuration of exporter and attributes, and you could skip all the buffering logic.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricKey.java
wisemanIV: agreed. I will update.

amazon-kinesis-client/.jqwik-database
wisemanIV: Yes looks like it. I will remove.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsPublisher.java
wisemanIV: I kept the map in OtelMetricsScope. There's no existing OTel-specific utility package in the CloudWatch SDK or opentelemetry-java-contrib that provides this mapping today.

wisemanIV:
> I'm not sure this design is interacting with otel at the right level of abstraction. It is hard-coding the OtlpHttpMetricExporter, and as far as I can tell it is not pulling in standard [environment configuration](https://opentelemetry.io/docs/languages/sdk-configuration/general/) of attributes and exporters.
> 
> Looking at:
> 
> * https://opentelemetry.io/docs/concepts/instrumentation/libraries/#opentelemetry-api
> * https://opentelemetry.io/docs/languages/java/instrumentation/#native-instrumentation
> * https://opentelemetry.io/docs/languages/java/api/#globalopentelemetry
> * https://opentelemetry.io/docs/concepts/instrumentation-scope/
> 
> It would seem the standard way for a library to natively instrument itself with OTEL is:
> 
> * "Libraries should only use the OpenTelemetry API" (that is, not the SDK packages)
> * "Native instrumentation should use `GlobalOpenTelemetry.getOrNoop()` as the default {@link OpenTelemetry} instance, and expose APIs for setting a custom instance."
> * "The [Instrumentation scope](https://opentelemetry.io/docs/specs/otel/common/instrumentation-scope/) represents a logical unit within the application code with which the emitted telemetry can be associated. [...] In the case of a library or framework, a common approach is to use an identifier as scope that is unique to the library or framework, such as a fully qualified name and version of the library or framework."
> 
> So then I think the pattern would be to use `meterBuilder`/`getMeter` instantiated by the `OpenTelemetry` instance that was returned from `GlobalOpenTelemetry.getOrNoop()`, with the appropriate instrumentation scope name set.
> 
> That is difficult though if you are trying to process pre-aggregated MetricDatum objects. Based on the pattern that's used in my application, you could change OtelMetricsScope to override `addData`, which handles individual data points. And then each OtelMetricsScope could have an internal cache for looking up OTEL meter instances by name, instantiated with the proper meter type.
> 
> Then you'd get auto-configuration of exporter and attributes, and you could skip all the buffering logic.

I agree with the feedback. I was trying to add an exported to the existing metrics pipeline but needed to go higher in the architecture.

wisemanIV:
@john-karp thanks for the feedback. I've made some significant changes to be more OTel native.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
```diff
@@ -0,0 +1,180 @@
...
+        addData(name, value, unit, MetricsLevel.DETAILED);
+    }
+
+    @Override
+    public void addData(String name, double value, StandardUnit unit, MetricsLevel level) {
+        checkNotEnded();
+        if (level.getValue() < metricsLevel.getValue()) {
+            return;
+        }
+        observations.add(new MetricObservation(name, value, unit));
```
john-karp: Now, I'm not entirely sure why the KCL metrics design queues up metrics internally before flushing in `end()` (presumably performance?), but I'm not sure it makes sense to carry this forward with Otel API, since I think API calls such as `counter.add` and `histogram.record` are [intended to be fast and low-allocation](https://opentelemetry.io/blog/2024/java-metric-systems-compared/). Doing it this way with the List of MetricObservations results in extra allocations that wouldn't otherwise be needed.

Also, it looks like in some cases scopes can be open for some time before being closed. If we don't wait for a flush, they'll be recorded at a time that will match up better with other telemetry such as logs.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
```diff
@@ -0,0 +1,180 @@
...
+        if (obs.unit == StandardUnit.COUNT) {
+            LongCounter counter = meter.counterBuilder(obs.name)
+                    .setUnit(convertUnit(obs.unit))
+                    .build();
+            counter.add((long) obs.value, attrs);
+        } else {
+            DoubleHistogram histogram = meter.histogramBuilder(obs.name)
+                    .setUnit(convertUnit(obs.unit))
+                    .build();
+            histogram.record(obs.value, attrs);
```
john-karp: I think this mapping works pretty well for the majority of the metrics in KCL. There are some StandardUnit.COUNT metrics that might be better represented as Gauge though:

- CurrentLeases, TotalLeases, NumWorkers (DynamoDBLeaseRenewer, DynamoDBLeaseTaker)
- ExpiredLeases, VeryOldLeases (the snapshot-style emission sites in DynamoDBLeaseTaker, LeaseAssignmentManager)
- ActiveStreams.Count, StreamsPendingDeletion.Count (Scheduler)
- QueueSize (StreamIdCacheManager)
- NumStreamsToSync (PeriodicShardSyncManager)
- NumNewAssignmentsForWorker.*, LeaderLeasesByWorker.*, etc. (LeaseAssignmentManager)

Hard-coding that list of exceptions doesn't seem great though; it makes it harder for people to know to update it here when they add or change a metric of KCL.

Ideally, all of the callers of MetricsScope.addData could be more explicit about what meter type semantics they are using. Then OtelMetricsScope could use that info to choose the type to use (and CloudWatchMetricsScope could ignore it). But that would require changes to the MetricsScope interface, and changes to every caller of MetricsScope.addData. And I'm not sure how open the KCL maintainers are to that.

amazon-kinesis-client/pom.xml
```diff
@@ -275,6 +294,22 @@
...
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk-testing</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>net.jqwik</groupId>
+      <artifactId>jqwik</artifactId>
+      <version>1.7.4</version>
+      <scope>test</scope>
+    </dependency>
```
john-karp: Can all these be dropped? Looks like the tests just use `OpenTelemetry.noop` which I think is part of `opentelemetry-api`.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/MetricsConfig.java
```diff
@@ -106,19 +110,78 @@ public class MetricsConfig {
...
-                    cloudWatchClient(),
-                    namespace(),
-                    metricsBufferTimeMillis(),
-                    metricsMaxQueueSize(),
-                    metricsLevel(),
-                    metricsEnabledDimensions(),
-                    publisherFlushBuffer());
+            switch (metricsBackend) {
+                case OTEL:
+                    OpenTelemetry otel = openTelemetry != null ? openTelemetry : GlobalOpenTelemetry.get();
```
john-karp: There's references to `OpenTelemetry` and `GlobalOpenTelemetry` classes here; as a result I'm not sure that Java's classloader will be able to load MetricsConfig class without the OpenTelemetry API package being available, even if metricsBackend == CLOUDWATCH. If that's the case, we either need to make the API package non-optional, or have a way to construct OtelMetricsFactory here without having to directly interact with OpenTelemetry classes.

john-karp (commented):
Thanks for making the changes. I hope the maintainers accept it. 😅

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
wisemanIV: removed buffering.

amazon-kinesis-client/pom.xml
wisemanIV: removed

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
wisemanIV: Yeah, great call on the gauge vs counter distinction, those snapshot-style metrics (CurrentLeases, TotalLeases, QueueSize, etc.) really are gauges semantically. Agreed that hard-coding a list of exceptions here would be brittle and easy to forget when new metrics get added.

I think the right long-term fix is what you described - having callers of addData declare the instrument semantics explicitly, so OtelMetricsScope can pick the right instrument type. That would mean extending the MetricsScope interface (maybe an addGauge method or an instrument type hint parameter) and updating the call sites. CloudWatch side could just ignore the hint since PutMetricData doesn't distinguish.

That's a bigger change though, and probably better as a follow-up once we see how the maintainers feel about the current PR. For now the counter/histogram split by unit is a reasonable default that works for the majority of KCL metrics. I'll file an issue to track the gauge support as a future improvement.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/MetricsConfig.java
wisemanIV: you're right that the classloader would need to resolve those references when loading MetricsConfig regardless of which backend is selected.

I've made opentelemetry-api a regular (non-optional) dependency in the latest push. The API jar is small (~200KB), has zero transitive dependencies, and the noop implementation means there's zero runtime cost when no SDK is configured.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
```diff
@@ -0,0 +1,177 @@
...
+        addData(name, value, unit, MetricsLevel.DETAILED);
+    }
+
+    @Override
+    public void addData(String name, double value, StandardUnit unit, MetricsLevel level) {
+        checkNotEnded();
+        if (level.getValue() < metricsLevel.getValue()) {
+            return;
+        }
+        Attributes attrs = attributesBuilder.build();
```
john-karp: At this point this instance should be able to retain and re-use the built `attrs` - the contract for MetricsScope is that there should be no more dimensions added after the first addData.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/MetricsBackend.java
```diff
@@ -0,0 +1,42 @@
...
+    CLOUDWATCH,
+
+    /**
+     * Publish metrics via the CloudWatch OpenTelemetry (OTEL) endpoint in OTLP format.
+     *
+     * @deprecated Use {@link #OTEL} instead, which follows OTel library instrumentation best practices
+     *     by depending only on the OTel API and letting the application owner configure the SDK.
+     */
+    @Deprecated
+    CLOUDWATCH_OTEL,
```
john-karp: This enum value (and the enum as a whole) has never been included in a GA KCL version, so there shouldn't be any compatibility reasons to have this enum value at all.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
```diff
@@ -0,0 +1,177 @@
...
+        checkNotEnded();
+        ended = true;
+    }
+
+    private void recordObservation(String name, double value, StandardUnit unit, Attributes attrs) {
+        if (unit == StandardUnit.COUNT) {
+            LongCounter counter = meter.counterBuilder(name)
+                    .setUnit(convertUnit(unit))
+                    .build();
+            counter.add((long) value, attrs);
```
john-karp: Any reason not to use DoubleCounter?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
```diff
@@ -0,0 +1,177 @@
...
+        DIMENSION_TO_ATTRIBUTE_MAP = Collections.unmodifiableMap(dimMap);
+    }
+
+    /**
+     * Mapping of CloudWatch StandardUnit to OTel UCUM unit strings.
+     */
+    private static final Map<StandardUnit, String> UNIT_MAP;
+
+    static {
+        Map<StandardUnit, String> map = new HashMap<>();
```
john-karp: Nit: use EnumMap.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    @Override
+    public void addData(String name, double value, StandardUnit unit, MetricsLevel level) {
+        checkNotEnded();
+        if (level.getValue() < metricsLevel.getValue()) {
+            return;
+        }
+        Attributes attrs = attributesBuilder.build();
+        recordObservation(name, value, unit, attrs);
```
john-karp: We might want to normalize the meter names to more closely match otel norms. I think this is best to get right from the start, since once this is released, and people have built dashboards and set up alarms downstream, making changes will be less palatable.

Specific issues:

OTEL Convention | Compliance
-- | --
Instrument name is valid per BNF (no `:` etc.) | ❌ at least 3 names violate (`GetLease:Error`, `CurrentState:*`)
Lowercase + snake_case + dots | ❌ ~100% violate
Namespace prefix | ❌ ~100% violate
No unit in name | ❌ ~30% of names embed Time/Millis/Bytes/Count/Size
No _total/_count/_sum/_bucket suffix | ❌ 4+ names end in Count
Outcome variants in attributes, not separate metrics | ❌ ~30 separate `*.Success/Fault/*Error` series
State variants in attributes, not in name | ❌ `CurrentState:*`, `GsiReadyStatus`
No high-cardinality data in name | ✓
Plural noun for counters | ⚠️ inconsistent
.duration for time histograms | ❌ uses `.Time`
.size for byte histograms | ❌ uses `Bytes` suffix

I think 90% of that could be corrected with a `String transformName(String kclName)` or `String transformName(String kclName, StandardUnit unit)` helper that does, in order:

1. **Special-case rename map (lookup first; if hit, skip the rest of the rewriting steps and go straight to the namespace prefix).** Metrics whose mechanical transformation would either preserve the unit-in-name violation, end in a Prometheus-colliding suffix, or produce an unintelligible name. Examples: `MillisBehindLatest` → `consumer.lag.duration`, `DataBytesProcessed` → `records.size`, `ActiveStreams.Count` → `streams.active`, `PendingStreamsDeletion.Count` → `streams.pending_deletion`, `NonExistingStreamDelete.Count` → `streams.deleted_nonexistent`, `DeletedStreams.Count` → `streams.deleted`, `NumStreamsToSync` → `streams.pending_sync`, `QueueSize` → `stream_id_cache.queue.size`, `NumWorkers` → `workers`, `NumWorkersWithInvalidEntry` → `workers.invalid_entry`, `NumWorkersWithFailingWorkerMetric` → `workers.failing_worker_metric`. The map captures everywhere a simple rename is more useful than a literal token-by-token transform.

2. **Substitute `:` with `.`** to remove the BNF-illegal character. Catches `GetLease:Error`, `CurrentState:3xWorker`, `CurrentState:2xCompatibleWorker`, plus any future name that follows the same convention. (None of these reach this step if the special-case map catches them first; this is a generic safety net.)

3. **Replace a trailing `.Time` segment with `.duration`** to align with OTel's histogram-naming convention. Affects all `<op>.Time` and `<op>.<sub>.Time` metrics — roughly 20 latency series including `RenewLease.Time`, `TakeLease.Time`, `UpdateLease.Time`, `ListLeases.Time`, `KinesisDataFetcher.GetRecords.Time`, `RecordProcessor.processRecords.Time`, etc. Pure suffix swap, no semantic decisions.

4. **Strip a trailing `.Count` segment** if it survives the special-case map. Avoids collision with the Prometheus exporter's automatic `_total` suffix on counters and aligns with the "no exporter-suffix in the name" rule. The four `*.Count` metrics from `Scheduler` are already covered by the special-case map; this rule is defensive for any future additions.

5. **Split PascalCase / camelCase word boundaries within each dot-separated segment**, inserting `_` at lowercase→uppercase transitions. So `RecordsProcessed` → `records_processed`, `RecordProcessor.processRecords` → `record_processor.process_records`, `LeaseSpillover` → `lease_spillover`. Treat runs of consecutive uppercase as one word followed by a normal transition (`GSIReadyStatus` → `gsi_ready_status`, not `g_s_i_ready_status`). Preserve existing dots as segment separators rather than rewriting them.

6. **Lowercase the entire string** as the final case-folding step.

7. **Prepend the namespace prefix `aws.kinesis.client.`** to every result. One fixed prefix, no per-name decision; subsystem grouping is encoded by the rest of the name (which the special-case map shapes for the metrics where mechanical CamelCase-splitting wouldn't produce the desired grouping).

8. **Validate the final string against the OTel instrument-name BNF** (`[A-Za-z][A-Za-z0-9_.\-/]{0,254}`) and the 255-character length cap. If it fails (which shouldn't happen if the steps above are correct) log a `WARN`, replace any remaining illegal characters with `_`, and emit under the sanitized name. Provides a safety net for future contributors who add an unanticipated naming pattern.

**Caveat**: The transform produces valid OTel names but not always the most idiomatically grouped segmentation. `RenewLease.Time` becomes `aws.kinesis.client.renew_lease.duration` rather than the arguably-more-idiomatic `aws.kinesis.client.lease.renew.duration` (which would group neatly with `lease.create.duration`, `lease.take.duration`, etc. for dashboard discoverability). To get that grouping mechanically, you'd add restructuring rules to the special-case map.

The remaining 10% (outcome variants in attributes, not separate metrics; state variants in attributes, not in name; dual `MillisBehindLatest` scopes easily leading to downstream double-counting) would probably best be addressed with broader changes that affect the call sites. However those changes are not needed for correctness, just style/idiom. So I think we can justify deferring those for a later followup.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/MetricsConfig.java
john-karp: I think we are supposed to use `GlobalOpenTelemetry.getOrNoop()` here - `get()` has the JVM-global side effect of locking-in the no-op instance if one has not (yet) been `set`.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
```diff
@@ -0,0 +1,177 @@
...
+
+    @Override
+    public void end() {
+        checkNotEnded();
+        ended = true;
+    }
+
+    private void recordObservation(String name, double value, StandardUnit unit, Attributes attrs) {
+        if (unit == StandardUnit.COUNT) {
+            LongCounter counter = meter.counterBuilder(name)
```
john-karp: Hm, my opinion has shifted a bit here. Until callers can pass in meter-type arguments to this class, I think it is worth hard-coding a set of metric names that should be recorded as gauges instead of counters. Because the definite pain resulting from recording metrics with improper types outweighs the risk that the set drifts away from reality. And even if there is drift, the outcome is not necessarily particularly worse than the status quo.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
```diff
@@ -0,0 +1,177 @@
...
+                    .setUnit(convertUnit(unit))
+                    .build();
+            counter.add((long) value, attrs);
+        } else {
+            DoubleHistogram histogram = meter.histogramBuilder(name)
+                    .setUnit(convertUnit(unit))
+                    .build();
```
john-karp: Note that in OTEL same-name-different-unit instruments are considered semantic errors, producing warnings, and possibly causing downstream exporters/backend to choke. If we want, we could catch that via a ConcurrentHashMap<String, Instrument> in OtelMetricsFactory, which would make it easier to discover such cases via CI. It might also improve performance a tiny bit. But I don't think its strictly necessary.

john-karp (commented):
I thought of a few more things to consider. Also, I think overall the unit test coverage might be worth taking a look at.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
wisemanIV: agreed fixed in next push

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/MetricsBackend.java
wisemanIV: agreed. fixed in next push.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
wisemanIV: Agreed. fixed in next push.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
wisemanIV: Agreed. Fixed in next push.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
wisemanIV: thanks for the detail. this version has implemented this transformation - 
Step 1: Special-case rename map with all 11 entries	✅ All 11 present
Step 2: Substitute : with .	✅ name.replace(':', '.')
Step 3: Replace trailing .Time with .duration	✅
Step 4: Strip trailing .Count	✅
Step 5: Split PascalCase/camelCase, uppercase runs as one word	✅ splitSegment handles both transitions
Step 6: Lowercase	✅
Step 7: Prepend aws.kinesis.client.	✅
Step 8: Validate BNF + 255 char limit	✅
Special-case hits skip steps 2-5	✅
transformName(String, StandardUnit) overload	✅ (delegates, reserved for future)

Agreed on the deferred scope.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
wisemanIV: deferring this.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
wisemanIV: OtelMetricsScope has a GAUGE_METRIC_NAMES set with 15 entries:

CurrentLeases, TotalLeases, NumWorkers, ExpiredLeases, VeryOldLeases,
ActiveStreams.Count, StreamsPendingDeletion.Count, QueueSize, NumStreamsToSync,
NumWorkersWithInvalidEntry, NumWorkersWithFailingWorkerMetric,
GsiReadyStatus, WorkerMetricsReadyStatus,
CurrentState:3xWorker, CurrentState:2xCompatibleWorker

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/MetricsConfig.java
wisemanIV: fixed in next push.

wisemanIV:
> I thought of a few more things to consider. Also, I think overall the unit test coverage might be worth taking a look at.

Thanks for the feedback @john-karp. Suggested changes implemented and unit tests extended.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
```diff
@@ -0,0 +1,213 @@
...
+        dimMap.put(MetricsUtil.STREAM_IDENTIFIER, "aws.kinesis.stream_name");
+        dimMap.put("WorkerIdentifier", "aws.kinesis.consumer.name");
+        DIMENSION_TO_ATTRIBUTE_MAP = Collections.unmodifiableMap(dimMap);
+    }
+
+    /**
+     * KCL metric names (original, pre-transformation) that should be recorded as gauges
+     * rather than counters or histograms.
+     */
+    static final Set<String> GAUGE_METRIC_NAMES = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
```
lucienlu-aws, line 62: What is the behavior if we add a metric to KCL and don't include it here? Would it be recorded as a counter in OTEL?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricNameTransformer.java
```diff
@@ -0,0 +1,207 @@
...
+ *   <li>Strip trailing {@code .Count}</li>
+ *   <li>Split PascalCase/camelCase — insert {@code _} at lowercase→uppercase transitions
+ *       within each dot-separated segment, treating runs of consecutive uppercase as one word.</li>
+ *   <li>Lowercase the entire string.</li>
+ *   <li>Prepend namespace prefix {@code aws.kinesis.client.}</li>
+ *   <li>Validate against OTel BNF and 255 char limit.</li>
+ * </ol>
+ */
+@Slf4j
+class OtelMetricNameTransformer {
```
lucienlu-aws, line 44: Since this is implicitly protected I'm assuming this can have backwards incompatible changes internally? Can you mark this as `@KinesisClientInternalAPI`?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricNameTransformer.java
```diff
@@ -0,0 +1,207 @@
...
+        map.put("ActiveStreams.Count", "streams.active");
+        map.put("StreamsPendingDeletion.Count", "streams.pending_deletion");
+        map.put("NonExistingStreamDelete.Count", "streams.deleted_nonexistent");
+        map.put("DeletedStreams.Count", "streams.deleted");
+        map.put("NumStreamsToSync", "streams.pending_sync");
+        map.put("QueueSize", "stream_id_cache.queue.size");
+        map.put("NumWorkers", "workers");
+        map.put("NumWorkersWithInvalidEntry", "workers.invalid_entry");
+        map.put("NumWorkersWithFailingWorkerMetric", "workers.failing_worker_metric");
+        SPECIAL_CASE_MAP = Collections.unmodifiableMap(map);
```
lucienlu-aws, line 73: What happens if we add a metric that is not in the map and does not comply with OTEL names? Considering the case where we add a metric and forget to add it here and the `transformName` method transforms it to something OTEL doesn't like. Does it not send the metric at all?

lucienlu-aws (commented):
Change generally looks good to me. Have you tested this in a running KCL app with OTEL configured and verified it works on OTEL end already?

wisemanIV:
> Change generally looks good to me. Have you tested this in a running KCL app with OTEL configured and verified it works on OTEL end already?

Yes. We deployed a sample KCL app on ECS Fargate (account 322962300869, us-east-1) using MetricsBackend.OTEL with the 3.4.3-SNAPSHOT from this branch. The app configures an OTel SDK SdkMeterProvider with a custom MetricExporter that sends OTLP/JSON to CloudWatch's monitoring.us-east-1.amazonaws.com/v1/metrics endpoint with SigV4 signing. We confirmed:

34 KCL metrics are being exported successfully (HTTP 200 from CloudWatch)
Metrics are queryable via CloudWatch's PromQL endpoint at /api/v1/query
Metric names follow the aws.kinesis.client.* convention (e.g., aws.kinesis.client.records_processed, aws.kinesis.client.record_processor.process_records.duration, aws.kinesis.client.current_leases)
Labels include aws.kinesis.operation, aws.kinesis.shard.id, @resource.service.name
A Grafana dashboard visualizes the metrics end-to-end

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
wisemanIV: If a new metric is added and NOT in GAUGE_METRIC_NAMES:

If its unit is StandardUnit.COUNT → it's recorded as a DoubleCounter (monotonically increasing)
If its unit is anything else (MILLISECONDS, BYTES, etc.) → it's recorded as a DoubleHistogram
So yes, a metric with StandardUnit.COUNT that isn't in the gauge set would be a counter. This is correct for most count metrics (RecordsProcessed, Success, etc.) but would be wrong for point-in-time counts like "CurrentLeases" — which is why those are explicitly listed in GAUGE_METRIC_NAMES.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricNameTransformer.java
wisemanIV: The metric is still sent — it's never dropped. The validate() method sanitizes illegal characters by replacing them with _ and logs a warning. So the worst case is a slightly mangled metric name with a log warning, not a lost metric. The OTel SDK itself is also lenient — it will accept most names and only reject truly malformed ones at the exporter level (which would log an error but not crash).

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricNameTransformer.java
wisemanIV: yes agreed. pushing

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    @Override
+    public void addData(String name, double value, StandardUnit unit, MetricsLevel level) {
+        checkNotEnded();
+        if (level.getValue() < metricsLevel.getValue()) {
+            return;
+        }
+        if (cachedAttributes == null) {
+            cachedAttributes = attributesBuilder.build();
```
john-karp, line 151: Apologies, I think one of my earlier comments was wrong. In a couple cases KCL code does call `addDimension` on a context after `addData`, via `addWorkerIdentifier`:

 - https://github.com/awslabs/amazon-kinesis-client/blob/0395a672ce291ec467bbc2a074edc29ae0e09ab1/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseRenewer.java#L160-L175
 - https://github.com/awslabs/amazon-kinesis-client/blob/0395a672ce291ec467bbc2a074edc29ae0e09ab1/amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseDiscoverer.java#L60-L105

I'm not sure why the worker identifier is being added in the `finally` blocks instead of immediately after the MetricsScope; @lucienlu-aws - do you happen to know?

wisemanIV: Good catch — this is a real behavioral difference between the CloudWatch and OTel scopes. The CloudWatch scope accumulates everything and applies dimensions to all metrics at end() time, so ordering doesn't matter. The OTel scope records observations immediately on addData, so dimensions added later (like WorkerIdentifier in the finally block) won't be attached to earlier observations.

Two options to fix:

Option A: Buffer observations until end()

Change OtelMetricsScope to buffer (name, value, unit) tuples during addData and defer all recordObservation calls to end(), at which point all dimensions have been accumulated. This makes the OTel scope order-independent, matching the CloudWatch scope's behavior.

Pro: Zero changes to existing KCL call sites. Resilient to future code that adds dimensions late.
Con: Holds observations in memory until end(). Slightly more complex scope implementation.
Option B: Move addDimension calls before addData at call sites

Move MetricsUtil.addWorkerIdentifier(scope, workerIdentifier) from the finally block to immediately after scope creation in the affected call sites (DynamoDBLeaseRenewer, DynamoDBLeaseDiscoverer, and any others with this pattern). The CloudWatch scope is unaffected by this reordering.

Pro: Simpler scope implementation (record-immediately stays). Explicit about the contract.
Con: Touches multiple files. Doesn't protect against future code accidentally adding dimensions late.

I'd lean toward Option B as cleaner. But happy to go either way. What's your preference?

john-karp: I'd lean towards option B, and adding to the MetricsScope javadoc a note about no addDimension after addData.

Optionally we could detect in addDimension when that contract is broken (cachedAttributes is already set) and log a warning, but that might be overkill? Overall I wouldn't describe MetricsScope as particularly rigorous. (For example it doesn't implement Java's AutoCloseable, which would have been a good way to help callers reliably trigger `end()`, which hasn't always happened - https://github.com/awslabs/amazon-kinesis-client/pull/1734)

wisemanIV: The PR now has two new commits:

0c10303 — @KinesisClientInternalApi annotation on OtelMetricNameTransformer
bb5bb97 — Dimension ordering fix: moved addWorkerIdentifier before addData in DynamoDBLeaseRenewer, added javadoc contract + AutoCloseable TODO to MetricsScope, and 3 new unit tests

wisemanIV:
Retested with latest changes:
Screenshot 2026-05-19 at 09 16 25

amazon-kinesis-client/src/main/java/software/amazon/kinesis/metrics/OtelMetricsScope.java
lucienlu-aws: I see this thread talks about a few places where `addDimension` is called after `addData`, but the new commit only updates DynamoDBLeaseRenewer. Do we need to update it in other places like DynamoDBLeaseDiscoverer as @john-karp called out?

wisemanIV: @lucienlu-aws all done now.

lucienlu-aws (approved):

PR #1729, opened 2026-04-09
Ensure heartbeat is stopped for graceful handoff lease in shutting do…

chenylee-aws:
*Description of changes:*
Fix graceful lease handoff when a ShardConsumer receives a graceful handoff request before reaching the initialized state. Due to lifecycle management, the ShardConsumer transitions directly from WAITING_ON_PARENT_SHARDS to shutdown, skipping ShutdownNotificationTask, leaving the lease never transferred or dropped.

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
@@ -592,7 +592,18 @@ public class Scheduler implements Runnable {
             for (ShardInfo shardInfo : getShardInfoForAssignments()) {
                 ShardConsumer shardConsumer = createOrGetShardConsumer(
                         shardInfo, processorConfig.shardRecordProcessorFactory(), leaseCleanupManager);
-
+                if (shardConsumer.isShutdown()) {
+                    final Lease currentLease = leaseCoordinator.getCurrentlyHeldLease(ShardInfo.getLeaseKey(shardInfo));
+                    if (currentLease != null
+                            && currentLease.concurrencyToken() != null
+                            && shardInfo
+                                    .concurrencyToken()
+                                    .equals(currentLease.concurrencyToken().toString())) {
+                        // dropping the lease to stop heartbeat so it gets cleaned up in the next round
+                        log.warn("Unexpected that lease is shutdown but still held. Stop heartbeat {}", currentLease);
+                        leaseCoordinator.dropLease(currentLease);
+                    }
+                }
                 shardConsumer.executeLifecycle();
                 assignedShards.add(shardInfo);
             }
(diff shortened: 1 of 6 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -592,7 +592,18 @@ void runProcessLoop() {
             for (ShardInfo shardInfo : getShardInfoForAssignments()) {
                 ShardConsumer shardConsumer = createOrGetShardConsumer(
                         shardInfo, processorConfig.shardRecordProcessorFactory(), leaseCleanupManager);
-
+                if (shardConsumer.isShutdown()) {
```
lucienlu-aws, line 595: With this commit, both ShutdownTask and ShutdownNotificationTask already drop the lease from the leaseCoordinator but is this for extra safety?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/LeaseGracefulShutdownHandler.java
```diff
@@ -150,8 +150,7 @@ private void monitorGracefulShutdownLeases() {
                 final ShardInfo shardInfo = entry.getKey();
                 leaseKey = leasePendingShutdown.lease.leaseKey();
 
-                if (leasePendingShutdown.shardConsumer.isShutdown()
```
lucienlu-aws, line 153: Why remove this check?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
chenylee-aws: Yes, this is an extra safety net. We don't expect it to be executed after this change

amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/LeaseGracefulShutdownHandler.java
chenylee-aws: The leasePendingShutdown.shardConsumer.isShutdown() check was redundant with the other two conditions. When a ShardConsumer reaches terminal state (isShutdown() == true),  lease should also have been dropped from OwnedLeases so the getCurrentlyHeldLease should return null unless the worker regained anther lease with different concurrency token. In either case the heartbeat should have stopped and the scheduler should have removed the shard from the currentAssignment map. 

I am removing it because due to the bad false shutdown issue, this check assumed lease is properly shutdown and prevented the leaseGracefulHandler from transferring the leases which would have stopped the heartbeat.

lucienlu-aws (approved):

PR #1733, opened 2026-04-21
Fix checkpointing logic to handle shard end correctly #211

john-karp:
*Issue #, if available:* #211

While the issue was originally reported a long time ago, I see this issue in production with KCL 3.3.0.

In ShardRecordProcessorCheckpointer.advancePosition(ExtendedSequenceNumber), the guard that suppresses re‑checkpointing an already‑checkpointed value compared the caller‑supplied extendedSequenceNumber against lastCheckpointValue. But just above that guard, the method may rewrite the checkpoint target to SHARD_END when the sequence number matches sequenceNumberAtShardEnd. Because the guard still looked at the pre‑rewrite value, the following very common shard‑end sequence was broken:

1. Record processor checkpoints the last record N during processRecords → lastCheckpointValue = N.
2. Framework transitions the shard: sequenceNumberAtShardEnd = N, largestPermittedCheckpointValue = SHARD_END.
3. On shutdown the app calls checkpoint(N), which re‑enters advancePosition(N). checkpointToRecord is rewritten to SHARD_END, but the guard evaluated !N.equals(N) == false and short‑circuited — SHARD_END was never persisted and lastCheckpointValue stayed at N.
4. ShutdownTask.applicationCheckpointAndVerification then raises IllegalArgumentException: Application didn't checkpoint at end of shard 

*Description of changes:*

One-line change in ShardRecordProcessorCheckpointer. Compare checkpointToRecord (the post‑rewrite value that will actually be persisted) instead of extendedSequenceNumber. This is the exact fix proposed by @posac in the original 2017 upstream issue (and endorsed by @klesniewski in 2018), which was never merged.

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/checkpoint/ShardRecordProcessorCheckpointer.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/checkpoint/ShardRecordProcessorCheckpointer.java
@@ -300,8 +300,13 @@ public class ShardRecordProcessorCheckpointer implements RecordProcessorCheckpoi
             checkpointToRecord = ExtendedSequenceNumber.SHARD_END;
         }
 
-        // Don't checkpoint a value we already successfully checkpointed
-        if (extendedSequenceNumber != null && !extendedSequenceNumber.equals(lastCheckpointValue)) {
+        // Don't checkpoint a value we already successfully checkpointed. Compare against
+        // checkpointToRecord (which may have been rewritten to SHARD_END above) rather than the
+        // caller-supplied extendedSequenceNumber; otherwise the transition to SHARD_END is skipped
+        // when the last record's sequence number was already checkpointed, causing the shard-end
+        // shutdown to throw "Application didn't checkpoint at end of shard" in a loop. See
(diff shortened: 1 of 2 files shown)


john-karp:
Apparently a similar PR was [rejected before](https://github.com/awslabs/amazon-kinesis-client/pull/345#pullrequestreview-146180506) with this reasoning:

> Blanket checkpointing around shard end could be dangerous. Children shard get picked up for processing only when the parent has been marked as SHARD_END. This could lead into a situation where the parent has been shutdown because of shard end and the children aren't picked up for processing due to the checkpoint of the parent not being SHARD_END leading to data loss.

The problem situation described in this comment is actually one that the current KCL code fails to handle, and the one-line change of this PR (and the prior one) should fix it. In advancePosition, checkpointToRecord only differs from extendedSequenceNumber when "we are about to checkpoint the very last sequence number for this shard" (existing code comment), so in this case the effect of using checkpointToRecord for the guard instead of extendedSequenceNumber is that we checkpoint it at SHARD_END, instead of failing to checkpoint. The previous behavior of failing to checkpoint in this case is what actually leads to children being blocked.

PR #1736, opened 2026-04-28
implement rolling deployment system for changes to lease assignment algorithm

vincentvilo-aws:
*Issue #, if available:*

*Description of changes:* Creating a rolling deployment system that will enable the lease assignment algorithm being deployed to only apply to workers that have received the changes.  

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
@@ -114,6 +114,7 @@ import software.amazon.kinesis.retrieval.RecordsPublisher;
 import software.amazon.kinesis.retrieval.RetrievalConfig;
 import software.amazon.kinesis.retrieval.polling.PollingConfig;
 import software.amazon.kinesis.schemaregistry.SchemaRegistryDecoder;
+import software.amazon.kinesis.segmenting.FleetSegmentingHandler;
 import software.amazon.kinesis.worker.WorkerMetricsSelector;
 import software.amazon.kinesis.worker.metricstats.WorkerMetricStatsDAO;
 import software.amazon.kinesis.worker.metricstats.WorkerMetricStatsManager;
@@ -225,6 +226,10 @@ public class Scheduler implements Runnable {
 
     private boolean leasesSyncedOnAppInit = false;
 
+    private final FleetSegmentingHandler segmentingHandler;
+
+    private final WorkerMetricStatsDAO workerMetricsDAO;
+
     @Getter(AccessLevel.NONE)
     private final AtomicBoolean leaderSynced = new AtomicBoolean(false);
 
@@ -322,6 +327,17 @@ public class Scheduler implements Runnable {
         this.leaseRefresher = this.leaseCoordinator.leaseRefresher();
 
         this.leaseAssignmentModeProvider = new MigrationAdaptiveLeaseAssignmentModeProvider();
+
+        this.workerMetricsDAO = new WorkerMetricStatsDAO(
+                leaseManagementConfig.dynamoDBClient(),
+                leaseManagementConfig.workerUtilizationAwareAssignmentConfig().workerMetricsTableConfig(),
+                leaseManagementConfig.workerUtilizationAwareAssignmentConfig().workerMetricsReporterFreqInMillis());
+
+        this.segmentingHandler = new FleetSegmentingHandler(
+                leaseManagementConfig,
+                new DynamoDbAsyncToSyncClientAdapter(leaseManagementConfig.dynamoDBClient()),
+                coordinatorConfig.coordinatorStateTableConfig().tableName());
+
         this.migrationComponentsInitializer = createDynamicMigrationComponentsInitializer();
         this.migrationStateMachine = new MigrationStateMachineImpl(
                 metricsFactory,
@@ -394,7 +410,8 @@ public class Scheduler implements Runnable {
                 metricsFactory,
                 leaseManagementConfig.leasesRecoveryAuditorExecutionFrequencyMillis(),
                 leaseManagementConfig.leasesRecoveryAuditorInconsistencyConfidenceThreshold(),
-                leaderSynced);
+                leaderSynced,
+                segmentingHandler);
         this.leaseCleanupManager = leaseManagementFactory.createLeaseCleanupManager(metricsFactory);
         this.schemaRegistryDecoder = this.retrievalConfig.glueSchemaRegistryDeserializer() == null
                 ? null
@@ -415,11 +432,6 @@ public class Scheduler implements Runnable {
                 leaseManagementConfig
                         .workerUtilizationAwareAssignmentConfig()
                         .inMemoryWorkerMetricsCaptureFrequencyMillis());
-
-        final WorkerMetricStatsDAO workerMetricsDAO = new WorkerMetricStatsDAO(
-                leaseManagementConfig.dynamoDBClient(),
-                leaseManagementConfig.workerUtilizationAwareAssignmentConfig().workerMetricsTableConfig(),
-                leaseManagementConfig.workerUtilizationAwareAssignmentConfig().workerMetricsReporterFreqInMillis());
 
         return DynamicMigrationComponentsInitializer.builder()
                 .metricsFactory(metricsFactory)
@@ -447,8 +459,10 @@ public class Scheduler implements Runnable {
                         System::nanoTime,
                         leaseManagementConfig.maxLeasesForWorker(),
                         leaseManagementConfig.gracefulLeaseHandoffConfig(),
+                        leaseManagementConfig.leaseAssignmentStrategy(),
                         leaseManagementConfig.leaseAssignmentIntervalMillis(),
-                        streamIdCacheManager))
+                        streamIdCacheManager,
+                        segmentingHandler))
                 .adaptiveLeaderDeciderCreator(() -> new MigrationAdaptiveLeaderDecider(metricsFactory))
                 .deterministicLeaderDeciderCreator(() -> new DeterministicShuffleShardSyncLeaderDecider(
                         leaseRefresher, Executors.newSingleThreadScheduledExecutor(), 1, metricsFactory))
@@ -457,10 +471,12 @@ public class Scheduler implements Runnable {
                         leaseCoordinator.workerIdentifier(),
                         metricsFactory,
                         leaseManagementConfig.dynamoDbLockBasedLeaderLeaseDurationInMillis(),
-                        leaseManagementConfig.dynamoDbLockBasedLeaderHeartbeatPeriodInMillis()))
+                        leaseManagementConfig.dynamoDbLockBasedLeaderHeartbeatPeriodInMillis(),
+                        segmentingHandler))
                 .workerIdentifier(leaseCoordinator.workerIdentifier())
                 .workerUtilizationAwareAssignmentConfig(leaseManagementConfig.workerUtilizationAwareAssignmentConfig())
                 .leaseAssignmentModeProvider(leaseAssignmentModeProvider)
+                .segmentingHandler(segmentingHandler)
                 .build();
     }
 
@@ -605,6 +621,7 @@ public class Scheduler implements Runnable {
                 checkAndSyncStreamShardsAndLeases();
                 leaderSynced.set(true);
                 streamInfoManager.start();
+                segmentingHandler.updateCurrentVersion();
             } else {
                 leaderSynced.set(false);
                 streamInfoManager.stop(false);
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
@@ -335,7 +335,6 @@ public class Scheduler implements Runnable {
 
         this.segmentingHandler = new FleetSegmentingHandler(
                 leaseManagementConfig,
-                new DynamoDbAsyncToSyncClientAdapter(leaseManagementConfig.dynamoDBClient()),
                 coordinatorConfig.coordinatorStateTableConfig().tableName(),
                 coordinatorStateDAO);
 
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
@@ -336,7 +336,8 @@ public class Scheduler implements Runnable {
         this.segmentingHandler = new FleetSegmentingHandler(
                 leaseManagementConfig,
                 new DynamoDbAsyncToSyncClientAdapter(leaseManagementConfig.dynamoDBClient()),
-                coordinatorConfig.coordinatorStateTableConfig().tableName());
+                coordinatorConfig.coordinatorStateTableConfig().tableName(),
+                coordinatorStateDAO);
 
         this.migrationComponentsInitializer = createDynamicMigrationComponentsInitializer();
         this.migrationStateMachine = new MigrationStateMachineImpl(
@@ -621,7 +622,6 @@ public class Scheduler implements Runnable {
                 checkAndSyncStreamShardsAndLeases();
                 leaderSynced.set(true);
                 streamInfoManager.start();
-                segmentingHandler.updateCurrentVersion();
             } else {
                 leaderSynced.set(false);
                 streamInfoManager.stop(false);
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
@@ -333,10 +333,7 @@ public class Scheduler implements Runnable {
                 leaseManagementConfig.workerUtilizationAwareAssignmentConfig().workerMetricsTableConfig(),
                 leaseManagementConfig.workerUtilizationAwareAssignmentConfig().workerMetricsReporterFreqInMillis());
 
-        this.segmentingHandler = new FleetSegmentingHandler(
-                leaseManagementConfig,
-                coordinatorConfig.coordinatorStateTableConfig().tableName(),
-                coordinatorStateDAO);
+        this.segmentingHandler = new FleetSegmentingHandler(leaseManagementConfig, coordinatorStateDAO);
 
         this.migrationComponentsInitializer = createDynamicMigrationComponentsInitializer();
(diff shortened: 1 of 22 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseCountBasedLeaseAssignmentDecider.java
```diff
@@ -0,0 +1,268 @@
...
+     * Assigns expired or unassigned leases to workers with the fewest current leases.
+     * Leases are sorted by lastCounterIncrementNanos to ensure consistent assignment order.
+     *
+     * @param expiredOrUnAssignedLeases list of leases that need to be assigned to workers
+     */
+    @Override
+    public void assignExpiredOrUnassignedLeases(final List<Lease> expiredOrUnAssignedLeases) {
+        if (expiredOrUnAssignedLeases.isEmpty()
+                || inMemoryStorageView.getActiveWorkerIdSet().isEmpty()) {
+            return;
```
ashwv-aws, line 61: Minor - would we need to publish a metrics or a log statement for this specific case?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseCountBasedLeaseAssignmentDecider.java
```diff
@@ -0,0 +1,268 @@
...
+            }
+        }
+    }
+
+    /**
+     * Calculates the target number of leases each worker should have.
+     *
+     * @return target lease count per worker, capped by maxLeasesForWorker
+     */
+    private int calculateTargetLeaseCount(final int totalLeases) {
```
ashwv-aws, line 173: Minor - `calculateTargetLeaseCountPerWorker`

amazon-kinesis-client/src/main/java/software/amazon/kinesis/segmenting/FleetSegmentingHandler.java
```diff
@@ -0,0 +1,161 @@
...
+@KinesisClientInternalApi
+public class FleetSegmentingHandler {
+
+    private static final String CURRENT_VERSION_KEY = "CurrentVersion";
+
+    @Getter
+    private final String versionHash;
+
+    @Getter
+    private final String versionHashKey = "versionHash";
```
chenylee-aws: nit: static final and you can just make it a public variable


Also should this be lamVersionHash since we are only considering the assignment agorithm ?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/segmenting/FleetSegmentingHandler.java
```diff
@@ -0,0 +1,161 @@
...
+
+    @Getter
+    private boolean isVersionEmittedByAllActiveWorkers = false;
+
+    @Getter
+    private final boolean isEnabled;
+
+    private final long versionHashExpiryMillis = TimeUnit.HOURS.toMillis(1);
```
chenylee-aws: static final and please add a comment on why 1 hours is chosen

amazon-kinesis-client/src/main/java/software/amazon/kinesis/segmenting/FleetSegmentingHandler.java
```diff
@@ -0,0 +1,161 @@
...
+    private final String versionHashLutKey = "versionHashLut";
```
chenylee-aws: static and camal case

amazon-kinesis-client/src/main/java/software/amazon/kinesis/segmenting/FleetSegmentingHandler.java
```diff
@@ -0,0 +1,161 @@
...
+    private final String leaderTableName;
+    private final DynamoDbClient ddbClient;
+
+    public FleetSegmentingHandler(
+            final LeaseManagementConfig config, final DynamoDbClient ddbClient, final String leaderTableName) {
+        this.leaderTableName = leaderTableName;
+        this.ddbClient = ddbClient;
+        this.versionHash =
+                String.valueOf(config.leaseAssignmentStrategy().name().hashCode());
```
chenylee-aws: why do we use hash instead of just numbering the enums? i think numbering will give better mapping and and more readable.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leader/DynamoDBLockBasedLeaderDecider.java
```diff
@@ -114,20 +123,29 @@ public synchronized Boolean isLeader(final String workerId) {
             return lastIsLeaderResult;
         }
         boolean response;
-        // Get the lockItem from storage (if present
-        final Optional<LockItem> lockItem = dynamoDBLockClient.getLock(LEADER_HASH_KEY, Optional.empty());
-        lockItem.ifPresent(item -> log.info("Worker : {} is the current leader.", item.getOwnerName()));
+
+        // before obtaining a leader lock, release any lock that does not match the worker's version
+        releaseLockIfVersionHashDoesNotMatch(CoordinatorState.LEADER_HASH_KEY);
```
chenylee-aws: why would version not match if worker owns the lock?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leader/DynamoDBLockBasedLeaderDecider.java
```diff
@@ -149,9 +167,42 @@ public synchronized Boolean isLeader(final String workerId) {
...
+    private void releaseLockIfVersionHashDoesNotMatch(final String lockKey) {
+        final Optional<LockItem> lockItem = dynamoDBLockClient.getLock(lockKey, Optional.empty());
+        if (lockItem.isPresent() && !doesVersionHashMatchLockVersionHash(lockItem.get())) {
+            dynamoDBLockClient.releaseLock(lockItem.get());
+        }
+    }
+
+    private void releaseDeployingLeaderLockIfDeployingLeaderIsNotNeeded() {
+        if (isWorkerLeader(CoordinatorState.LEADER_HASH_KEY)
+                        && isWorkerLeader(CoordinatorState.DEPLOYING_LEADER_HASH_KEY)
```
chenylee-aws: when does it happen? are we allowing worker to be both current and deploying leader?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/segmenting/FleetSegmentingHandler.java
```diff
@@ -0,0 +1,161 @@
...
+                && item.containsKey(versionHashLutKey)
+                && !isVersionHashExpired(
+                        Long.parseLong(item.get(versionHashLutKey).s()));
+    }
+
+    private boolean isVersionHashExpired(final long lut) {
+        return Duration.between(Instant.ofEpochSecond(lut), Instant.now()).toMillis() > versionHashExpiryMillis;
+    }
+
+    private GetItemResponse getItemFromCoordinatorTable(final String key) {
```
chenylee-aws: is there no exsiting item access method from the coordinator table dao?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/segmenting/FleetSegmentingHandler.java
```diff
@@ -0,0 +1,161 @@
...
+        if (lutStr == null) {
+            return true;
+        }
+        return isVersionHashExpired(Long.parseLong(lutStr));
+    }
+
+    public void setIsVersionEmittedByAllActiveWorkers(
+            final List<WorkerMetricStats> activeWorkerMetrics, final List<WorkerMetricStats> workersOnVersionHash) {
+        isVersionEmittedByAllActiveWorkers =
+                new HashSet<>(activeWorkerMetrics).equals(new HashSet<>(workersOnVersionHash));
```
chenylee-aws, line 147: can just do containAll

amazon-kinesis-client/src/main/java/software/amazon/kinesis/segmenting/FleetSegmentingHandler.java
```diff
@@ -0,0 +1,161 @@
...
+    private static final String CURRENT_VERSION_KEY = "CurrentVersion";
+
+    @Getter
+    private final String versionHash;
+
+    @Getter
+    private final String versionHashKey = "versionHash";
+
+    @Getter
+    private boolean isVersionEmittedByAllActiveWorkers = false;
```
chenylee-aws: should it be volatile?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leader/DynamoDBLockBasedLeaderDecider.java
vincentvilo-aws: During testing, there were cases where a worker has the Leader lock, the same worker gets upgraded, and its version hash is updated. When the worker was upgraded, it never released the Leader lock, so the worker was in a state where it is the leader but its version hash did not match.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/segmenting/FleetSegmentingHandler.java
```diff
@@ -0,0 +1,161 @@
...
+     */
+    public void updateCurrentVersion() {
+        if (isVersionEmittedByAllActiveWorkers) {
+            Map<String, AttributeValue> currentVersionMap = getVersionHashWithLastUpdatedTimeForLockTable();
+            currentVersionMap.put("key", AttributeValue.fromS(CURRENT_VERSION_KEY));
+            final PutItemRequest putItemRequest = PutItemRequest.builder()
+                    .item(currentVersionMap)
+                    .tableName(leaderTableName)
+                    .build();
+            ddbClient.putItem(putItemRequest);
```
chenylee-aws: this should not be in the class IMO

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leader/DynamoDBLockBasedLeaderDecider.java
vincentvilo-aws: When deployment is finished, the current version is changed so that it matches the deploying version. Once all workers are emitting the same version, the workers will contend for the Leader lock. In this case, the current version will be the same as version hash of the deploying leader. Sometimes one worker would have both leader locks due to the order of which hosts received the deployment. For this case, we want to make sure that the deploying leader is released

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -618,6 +621,7 @@ void runProcessLoop() {
                 checkAndSyncStreamShardsAndLeases();
                 leaderSynced.set(true);
                 streamInfoManager.start();
+                segmentingHandler.updateCurrentVersion();
```
chenylee-aws: This should be done as part of the leader heartbeat and lock acquisition

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leader/DynamoDBLockBasedLeaderDecider.java
chenylee-aws: this should be avoidable if you attach hash as part of lock acquisition

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leader/DynamoDBLockBasedLeaderDecider.java
chenylee-aws: i feel like we shouldn't get into the two leader state. maybe drop one lease before you acquire the other one

amazon-kinesis-client/src/main/java/software/amazon/kinesis/segmenting/FleetSegmentingHandler.java
vincentvilo-aws: For now the version hash is only created based off the assignment algorithm, but in the future, the version hash could be a function of other lease management config settings, so I feel like it makes sense to leave it as `versionHash`

amazon-kinesis-client/src/main/java/software/amazon/kinesis/segmenting/FleetSegmentingHandler.java
vincentvilo-aws: Going to modify the class to use the coordinator state dao for accessing the coordinator table

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseCountBasedLeaseAssignmentDecider.java
vincentvilo-aws: Since this PR is going to focus on introducing the segmenting handler, i'll hold off on addressing the comments related to the lease count decider

amazon-kinesis-client/src/main/java/software/amazon/kinesis/segmenting/FleetSegmentingHandler.java
vincentvilo-aws: Changed this to be a function of the worker metric expiry threshold and the leader lease duration:

```
        // 2 * workerMetricsReporterFreqInMillis is the current threshold for considering a WorkerMetricStats to
        // be expired. Since the versionHashExpiry is also used when checking if a leader's version hash is stale,
        // we will take the max of the leader lease duration and the stale metrics threshold.
        this.versionHashExpiryMillis = Math.max(
                config.dynamoDbLockBasedLeaderLeaseDurationInMillis(),
                2 * config.workerUtilizationAwareAssignmentConfig().workerMetricsReporterFreqInMillis());
```

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseAssignmentManager.java
```diff
@@ -223,20 +227,28 @@ private void performAssignment() {
...
-
-            if (!expiredOrUnAssignedLeases.isEmpty()) {
-                // When expiredOrUnAssignedLeases is not empty, that means
-                // that we were not able to assign all expired or unassigned leases and hit the maxThroughput
-                // per worker for all workers.
-                log.warn("Not able to assign all expiredOrUnAssignedLeases");
-                metricsScope.addData(
-                        "LeaseSpillover", expiredOrUnAssignedLeases.size(), StandardUnit.COUNT, MetricsLevel.SUMMARY);
+
+            if (!segmentingHandler.isEnabled() || segmentingHandler.isOnCurrentVersion()) {
```
chenylee-aws: nit: encapsulate this logic in a method since it's also used somewhere else.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseAssignmentManager.java
```diff
@@ -716,6 +728,33 @@ public Double getTotalAssignedThroughput(final String workerId) {
...
+        public List<WorkerMetricStats> getWorkersOnVersionHash() {
+            if (!segmentingHandler.isEnabled()) {
+                return activeWorkerMetrics;
+            }
+            return activeWorkerMetrics.stream()
+                    .filter(workerMetricStats -> workerMetricStats.getProperties() != null
+                            && workerMetricStats
+                                    .getProperties()
+                                    .get(FleetSegmentingHandler.VERSION_HASH_KEY)
+                                    .equals(segmentingHandler.getVersionHash())
```
chenylee-aws: this filtering logic should probably stay in the handler i think

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseCountBasedLeaseAssignmentDecider.java
```diff
@@ -56,19 +57,23 @@ public final class LeaseCountBasedLeaseAssignmentDecider implements LeaseAssignm
...
+                || inMemoryStorageView.getAssignableWorkers().isEmpty()) {
             return;
         }
 
         expiredOrUnAssignedLeases.sort(Comparator.comparingLong(Lease::lastCounterIncrementNanos));
 
         final Map<String, Integer> leaseCountPerActiveWorker = getActiveWorkerLeaseCounts();
-        final int target = calculateTargetLeaseCount();
+        final int totalLeases = leaseCountPerActiveWorker.values().stream()
+                        .mapToInt(Integer::intValue)
```
chenylee-aws: Please add comment on why total leases is calculated this way for future reader

amazon-kinesis-client/src/main/java/software/amazon/kinesis/segmenting/FleetSegmentingHandler.java
```diff
@@ -0,0 +1,165 @@
...
+        String lutStr = worker.getProperties().get(VERSION_HASH_LUT_KEY);
+        if (lutStr == null) {
+            return true;
+        }
+        return isVersionHashExpired(Long.parseLong(lutStr));
+    }
+
+    public void setIsVersionEmittedByAllActiveWorkers(
+            final List<WorkerMetricStats> activeWorkerMetrics, final List<WorkerMetricStats> workersOnVersionHash) {
+        isVersionEmittedByAllActiveWorkers = activeWorkerMetrics.containsAll(workersOnVersionHash)
```
chenylee-aws: nit: new HashSet<>(activeWorkerMetrics).equals(new HashSet<>(workersOnVersionHash));

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LeaseCountBasedLeaseAssignmentDecider.java
```diff
@@ -156,9 +170,8 @@ public void balanceWorkerVariance() {
      *
      * @return target lease count per worker, capped by maxLeasesForWorker
      */
-    private int calculateTargetLeaseCount() {
-        final int totalLeases = inMemoryStorageView.getLeaseList().size();
-        final int totalWorkers = inMemoryStorageView.getActiveWorkerIdSet().size();
+    private int calculateTargetLeaseCount(final int totalLeases) {
```
chenylee-aws, line 173: nit: it should take worker from caller as well if we are passing in leases

chenylee-aws (approved):

PR #1744, opened 2026-05-20
Fix FanoutRecordsPublisher restartFrom behavior at SHARD_END

lucienlu-aws:
*Issue #, if available:*

*Description of changes:*
If the FanoutRecordsPublisher reaches the last SubscribeToShardEvent, the `continuationSequenceNumber` returned by the service is `null`. This causes an issue if the publisher runs into an error and the ShardConsumerSubscriber calls `restartFrom` for the lastAccepted event ([ref](https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShardConsumerSubscriber.java#L99)). The FanoutRecordsPublisher would set the continuationSequenceNumber to `null` ([ref](https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java#L162))causing the subscribeToShard requests to continuously fail where the only remediation would be to force the worker to drop the lease.

This commit updates the FanoutRecordsPublisher to handle the SHARD_END event case. If the publisher restartsFrom SHARD_END, it will instead call subscribeToShard with LATEST, which will return the last SHARD_END event from the service and continue the normal SHARD_END tasks through the ShardConsumer and RecordProcessor.

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
@@ -159,9 +159,7 @@ public class FanOutRecordsPublisher implements RecordsPublisher {
                 throw new IllegalArgumentException(
                         "Provided ProcessRecordsInput not created from the FanOutRecordsPublisher");
             }
-            ExtendedSequenceNumber continuationSeqNum =
-                    ((FanoutRecordsRetrieved) recordsRetrieved).continuationSequenceNumber();
-            currentSequenceNumber = continuationSeqNum;
+            currentSequenceNumber = ((FanoutRecordsRetrieved) recordsRetrieved).continuationSequenceNumber();;
         }
     }
 
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
@@ -84,7 +84,7 @@ public class FanOutRecordsPublisher implements RecordsPublisher {
 
     @Getter
     @VisibleForTesting
-    private String currentSequenceNumber;
+    private ExtendedSequenceNumber currentSequenceNumber;
 
     private InitialPositionInStreamExtended initialPositionInStreamExtended;
     private boolean isFirstConnection = true;
@@ -130,7 +130,7 @@ public class FanOutRecordsPublisher implements RecordsPublisher {
                     extendedSequenceNumber,
                     initialPositionInStreamExtended);
             this.initialPositionInStreamExtended = initialPositionInStreamExtended;
-            this.currentSequenceNumber = extendedSequenceNumber.sequenceNumber();
+            this.currentSequenceNumber = extendedSequenceNumber;
             this.isFirstConnection = true;
         }
     }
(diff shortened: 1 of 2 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
```diff
@@ -314,21 +318,31 @@ private boolean hasValidFlow() {
         return flow != null;
     }
 
-    private void subscribeToShard(String sequenceNumber) {
+    private void subscribeToShard(ExtendedSequenceNumber sequenceNumber) {
         synchronized (lockObject) {
             // Clear the delivery queue so that any stale entries from previous subscription are discarded.
             resetRecordsDeliveryStateOnSubscriptionOnInit();
+            log.info("ShardId: {}, consumrArn: {}", shardId, consumerArn);
```
chenylee-aws: might want to remove noisy log

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
```diff
@@ -159,7 +159,9 @@ public void restartFrom(RecordsRetrieved recordsRetrieved) {
                 throw new IllegalArgumentException(
                         "Provided ProcessRecordsInput not created from the FanOutRecordsPublisher");
             }
-            currentSequenceNumber = ((FanoutRecordsRetrieved) recordsRetrieved).continuationSequenceNumber();
+            ExtendedSequenceNumber continuationSeqNum =
+                    ((FanoutRecordsRetrieved) recordsRetrieved).continuationSequenceNumber();
+            currentSequenceNumber = continuationSeqNum;
```
chenylee-aws: nit: just assign to currentSequenceNumber directly

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
```diff
@@ -792,10 +814,10 @@ public void onComplete() {
 
     @Accessors(fluent = true)
     @Data
-    static class FanoutRecordsRetrieved implements RecordsRetrieved {
+    public static class FanoutRecordsRetrieved implements RecordsRetrieved {
```
chenylee-aws: do we need the scope change?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
lucienlu-aws: Was left over from testing I think, thanks

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
lucienlu-aws: Updated

amazon-kinesis-client/src/main/java/software/amazon/kinesis/retrieval/fanout/FanOutRecordsPublisher.java
lucienlu-aws: No, put back to protected scope now

chenylee-aws (approved):

PR #1773, opened 2026-06-11
Introduce single table migration

akidambisrinivasan:
*Issue #, if available:*

*Description of changes:*

Introduce a mechanism for KCL to migrate to single lease table for all entities

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/CoordinatorStateDAO.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/CoordinatorStateDAO.java
@@ -14,13 +14,15 @@
  */
 package software.amazon.kinesis.coordinator;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import javax.annotation.concurrent.ThreadSafe;
 
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDBLockClientOptions.AmazonDynamoDBLockClientOptionsBuilder;
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBLockClient;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient;
@@ -310,17 +312,93 @@ public class CoordinatorStateDAO {
 
     // ==================== Lock Client ====================
 
+    private AmazonDynamoDBLockClient leaseTableLockClient;
+    private AmazonDynamoDBLockClient legacyTableLockClient;
+
     /**
-     * Get the DDB Lock Client options builder pointing to the appropriate table.
-     * Routes based on table migration status:
-     * - COMPLETE: lease table lock
-     * - All others: legacy table lock
+     * Initialize the DDB lock clients for leader election. Creates lock clients for both
+     * the legacy table and the lease table, unless migration is already COMPLETE (in which
+     * case only the lease table lock client is created).
+     *
+     * <p>The lock clients have background heartbeat threads that keep acquired locks alive.
+     * An idle lock client (one that hasn't acquired a lock) will have a running heartbeat
+     * thread but it will be effectively a no-op since there are no locks to heartbeat.</p>
+     *
+     * <p>This method is called once during startup and is not thread-safe with respect to
+     * concurrent calls. It must be called before any calls to {@link #getDDBLockClient()}.</p>
+     *
+     * @param leaseDurationMillis the lock lease duration in milliseconds
+     * @param heartbeatPeriodMillis the heartbeat period in milliseconds
+     * @param workerId the owner name for the lock
      */
-    public AmazonDynamoDBLockClientOptionsBuilder getDDBLockClientOptionsBuilder() {
-        if (isTableMigrationComplete()) {
-            return leaseTableDaoDelegate.getDDBLockClientOptionsBuilder();
+    public void initializeLockClients(
+            final long leaseDurationMillis, final long heartbeatPeriodMillis, final String workerId) {
+        leaseTableLockClient = new AmazonDynamoDBLockClient(leaseTableDaoDelegate
+                .getDDBLockClientOptionsBuilder()
+                .withTimeUnit(TimeUnit.MILLISECONDS)
+                .withLeaseDuration(leaseDurationMillis)
+                .withHeartbeatPeriod(heartbeatPeriodMillis)
+                .withCreateHeartbeatBackgroundThread(true)
+                .withOwnerName(workerId)
+                .build());
+
+        if (!isTableMigrationComplete()) {
+            legacyTableLockClient = new AmazonDynamoDBLockClient(legacyTableDaoDelegate
+                    .getDDBLockClientOptionsBuilder()
+                    .withTimeUnit(TimeUnit.MILLISECONDS)
+                    .withLeaseDuration(leaseDurationMillis)
+                    .withHeartbeatPeriod(heartbeatPeriodMillis)
+                    .withCreateHeartbeatBackgroundThread(true)
+                    .withOwnerName(workerId)
+                    .build());
         }
-        return legacyTableDaoDelegate.getDDBLockClientOptionsBuilder();
+
+        log.info(
+                "Lock clients initialized. LeaseTable lock client: created, Legacy lock client: {}",
+                legacyTableLockClient != null ? "created" : "not created (migration already complete)");
+    }
+
+    /**
+     * Get the active DDB lock client based on the current table migration status.
+     * Routes to the lease table lock client when migration status is COMPLETE,
+     * otherwise routes to the legacy table lock client.
+     *
+     * <p>Thread-safe: the fields are assigned once during {@link #initializeLockClients} (startup)
+     * and never reassigned. The routing decision is based on the {@link TableMigrationStatusProvider}
+     * which is updated atomically. The caller ({@code isLeader()}) is synchronized,
+     * providing happens-before guarantees.</p>
+     *
+     * @return the appropriate {@link AmazonDynamoDBLockClient} for the current migration state
+     */
+    public AmazonDynamoDBLockClient getDDBLockClient() {
+        if (isTableMigrationComplete()) {
+            return leaseTableLockClient;
+        }
+        return legacyTableLockClient;
+    }
+
+    /**
+     * Shutdown both lock clients, closing their background heartbeat threads.
+     * Should be called during scheduler shutdown to ensure no background threads
+     * remain running.
+     */
+    public void shutdownLockClients() {
+        if (legacyTableLockClient != null) {
+            try {
+                log.info("Closing legacy table lock client");
+                legacyTableLockClient.close();
+            } catch (final IOException e) {
+                log.error("Failed to close legacy table lock client", e);
+            }
+        }
+        if (leaseTableLockClient != null) {
+            try {
+                log.info("Closing lease table lock client");
+                leaseTableLockClient.close();
+            } catch (final IOException e) {
+                log.error("Failed to close lease table lock client", e);
+            }
+        }
     }
 
     // ==================== Transactional Operations ====================
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/CoordinatorStateDAO.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/CoordinatorStateDAO.java
@@ -14,513 +14,364 @@
  */
 package software.amazon.kinesis.coordinator;
 
-import java.time.Duration;
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
+import javax.annotation.concurrent.ThreadSafe;
 
-import com.amazonaws.services.dynamodbv2.AmazonDynamoDBLockClientOptions;
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDBLockClientOptions.AmazonDynamoDBLockClientOptionsBuilder;
-import com.google.common.collect.ImmutableMap;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.collections4.MapUtils;
-import software.amazon.awssdk.core.waiters.WaiterResponse;
 import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient;
-import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
-import software.amazon.awssdk.services.dynamodb.model.AttributeAction;
-import software.amazon.awssdk.services.dynamodb.model.AttributeDefinition;
-import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
-import software.amazon.awssdk.services.dynamodb.model.AttributeValueUpdate;
-import software.amazon.awssdk.services.dynamodb.model.BillingMode;
-import software.amazon.awssdk.services.dynamodb.model.ConditionalCheckFailedException;
-import software.amazon.awssdk.services.dynamodb.model.CreateTableRequest;
-import software.amazon.awssdk.services.dynamodb.model.CreateTableResponse;
-import software.amazon.awssdk.services.dynamodb.model.DeleteItemRequest;
-import software.amazon.awssdk.services.dynamodb.model.DescribeTableRequest;
-import software.amazon.awssdk.services.dynamodb.model.DescribeTableResponse;
-import software.amazon.awssdk.services.dynamodb.model.DynamoDbException;
 import software.amazon.awssdk.services.dynamodb.model.ExpectedAttributeValue;
-import software.amazon.awssdk.services.dynamodb.model.GetItemRequest;
-import software.amazon.awssdk.services.dynamodb.model.GetItemResponse;
-import software.amazon.awssdk.services.dynamodb.model.KeySchemaElement;
-import software.amazon.awssdk.services.dynamodb.model.KeyType;
-import software.amazon.awssdk.services.dynamodb.model.ProvisionedThroughput;
-import software.amazon.awssdk.services.dynamodb.model.ProvisionedThroughputExceededException;
-import software.amazon.awssdk.services.dynamodb.model.PutItemRequest;
-import software.amazon.awssdk.services.dynamodb.model.ResourceNotFoundException;
-import software.amazon.awssdk.services.dynamodb.model.ScalarAttributeType;
-import software.amazon.awssdk.services.dynamodb.model.ScanRequest;
-import software.amazon.awssdk.services.dynamodb.model.ScanResponse;
-import software.amazon.awssdk.services.dynamodb.model.TableDescription;
-import software.amazon.awssdk.services.dynamodb.model.TableStatus;
-import software.amazon.awssdk.services.dynamodb.model.UpdateItemRequest;
-import software.amazon.awssdk.services.dynamodb.waiters.DynamoDbAsyncWaiter;
-import software.amazon.awssdk.utils.CollectionUtils;
+import software.amazon.awssdk.services.dynamodb.model.TransactWriteItem;
+import software.amazon.awssdk.services.dynamodb.model.TransactWriteItemsRequest;
+import software.amazon.awssdk.services.dynamodb.model.TransactionCanceledException;
 import software.amazon.kinesis.annotations.KinesisClientInternalApi;
-import software.amazon.kinesis.common.FutureUtils;
 import software.amazon.kinesis.coordinator.CoordinatorConfig.CoordinatorStateTableConfig;
-import software.amazon.kinesis.coordinator.migration.MigrationState;
-import software.amazon.kinesis.coordinator.streamInfo.StreamInfo;
-import software.amazon.kinesis.leases.DynamoUtils;
+import software.amazon.kinesis.coordinator.delegate.CoordinatorStateDAODelegate;
+import software.amazon.kinesis.coordinator.delegate.LeaseTableCoordinatorStateDAODelegate;
+import software.amazon.kinesis.coordinator.delegate.LegacyTableCoordinatorStateDAODelegate;
+import software.amazon.kinesis.coordinator.migration.TableMigrationStateMachine;
+import software.amazon.kinesis.coordinator.migration.TableMigrationStatus;
+import software.amazon.kinesis.coordinator.migration.TableMigrationStatusProvider;
+import software.amazon.kinesis.leases.EntityType;
 import software.amazon.kinesis.leases.exceptions.DependencyException;
 import software.amazon.kinesis.leases.exceptions.InvalidStateException;
 import software.amazon.kinesis.leases.exceptions.ProvisionedThroughputException;
-import software.amazon.kinesis.utils.DdbUtil;
-
-import static java.util.Objects.nonNull;
-import static software.amazon.kinesis.common.FutureUtils.unwrappingFuture;
-import static software.amazon.kinesis.coordinator.CoordinatorState.COORDINATOR_STATE_TABLE_HASH_KEY_ATTRIBUTE_NAME;
 
 /**
- * Data Access Object to abstract accessing {@link CoordinatorState} from
- * the CoordinatorState DDB table.
+ * Data Access Object that routes {@link CoordinatorState} operations to the appropriate
+ * DDB table based on the current {@link TableMigrationStatus}.
+ *
+ * <h2>Read Logic</h2>
+ * <ul>
+ *   <li>If status != COMPLETE: read from legacy table first; if found, return it. Otherwise read from lease table.
+ *       The legacy delegate handles the case where the table doesn't exist (returns null/empty).</li>
+ *   <li>If status == COMPLETE: read only from lease table.</li>
+ * </ul>
+ *
+ * <h2>Write Logic</h2>
+ * <ul>
+ *   <li>cache table migration startup upon initialize(). </li>
+ *   <li>If cachedStatus == COMPLETE or PENDING: write to lease table only.</li>
+ *   <li>If cachedStatus == INIT or DEPLOYED: write to legacy table only.</li>
+ * </ul>
+ *
+ * The difference between read and write are:
+ * 1. Reads are merged from both tables while writes go only to one table. This is
+ *    because when status is INIT/DEPLOYED, the application could be on a rollback
+ *    + rollforward path, so even if application is currently writing to
+ *    Legacy table, in the past workers may have written to Lease table. So its safer
+ *    to read both until the final transition to COMPLETED.
+ * 2. Reads dynamically flip on completed state, but wokers flip to Lease table
+ *    when it reboots. This is to ensure the worker writes move as the deployment
+ *    progresses in the cluster fleet.
+ *
+ * <h2>Special Cases</h2>
+ * <ul>
+ *   <li>The table migration state machine directly uses the {@link #leaseTableDaoDelegate} field
+ *       for its reads/writes and sets the {@link TableMigrationStatusProvider} before calling initialize().</li>
+ *   <li>All operations require {@link #initialize()} to have been called first.</li>
+ * </ul>
+ *
+ * <h2>Lock Client</h2>
+ * The lock client options builder routes based on status:
+ * <ul>
+ *   <li>COMPLETE: lease table lock</li>
+ *   <li>All others (INIT, DEPLOYED, PENDING, UNKNOWN): legacy table lock</li>
+ * </ul>
+ * During transition from PENDING to COMPLETE, its possible workers that are caught up to COMPLETE
+ * state will grab lock from lease table however those that grab the lock when their local state
+ * is still PENDING will grab from the legacy lock. However TableMigrationStateMachine guarantees
+ * that it performs a post leadership acquire check where if the local state is different from the
+ * DDB state it will force the lock to be relinquished and then return false to is caller on
+ * isLeader method. The next isLeader call will use the COMPLETED status to grab the lock correctly.
+ * Given that you can never rollback from COMPLETED means that we dont have a rollback scenario
+ * to handle correct lock acquision that guarantees only a single leader at a time. Refer to
+ * {@link TableMigrationStateMachine.handleLeaderLockResult}
+ *
+ * <h2>TODO</h2>
+ * Update and delete operations currently route based on the table migration status which may not
+ * match where the entry was originally read from. If a caller reads from legacy but the status
+ * transitions to PENDING/COMPLETE before the update, the update targets the lease table where
+ * the entry may not yet exist. Consider tracking read-source per entry and routing to the
+ * appropriate delegate based on the source. For now, the routing is best-effort and
+ * callers that perform read-then-update must handle conditional check failures gracefully.
+ * There is no usecase for an update coordinator state except for the migration state and
+ * table migration state. Table Migration state is always updated from state machine using
+ * appropriate delegate class directly and never through this DOA.
+ * For more details refer to TableMigrationStateMachine.
+ * And for migration state the assumption is both state machines are not running concurrently.
  */
 @Slf4j
 @KinesisClientInternalApi
+@ThreadSafe
 public class CoordinatorStateDAO {
+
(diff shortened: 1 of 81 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationState.java
```diff
@@ -0,0 +1,265 @@
...
+
+        try {
+            final HashMap<String, AttributeValue> mutableAttributes = new HashMap<>(attributes);
+            final TableMigrationStatus tableMigrationStatus = TableMigrationStatus.valueOf(mutableAttributes
+                    .remove(TABLE_MIGRATION_STATUS_ATTRIBUTE_NAME)
+                    .s());
+            final String modifiedBy =
+                    mutableAttributes.remove(MODIFIED_BY_ATTRIBUTE_NAME).s();
+            final long modifiedTimestamp = Long.parseLong(
+                    mutableAttributes.remove(MODIFIED_TIMESTAMP_ATTRIBUTE_NAME).n());
```
lucienlu-aws, line 129: Why make a copy and call remove instead of accessing original HashMap and calling get()?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationState.java
```diff
@@ -0,0 +1,265 @@
...
+                    TABLE_MIGRATION_HASH_KEY,
+                    tableMigrationStatus,
+                    modifiedBy,
+                    modifiedTimestamp,
+                    historyList,
+                    mutableAttributes);
+
+            if (!mutableAttributes.isEmpty()) {
+                log.info("Unknown attributes {} for state {}", mutableAttributes, tableMigrationState);
+            }
```
lucienlu-aws, line 147: We copy the passed in HashMap and remove the necessary elements and then construct the TableMigrationState object? Are extra elements not expected? What are the situations where there can be extra attributes here? There is a log when there are extra attributes too.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationState.java
```diff
@@ -0,0 +1,265 @@
...
+    public TableMigrationState update(final TableMigrationStatus tableMigrationStatus, final String modifiedBy) {
+        log.info(
+                "Table Migration state is being updated to {} current state {} caller {}",
+                tableMigrationStatus,
+                this,
+                StackTraceUtils.getPrintableStackTrace(Thread.currentThread().getStackTrace()));
+        addTableMigrationStateHistoryEntry(this.tableMigrationStatus, this.modifiedBy, this.modifiedTimestamp);
+        this.tableMigrationStatus = tableMigrationStatus;
+        this.modifiedBy = modifiedBy;
+        this.modifiedTimestamp = System.currentTimeMillis();
```
lucienlu-aws, line 187: Do any of our tests rely on this modifiedTimestamp? i.e. comparing time since this timestamp. If so, we should pass a Clock instance to the constructor to allow mocking in unit tests to avoid flaky tests

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationState.java
```diff
@@ -0,0 +1,265 @@
...
+        return this;
+    }
+
+    public void addTableMigrationStateHistoryEntry(
+            final TableMigrationStatus lastTableMigrationStatus,
+            final String lastModifiedBy,
+            final long lastModifiedTimestamp) {
+        history.add(
+                0,
+                new TableMigrationStateHistoryEntry(lastTableMigrationStatus, lastModifiedBy, lastModifiedTimestamp));
```
lucienlu-aws, line 197: Not sure where this history is used yet and if we need to view more than the most recent element, but it looks like we are treating it like a stack. Should we just use Java stack data type instead?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/EntityType.java
```diff
@@ -0,0 +1,137 @@
...
+        private final EntityType entityType;
+
+        CoordinatorStateType(final EntityType entityType) {
+            this.entityType = entityType;
+        }
+
+        /**
+         * Get the parent {@link EntityType} for this coordinator state type.
+         */
+        public EntityType getEntityType() {
```
lucienlu-aws, line 85: lombok `@Getter`?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/EntityType.java
```diff
@@ -0,0 +1,137 @@
...
+                }
+            }
+            return null;
+        }
+    }
+
+    /**
+     * The string value persisted in DynamoDB as the entityType attribute.
+     */
+    public String getDdbValue() {
```
lucienlu-aws, line 117: lombok `@Getter`?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/segmenting/FleetSegmentingHandler.java
```diff
@@ -188,7 +190,8 @@ private void updateLeaderVersionHashLut() {
     private Map<String, ExpectedAttributeValue> getVersionHashHeartbeatExpectationMap(final String leaderKey) {
         final Map<String, ExpectedAttributeValue> expectations = new HashMap<>();
         expectations.put(
-                CoordinatorState.COORDINATOR_STATE_TABLE_HASH_KEY_ATTRIBUTE_NAME,
+                // TODO: fix this
```
lucienlu-aws: What is TODO here for? Add details to comment

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationState.java
```diff
@@ -0,0 +1,265 @@
...
+        this(
+                TABLE_MIGRATION_HASH_KEY,
+                TableMigrationStatus.TABLE_MIGRATION_STATUS_INIT,
+                modifiedBy,
+                System.currentTimeMillis(),
+                new ArrayList<>(),
+                new HashMap<>());
+    }
+
+    public HashMap<String, AttributeValue> serialize() {
```
nikhil-m-amazon, line 97: `serialize()` and `getDynamoUpdate()` (line 203) build a fresh `HashMap` and never call `super`. The base `CoordinatorState.serialize()` (CoordinatorState.java:75) is what writes the `entityType` attribute, so TABLE_MIGRATION items get persisted without it. `MigrationState` does this right at line 93.

Single-item reads still work because `resolveEntityType()` falls back to key inference, which hides the problem. But `listCoordinatorStateByEntityType()` filters on `entityType` server-side, so it comes back empty for TABLE_MIGRATION. Can we seed both maps from `super`, i.e. `new HashMap<>(super.serialize())` and merge in `super.getDynamoUpdate()`?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leader/LeaderLock.java
```diff
@@ -0,0 +1,29 @@
+package software.amazon.kinesis.leader;
```
nikhil-m-amazon, line 15: Nit (can be ignored): Missing the Apache license header that the other new files have. CI's license check will probably fail on it.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leases/dynamodb/DynamoDBLeaseSerializer.java
```diff
@@ -130,40 +138,51 @@ public Lease fromDynamoRecord(final Map<String, AttributeValue> dynamoRecord) {
...
-        if (DynamoUtils.safeGetDouble(dynamoRecord, THROUGHPUT_KBPS) != null) {
-            leaseToUpdate.throughputKBps(DynamoUtils.safeGetDouble(dynamoRecord, THROUGHPUT_KBPS));
-        }
-
-        if (DynamoUtils.safeGetString(dynamoRecord, CHECKPOINT_OWNER) != null) {
-            leaseToUpdate.checkpointOwner(DynamoUtils.safeGetString(dynamoRecord, CHECKPOINT_OWNER));
-        }
-
-        return leaseToUpdate;
+        log.warn("dynamoRecord is not a Lease entity {}", dynamoRecord);
```
nikhil-m-amazon: Nit (can be ignored): This warns once per non-lease item on every scan and dumps the whole record. In single-table mode that's the normal steady state, so it'll flood the logs. Should be `debug`, and probably drop the full-record interpolation.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/CoordinatorStateDAO.java
```diff
@@ -14,513 +14,364 @@
...
+     * @param transactWriteItems the list of transact write items (max 100 per DDB limit)
+     * @throws DependencyException if DDB fails unexpectedly or the transaction is cancelled
+     */
+    public void executeTransactWrite(final List<TransactWriteItem> transactWriteItems) throws DependencyException {
+        final TransactWriteItemsRequest request = TransactWriteItemsRequest.builder()
+                .transactItems(transactWriteItems)
+                .build();
+        try {
+            dynamoDbAsyncClient.transactWriteItems(request).get();
+        } catch (final Exception e) {
```
nikhil-m-amazon: This catches generic `Exception` and calls `.get()` directly, so an `InterruptedException` gets wrapped into `DependencyException` and the interrupt flag is lost. Can we split out the `InterruptedException` case and call `Thread.currentThread().interrupt()`?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/delegate/LeaseTableCoordinatorStateDAODelegate.java
```diff
@@ -0,0 +1,16 @@
+package software.amazon.kinesis.coordinator.delegate;
+
+import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient;
+import software.amazon.kinesis.leases.dynamodb.DynamoDBLeaseSerializer;
+
+public class LeaseTableCoordinatorStateDAODelegate extends CoordinatorStateDAODelegate {
```
nikhil-m-amazon, line 22: Nit (can be ignored): Missing `@KinesisClientInternalApi`. The sibling delegates and the base class all have it.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/MigrationAwareLAMDataManager.java
```diff
@@ -0,0 +1,419 @@
...
+        // Step 1: Start lease table scan async on the executor service.
+        final CompletableFuture<Map<EntityType, EntityScanList>> leaseTableFuture = CompletableFuture.supplyAsync(
+                () -> loadWithRetry(() -> entityDAO.scanEntities(EntityType.LEASE, EntityType.WORKER_METRIC_STATS)),
+                executorService);
+
+        // Step 2: Load legacy worker metrics synchronously on the calling thread (small table).
+        final List<WorkerMetricStats> legacyTableWorkerMetrics;
+        if (shouldScanLegacyTable(status)) {
+            legacyTableWorkerMetrics =
+                    loadWithRetry(workerMetricsDAO.getLegacyTableDaoDelegate()::getAllWorkerMetricStats);
```
nikhil-m-amazon, line 133: **C3-DES-1** If the legacy worker-metrics scan fails, `loadData` throws and the whole assignment cycle aborts, even though the lease-table scan (the part assignment actually needs) succeeded. So a transient blip on the legacy table stalls all lease assignment for the entire migration window. Could we degrade instead: treat legacy metrics as empty for that cycle, skip publishing the summary, and let assignment proceed?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/MigrationAwareLAMDataManager.java
```diff
@@ -0,0 +1,419 @@
...
+        int retryAttempt = 0;
+        while (true) {
+            try {
+                return loadFunction.call();
+            } catch (final Exception e) {
+                if (retryAttempt < DDB_LOAD_RETRY_ATTEMPT) {
+                    log.warn("Failed to load, retrying (attempt {})", retryAttempt + 1, e);
+                    retryAttempt++;
+                } else {
+                    throw new CompletionException(e);
```
nikhil-m-amazon, line 272: **C3-DES-3** `loadWithRetry` catches generic `Exception` and rethrows as `CompletionException`. That hides `ProvisionedThroughputException`, so the back-off keyed on it never fires, and it swallows `InterruptedException` without restoring the flag. The typed `throws` on `loadData` then becomes unreachable. `LeaseAssignmentManager.loadInMemoryStorageView` has the same pattern. Suggest unwrapping the real cause and restoring the interrupt flag.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/MigrationAwareLAMDataManager.java
```diff
@@ -0,0 +1,419 @@
...
+            final LeaseManagementConfig.WorkerUtilizationAwareAssignmentConfig config) {
+        this.entityDAO = entityDAO;
+        this.workerMetricsDAO = workerMetricsDAO;
+        this.tableMigrationStatusProvider = tableMigrationStatusProvider;
+        this.migrationSummaryConsumer = migrationSummaryConsumer;
+        this.workerMetricsExpiryDuration = Duration.ofMillis(
+                DEFAULT_NO_OF_SKIP_STAT_FOR_DEAD_WORKER_THRESHOLD * config.workerMetricsReporterFreqInMillis());
+        this.supportCodeExpiryThreshold = Duration.ofMillis(config.workerMetricsReporterFreqInMillis() * 3);
+        this.staleWorkerMetricsCleanupDuration = config.staleWorkerMetricsEntryCleanupDuration();
+        this.executorService = Executors.newFixedThreadPool(
```
nikhil-m-amazon, line 115: **C3-DES-5** The executor is created in the constructor and never shut down. Each instance allocates its own pool, so if the Scheduler gets recreated the pools leak. The old LAM used static pools to avoid exactly this. Could we make it static, or add a `shutdown()` that LAM's stop path calls?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/MigrationAwareLAMDataManager.java
```diff
@@ -0,0 +1,419 @@
...
+
+        return LAMDataSnapshot.builder()
+                .leases(leases)
+                .leaseDeserializationFailures(leaseDeserializationFailures)
+                .workerMetricStats(activeWorkerMetrics)
+                .build();
+    }
+
+    /**
+     * Deletes stale worker metrics entries asynchronously on the common ForkJoinPool.
```
nikhil-m-amazon: Nit (can be ignored): Comment says "common ForkJoinPool" but the deletion actually runs on `executorService`. Stale comment.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/LAMDataSnapshot.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.List;
+
+import lombok.Builder;
+import lombok.Value;
+import software.amazon.kinesis.annotations.KinesisClientInternalApi;
+import software.amazon.kinesis.leases.Lease;
+import software.amazon.kinesis.worker.metricstats.WorkerMetricStats;
+
+/**
+ * Immutable snapshot of the data loaded by {@link LAMDataProvider} for a single
```
nikhil-m-amazon: Nit (can be ignored): Javadoc links `{@link LAMDataProvider}`, which doesn't exist. The interface is `LAMDataManager`.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
```diff
@@ -432,10 +454,17 @@ private DynamicMigrationComponentsInitializer createDynamicMigrationComponentsIn
                         .workerUtilizationAwareAssignmentConfig()
                         .inMemoryWorkerMetricsCaptureFrequencyMillis());
 
+        final EntityDAO entityDAO = new DynamoDBLeaseTableDao(
+                leaseManagementConfig.dynamoDBClient(),
+                leaseManagementConfig.tableName(),
+                leaseSerializer,
+                coordinatorStateDAO.getLeaseTableDaoDelegate(),
+                Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors()),
```
nikhil-m-amazon: There's no shutdown for the table migration state machine. The async copy runs on the common pool and only gets cancelled on leadership loss or rollback. `Scheduler.shutdown()` never stops it, so a leader shutting down mid-copy keeps mutating both DDB tables after the worker thinks it's done. The fixed thread pool created here for the migration EntityDAO is never shut down either.

Can we add an idempotent `shutdown()` to the `TableMigrationStateMachine` interface that cancels the in-flight move and shuts the pool, and call it alongside `migrationStateMachine.shutdown()`? A dedicated executor would also be better than the common pool here.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DynamicMigrationComponentsInitializer.java
```diff
@@ -155,58 +152,22 @@ public final class DynamicMigrationComponentsInitializer {
         this.segmentingHandler = segmentingHandler;
     }
 
-    public void initialize(final ClientVersion migrationStateMachineStartingClientVersion) throws DependencyException {
+    /**
+     * Phase 1 initialization (CLIENT_VERSION_INIT): Purely passive 2.x compatible mode.
+     * No WorkerMetrics collection/reporting, no GSI, no LAM.
+     * Uses deterministic leader decider and lease-count-based assignment.
+     */
+    public synchronized void initializeClientVersionForPhase1() {
```
nikhil-m-amazon, line 212: Phase 1 sets `initialized = true` but never assigns `leaseAssignmentManager`. The other four init paths do. `shutdown()` gates on `initialized` and then calls `leaseAssignmentManager.stop()` unconditionally, so it NPEs. And per the InitState javadoc, Phase 1 is where workers sit until Phase 2 gets deployed, so hitting shutdown from Phase 1 is the normal path, not an edge case. Can we null-guard the LAM and metrics teardown?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DynamicMigrationComponentsInitializer.java
```diff
@@ -307,6 +268,27 @@ public synchronized void initializeClientVersionForUpgradeFrom2x(final ClientVer
             throws DependencyException {
         log.info("Initializing KCL components for upgrade from 2x from {}", fromClientVersion);
 
+        if (fromClientVersion == ClientVersion.CLIENT_VERSION_INIT) {
```
nikhil-m-amazon, line 322: The startup block is copy-pasted across all four init methods, and each one sets `initialized=true` on its own. That's exactly how the Phase 1 NPE slipped in. Pulling this into a shared `initializeStartupComponents(dualMode, assignmentMode, leaderDeciderSupplier)` would kill the duplication and the class of bug that comes with it.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationState.java
akidambisrinivasan: Because CoordiantorState keeps all unrecognized attributes so we need to remove the once that are used to construct the child class object

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationState.java
akidambisrinivasan: Yes this was the bug Lucien also found in testing, i fixed it to call super.serialize to include the entityType now

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/MigrationAwareLAMDataManager.java
akidambisrinivasan: assignment also needs worker metrics, this is how even the previous lam worked I think. If it encounters errors reading either it retries and after some time it will give up leadership but Ill confirm. Because if we dont scan legacy worker metrics we may do wrong assignment like moving leases away from those worker which we dont want to happen.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DynamicMigrationComponentsInitializer.java
akidambisrinivasan: Ack, will resolve this.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/MigrationAwareLAMDataManager.java
akidambisrinivasan: ya this was mainly moving existing code from LAM to a different class, and the scheduler originally in LAM was never shutdown either. Ill fix this

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/Scheduler.java
akidambisrinivasan: Yes it should be there I'll fix this

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/assignment/MigrationAwareLAMDataManager.java
akidambisrinivasan: Yes this makes sense, fixed it

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationState.java
akidambisrinivasan: yes this is what is used for the bake time, I can pass clock to TableMigrationStateMachine.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationState.java
akidambisrinivasan: If there are new attributes in DDB that the schema code in the worker is not aware of it will include it or if a state is deserialized incorrectly its helpful to see what other attributes were present. Mainly for debugging.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationState.java
akidambisrinivasan: Its used for debugging, its helpful to know which worker did the transition and for the why we can then look at logs from that worker. Its the same thing for 2 -> 3 migration state. DDB doesnt support stack, list is the way to do it. It could be a locally but serialized into list, but since its size limited its easy to remove from bottom of list than from stack. Its stores upto 10 history entries which should be good for debugging I think.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/segmenting/FleetSegmentingHandler.java
akidambisrinivasan: table migration does not handle fleet segmenting handler, I will work on address this today

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DynamicMigrationComponentsInitializer.java
akidambisrinivasan: metrics is always created so it can be shutdown, will null guard lam.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/worker/metricstats/WorkerMetricStatsDAO.java
```diff
@@ -12,197 +12,177 @@
...
+     * @throws DependencyException if DDB fails unexpectedly
+     * @throws InvalidStateException if a required table does not exist
+     * @throws ProvisionedThroughputException if DDB lacks capacity
+     */
+    public List<WorkerMetricStats> getAllWorkerMetricStats()
+            throws DependencyException, InvalidStateException, ProvisionedThroughputException {
+        ensureInitialized();
 
-        Preconditions.checkArgument(nonNull(workerMetrics.getLastUpdateTime()), "LastUpdateTime field not set");
+        if (cachedStatus == TableMigrationStatus.TABLE_MIGRATION_STATUS_COMPLETE) {
```
akidambisrinivasan: This should not be cached Status that is only used for writing decisions

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationStateMachineImpl.java
```diff
@@ -0,0 +1,1047 @@
...
+        if (stateFromDDB == null) {
+            // First time min support code is met — write INIT to legacy DDB.
+            // The modifiedTimestamp on this write serves as the bake start time.
+            log.info("INIT: min support code met, writing INIT to legacy DDB");
+            // If we encounter error while writing to DDB, no need to fail isLeader call
+            // we can retry writing next time.
+            writeStatusToLegacySafe(TableMigrationStatus.TABLE_MIGRATION_STATUS_INIT, null);
+            return;
+        }
+
```
akidambisrinivasan, line 464: should we check if ddb state.status == INIT before using last modified time for baking ?

akidambisrinivasan: if ddb state is DEPLOYED we need to transition status provider to DEPLOYED (This could be the case where we attempted a transition to DEPLOYED but perceived a failure and did not update provider from line 452

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationStateMachineImpl.java
```diff
@@ -0,0 +1,1047 @@
...
+     */
+    private TableMigrationStatus determineAndPersistStatus(final TableMigrationState stateFromDDB)
+            throws DependencyException, InvalidStateException {
+
+        // Case 1: No state in DDB.
+        if (stateFromDDB == null) {
+            if (!legacyDao.isEnabled()) {
+                // No legacy table — 2→3 migration short-circuit: write COMPLETE to lease table directly.
+                log.info("No state in DDB, no legacy table -> writing COMPLETE (2->3 migration)");
+                writeStatusToLeaseTable(TableMigrationStatus.TABLE_MIGRATION_STATUS_COMPLETE, null);
```
akidambisrinivasan: note: this was removed in later commits so please ignore it

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leader/DynamoDBLockBasedLeaderDecider.java
```diff
@@ -151,6 +167,14 @@ public synchronized Boolean isLeader(final String workerId) {
             response = lockItem.get().getOwnerName().equals(workerId);
         }
 
+        try {
+            tableMigrationStateMachine.handleLeaderLockResult(response);
+        } catch (final DependencyException | InvalidStateException e) {
```
akidambisrinivasan, line 170: Does it even throw DependencyException and when and why should we release lock?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leader/DynamoDBLockBasedLeaderDecider.java
```diff
@@ -151,6 +167,14 @@ public synchronized Boolean isLeader(final String workerId) {
...
+            log.warn("handleLeaderLockResult failed, releasing lock and returning false", e);
+            releaseLeadershipIfHeld();
+            response = false;
```
akidambisrinivasan, line 173: should we reset lastIsLeaderResult to false ?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leader/DynamoDBLockBasedLeaderDecider.java
akidambisrinivasan: yes if the TableMigrationState from DDB could not be read it throws dependency exception because we it could be COMPLETE and if it is we should not be holding the Legacy table lock.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leader/DynamoDBLockBasedLeaderDecider.java
akidambisrinivasan: this is happening in line 179 already

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationState.java
```diff
@@ -0,0 +1,265 @@
...
+@ToString(callSuper = true)
+@Slf4j
+@KinesisClientInternalApi
+public class TableMigrationState extends CoordinatorState {
+    /**
+     * Key value for the item in the CoordinatorState table
+     */
+    public static final String TABLE_MIGRATION_HASH_KEY = "TableMigration3.5";
+    /**
+     * Attribute name in migration state item, whose value is used during
```
akidambisrinivasan, line 51: Comment needs to be updated

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationState.java
```diff
@@ -0,0 +1,265 @@
...
+
+    public static final String MODIFIED_BY_ATTRIBUTE_NAME = "mb";
+    public static final String MODIFIED_TIMESTAMP_ATTRIBUTE_NAME = "mts";
+    public static final String HISTORY_ATTRIBUTE_NAME = "h";
+    private static final int MAX_HISTORY_ENTRIES = 10;
+
+    /**
+     * Default bake time in seconds (60 seconds).
+     */
+    public static final long DEFAULT_BAKE_TIME_SECONDS = 60L;
```
akidambisrinivasan, line 66: I think default should be atleast 60 mins not seconds

amazon-kinesis-client/src/main/java/software/amazon/kinesis/leader/DynamoDBLockBasedLeaderDecider.java
```diff
@@ -63,7 +67,8 @@ static DynamoDBLockBasedLeaderDecider create(
             final String workerId,
             final Long leaseDuration,
             final Long heartbeatPeriod,
-            final MetricsFactory metricsFactory) {
+            final MetricsFactory metricsFactory,
+            final TableMigrationStateMachine tableMigrationStateMachine) {
         final AmazonDynamoDBLockClient dynamoDBLockClient = new AmazonDynamoDBLockClient(coordinatorStateDao
                 .getDDBLockClientOptionsBuilder()
```
chenylee-aws: when the current worker moves to migration status to COMPLETE, which component will flip the lockClient to use the leader in the lease table?

akidambisrinivasan: Good catch, the dynamodb lock client should change in leader decider, Ill fix it.

chenylee-aws (approved):

PR #1779, opened 2026-06-20
Add Metrics for table migration

akidambisrinivasan:
Metrics Summary:

Leader emitted always:
  TableMigration:
    - StatusOrdinal (NONE): Ordinal of DDB status (0=UNKNOWN, 1=INIT, 2=DEPLOYED, 3=PENDING, 4=COMPLETE) WorkerMetrics:
    - FleetMinSupportCode (NONE): Minimum support code across all lease-owning workers

Leader emitted during migration:
  TableMigration:
    - Phase1Worker (COUNT): Workers with SINGLE_TABLE_MIGRATION support code minus Phase2Workers
    - Phase2Worker (COUNT): Workers writing WorkerMetricStats to lease table
    - PrePhase1Worker (COUNT): Workers on older versions without migration support code
    - WriteFault (COUNT): 1 on DDB write failure, 0 on success
    - DeleteFault (COUNT): 1 on DDB write failure, 0 on success
    - CompletionFault (COUNT): 1 when transactional write of COMPLETE fails, 0 on success TableMigrationAsyncMove:
    - Success (COUNT): 1 on successful move of coordinator state entries, 0 on failure
    - Time (MILLISECONDS): Duration of the async move operation
    - BatchCount (COUNT): Number of successfully moved batches

All workers emitted during migration:
  TableMigration:
    - ReadFault (COUNT): 1 on DDB read failure in readStateFromLegacy, 0 on success
    - Time (MILLISECONDS): Duration of state machine run TableMigrationInitialize:
    - Success (COUNT): 1 on successful initialize, 0 on failure
    - Time (MILLISECONDS): Duration of initialize

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationStateMachineImpl.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationStateMachineImpl.java
@@ -153,6 +153,7 @@ public class TableMigrationStateMachineImpl implements TableMigrationStateMachin
     private static final String PRE_PHASE1_WORKER_METRIC = "PrePhase1Worker";
     private static final String READ_FAULT_METRIC = "ReadFault";
     private static final String WRITE_FAULT_METRIC = "WriteFault";
+    private static final String DELETE_FAULT_METRIC = "DeleteFault";
     private static final String COMPLETION_FAULT_METRIC = "CompletionFault";
     private static final String MOVE_BATCH_COUNT_METRIC = "BatchCount";
 
@@ -305,7 +306,6 @@ public class TableMigrationStateMachineImpl implements TableMigrationStateMachin
             return;
         }
         final long startTime = System.currentTimeMillis();
-        boolean success = false;
 
         try {
             // Refresh from legacy DDB and re-apply config override to determine effective local status.
@@ -346,7 +346,6 @@ public class TableMigrationStateMachineImpl implements TableMigrationStateMachin
                     pendingMoveFuture.cancel(true);
                     pendingMoveFuture = null;
                 }
-                success = true;
                 return;
             }
 
@@ -368,10 +367,8 @@ public class TableMigrationStateMachineImpl implements TableMigrationStateMachin
                 default:
                     break;
             }
-
-            success = true;
         } finally {
-            MetricsUtil.addSuccessAndLatency(scope, success, startTime, MetricsLevel.SUMMARY);
+            MetricsUtil.addLatency(scope, null, startTime, MetricsLevel.SUMMARY);
             MetricsUtil.endScope(scope);
         }
     }
@@ -1164,7 +1161,7 @@ public class TableMigrationStateMachineImpl implements TableMigrationStateMachin
                     state,
                     e);
         } finally {
-            scope.addData(WRITE_FAULT_METRIC, success ? 0 : 1, StandardUnit.COUNT, MetricsLevel.SUMMARY);
+            scope.addData(DELETE_FAULT_METRIC, success ? 0 : 1, StandardUnit.COUNT, MetricsLevel.SUMMARY);
         }
     }
 
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationStateMachineImpl.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationStateMachineImpl.java
@@ -24,6 +24,7 @@ import java.util.concurrent.TimeUnit;
 
 import lombok.extern.slf4j.Slf4j;
 import software.amazon.awssdk.annotations.ThreadSafe;
+import software.amazon.awssdk.services.cloudwatch.model.StandardUnit;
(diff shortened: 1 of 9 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationStateMachineImpl.java
```diff
@@ -1082,12 +1204,15 @@ private void writeStatusToLegacy(final TableMigrationStatus status, final TableM
...
                     log.info("Updated table migration state in legacy: {}", status);
+                    success = true;
                 } else {
                     log.warn("Conditional write to legacy failed for {}. Another worker advanced state.", status);
                 }
             }
         } catch (final InvalidStateException | ProvisionedThroughputException e) {
             throw new DependencyException("Failed to write table migration status " + status + " to legacy", e);
+        } finally {
+            scope.addData(WRITE_FAULT_METRIC, success ? 0 : 1, StandardUnit.COUNT, MetricsLevel.SUMMARY);
```
nikhil-m-amazon, line 1212: This emits `WriteFault=1` whenever `success` is false, but `success` stays false on two benign branches: the lost conditional write at line 1209 ("Another worker advanced state") and the "already exists in legacy, another worker created it" case at 1199. So `WriteFault` will spike during normal operation, and the doc defines it as "1 on DDB write failure." Can we set `success=true` on those branches so `WriteFault` only counts real DDB dependency failures?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationStateMachineImpl.java
```diff
@@ -1021,38 +1132,46 @@ private synchronized boolean isLegacyWorkerMetricsEmpty() {
...
             legacyDao.deleteCoordinatorState(state.getKey());
             log.info("Deleted table migration state from legacy (best effort reset). Previous state: {}", state);
+            success = true;
         } catch (final ProvisionedThroughputException | InvalidStateException | DependencyException e) {
             log.warn(
                     "Failed to delete table migration state from legacy (best effort), will retry next cycle. State: {}",
                     state,
                     e);
+        } finally {
+            scope.addData(WRITE_FAULT_METRIC, success ? 0 : 1, StandardUnit.COUNT, MetricsLevel.SUMMARY);
```
nikhil-m-amazon: `deleteStateSafe` emits `WriteFault` here, and `writeStatusToLegacy` emits `WriteFault` too (line 1215). A single `handleLeaderLockResult` cycle can do both (the Phase 1 rollback reset deletes and then writes), so two `WriteFault` datapoints land on the same scope. Could the delete path use its own metric name, e.g. `DeleteFault`, so each fault points at a specific op?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationStateMachineImpl.java
```diff
@@ -224,91 +246,160 @@ public synchronized void initialize() throws DependencyException, InvalidStateEx
...
+                    handlePendingState(stateFromDDB, scope);
+                    break;
+                default:
+                    break;
             }
-            return;
+
+            success = true;
+        } finally {
+            MetricsUtil.addSuccessAndLatency(scope, success, startTime, MetricsLevel.SUMMARY);
```
nikhil-m-amazon: When another leader completes the migration we throw `InvalidStateException` at line 324 as a deliberate signal to release the lock. That throw happens before `success = true`, so this `finally` records `Success=0` for the `TableMigration` operation on what's actually the normal completion handoff. The `DependencyException` we propagate on a transient DDB read blip hits the same path. So `Success` dips to 0 right when a cutover succeeds and on ordinary throttles. Is that intended?

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationStateMachineImpl.java
akidambisrinivasan: Those two are still write failures, the worker decided to create state and failed, and worker decided to update state and failed, I wanted to capture those race, eventually it should resolve, but its useful to see those race especially when it leads to inconsistent state due to a bug and doesn't resolve.  Let me know what you think.

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/TableMigrationStateMachineImpl.java
akidambisrinivasan: You are right, the method itself doesnt or shouldnt fail the only exceptions are intended to be thrown to let the caller release the lock. Perhaps returning a boolean would have been a better way to achieve that. My main intention of the metric here is to capture the latency of the method to make sure we can monitor that it doesnt take too long since it happens during isLeader checks which should be fast. Ill just keep latency and remove the success since individual failures like ddb read/write failures etc are already covered by their own metrics.

nikhil-m-amazon:
LGTM after the latest commit, TY.

chenylee-aws (approved):

PR #1799, opened 2026-07-28
Addressed memory leaks and improved overall memory usage

pelaezryan:
1. Fixed a memory leak in PeriodicShardSyncManager where hashRangeHoleTrackerMap retained entries for deleted streams by adding cleanup via retainAll after each sync run.
2. Reduced memory overhead in LeaseAssignmentManager by replacing full Lease object storage in prevRunLeasesState with a lightweight LeaseCounterSnapshot that only retains the fields actually needed
3. streamToShardSyncTaskManagerMap in the Scheduler is a class level map that is always added to but never cleaned up. The only time this map was getting "flushed" was during a deployment where the leader is bounced. This change introduces cleanup logic to that map.

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/PeriodicShardSyncManager.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/PeriodicShardSyncManager.java
@@ -216,19 +216,16 @@ class PeriodicShardSyncManager {
             boolean isRunSuccess = false;
             final long runStartMillis = System.currentTimeMillis();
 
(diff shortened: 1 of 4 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/PeriodicShardSyncManager.java
akidambisrinivasan, line 304: Actually as a good practice we should clean up any state on leader when the leader changes

akidambisrinivasan (approved):

PR #1804, opened 2026-09-03
Ensure worker re-initializes after startup failure

ehasah-aws:
*Issue #, if available:*

*Description of changes:*
 - Fix initialization guard so retries actually re-initialize after a transient failure
 - Prevent duplicate background threads when initialization is retried

### DIFF
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DynamicMigrationComponentsInitializer.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DynamicMigrationComponentsInitializer.java
@@ -303,7 +303,7 @@ public final class DynamicMigrationComponentsInitializer {
         if (blockingWait) {
             log.info("Waiting for Lease table GSI creation");
             final long secondsBetweenPolls = 10L;
-            final long timeoutSeconds = 600L;
+            final long timeoutSeconds = 3600L;
             final boolean isIndexActive =
                     leaseRefresher.waitUntilLeaseOwnerToLeaseKeyIndexExists(secondsBetweenPolls, timeoutSeconds);
 
--- a/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DynamicMigrationComponentsInitializer.java
+++ b/amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DynamicMigrationComponentsInitializer.java
@@ -169,6 +169,10 @@ public final class DynamicMigrationComponentsInitializer {
             final LeaseAssignmentMode assignmentMode,
             final Supplier<? extends LeaderDecider> leaderDeciderSupplier,
             final boolean createLamAndStartMetrics) {
+        if (initialized) {
+            log.info("Startup components already initialized, skipping initializeStartupComponents");
+            return;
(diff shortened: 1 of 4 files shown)


amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DynamicMigrationComponentsInitializer.java
```diff
@@ -299,12 +303,7 @@ private void createGsi(final boolean blockingWait) throws DependencyException {
         if (blockingWait) {
             log.info("Waiting for Lease table GSI creation");
             final long secondsBetweenPolls = 10L;
-            // TODO: there exists an issue where if this timeout is reached the KCL starts up erroneously for a new
-            //   application (Scheduler appears to start but doesn't take any leases and some threads don't start
-            //   properly) where the only recovery is bouncing the worker after the GSI is created.
-            //   Extending the timeout to a long time to ensure GSI is created as a short term fix and will revisit
-            //   on the proper fix in the future
-            final long timeoutSeconds = 3600L;
+            final long timeoutSeconds = 600L;
```
lucienlu-aws: How long does GSI initialization take? From my testing previously, a 4 shard new application took 15-20 mins to initialize the GSI, so we should probably just set the timeout to longer than that to avoid unnecessary retries. We can reduce it though

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/migration/MigrationStateMachineImpl.java
```diff
@@ -113,16 +113,19 @@ public void initialize() throws DependencyException {
...
-            // retries the whole initialization loop.
+            // and initializes components. If enter() throws DependencyException,
+            // startingClientVersion remains null so the next retry re-enters this block.
             final MigrationClientVersionState startingState =
-                    createMigrationClientVersionState(startingClientVersion, startingMigrationState);
+                    createMigrationClientVersionState(dataForInitialization.getKey(), startingMigrationState);
             startingState.enter(ClientVersion.CLIENT_VERSION_INIT);
+
+            // Only set after enter() succeeds — this is the guard that prevents
+            // double-initialization. Setting it before enter() caused the zombie bug.
```
lucienlu-aws: Do we need the second sentence comment?

ehasah-aws: I can clean the comment

amazon-kinesis-client/src/main/java/software/amazon/kinesis/coordinator/DynamicMigrationComponentsInitializer.java
ehasah-aws: I just reverted to what we had initially. For me it took about same time 15-20 minutes. I read https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.OnlineOps.html, GSI creation time depends on few factors like size of table, number of items in the table, etc. 
I agree with your point though. I will change it back to 3600L

lucienlu-aws (approved):

